PR #2, opened 2024-02-27
Create config-fermat.sh

xanthe-cat:
Shell-script cowritten by Teal Dulcet and Catherine Cowie to configure Mlucas for production work on Fermat numbers, by testing the Fermat modular squaring code at various FFT lengths over 10000 iterations, on the appropriate Fermat number for that size, which then writes timing and radix set data to fermat.cfg in the Mlucas executable directory Mlucas/obj. A log file config-fermat.log records the Mlucas output. During installation a copy of this script should be copied to the Mlucas executable directory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
         grep 'warning:' build.log | sed 's/\x1B\[\([0-9]\+\(;[0-9]\+\)*\)\?m//g' | awk '{ print $NF }' | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
         echo '```' >> $GITHUB_STEP_SUMMARY
         for s in tt t s m l; do time ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" |& tee -a test.log; done
+        bash -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
     - uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


xanthe-cat:
Hi Teal,
I’m happy with the script as is now (I don’t think any further micro-optimisations are desirable unless there are actual bug issues raised). I did change the log file config.fermat.log to use the hyphen seeing as this file is called config-fermat.sh.
I imagine this file might better be located in the src directory (even though it’s not part of building Mlucas) and the makemake.sh script simply duplicates it into whatever obj directory it’s making as part of the build.

tdulcet (approved):
Thank you for the PR and for all your work on this script! I can merge it as soon as you have had a chance to review my changes. You may want to review the CI output as well, just click on the "Checks" tab -> any of the Mlucas Linux or macOS jobs -> the "Script" step and then scroll down to the very bottom to see the output from this script.

I do not believe we would need to duplicate the script, as users should be able to just run something like this (similar to what I did in the CI):
```bash
cd obj && bash ../config-fermat.sh [Mlucas args]...
```

xanthe-cat (commented):
All looks good; if the script is going to reside in the main directory the comment at L43 should perhaps use the example of e.g. bash ../config-fermat.sh but I can live with this.

tdulcet:
I updated your comment as requested. Please feel free to make any further changes. I will give @ldesnogu a chance to review this if they want, before merging it.

PR #11, opened 2024-03-04
Fermat number-related patches to Mlucas.c

### LINKED ISSUE #3: FFT selection code
As noted previously in #1 the default FFT selected for small exponents may be accurate but it misses the mark for Fermat exponents which have a limited range of possible FFTs that may be used. At Teal’s suggestion I’ve tested modifications to `Mlucas.c`, firstly at [line 1261](https://github.com/tdulcet/Mlucas/blob/e63bdf1bf49d58a754bc80a80a4db05bcb725180/src/Mlucas.c#L1261):

```c
if(!fft_length || (!INTERACT && 8*fft_length > 9*kblocks && MODULUS_TYPE == MODULUS_TYPE_MERSENNE)) {
```

The added condition in the second clause ensures the 9 > 8 rule for selecting the next FFT doesn’t clobber Fermat exponents and retains the same behaviour for Mersennes. The more extensive modification at lines 1304 ff. or thereabouts prevent Mlucas running the self-test when Mlucas selects a wrong default FFT in spite of a user-supplied FFT.

```c
	SETUP_FFT: /* CXC: addition by Teal D inserted at line 1308 to printf the preferred FFT length; also line 1310 if statement */
		/* Look for a best-FFT-radix-set entry in the .cfg file: */
		dum = get_preferred_fft_radix(kblocks);
		if(!dum) {	// Need to run a timing self-test at this FFT length before proceeding:
			sprintf(cbuf,"INFO: FFT length %d = %d K not found in the '%s' file.\n", n, kblocks, CONFIGFILE);
			fprintf(stderr,"%s", cbuf);
			if (!fft_length) return ERR_RUN_SELFTEST_FORLENGTH + (kblocks << 8);
		}
```

### LINKED ISSUE #4: Final residue of PRP tests not saved for cofactor testing
Again seemingly a problem for the Fermats; Mlucas 21 now has cofactor testing, where either a Mersenne or Fermat number with known factors is PRP tested for compositeness of the cofactor. This is achieved by doing a 3-PRP test of the large number, then exponentiating 3 by the product of known factors – 1, and subtracting the two results modulo the cofactor:

_P_ = _Q_ · _C_, where _P_ = the Mersenne or Fermat number; _Q_ = _p_<sub>1</sub>·_p_<sub>2</sub>·…·_p<sub>i</sub>_ is product of known factors, and _C_ is the cofactor;
_A_ = 3<sup>_P_–1</sup> (mod _P_);
_B_ = 3<sup>_Q_–1</sup> (mod _P_); and by Fermat’s little theorem,
_A_ – _B_ (mod _C_) should be non-zero if _C_ is composite; otherwise _C_ is probably prime to the base 3<sup>_Q_</sup>.

The issue with the Fermat numbers is that the _A_ result is difficult to obtain (Ernst spent 5 years chasing the result for _F_<sub>30</sub>) so the `worktodo` line
```
Fermat,Test=30
```
should save the final residue of the Pépin test, which is 3<sup>½(_F_<sub>30</sub>–1)</sup> (mod _F_<sub>30</sub>), which is exactly one modular squaring before obtaining the _A_ value.

This does not seem to be happening with the Pépin test. Instead when the PRP test completes the Res64 modulo 2<sup>64</sup> residue is written to stdout and the .stat file, but the full residue is not saved, preventing the cofactor test.

**Edited to add**: there is a rather [hacky](https://www.mersenneforum.org/showpost.php?p=651559&postcount=104) workaround to generate the final residue, but it would be preferable not to have to resort to using anything so crude; and indeed there are other superior methods.

The same seems to be true of Mersenne numbers, however the cofactor code fortunately runs correctly and immediately after the primality test as there is only one `worktodo` entry, and so it does not need to “restart”; e.g. having added a ``worktodo`` entry of
```
PRP=n/a,1,2,1063427,-1,99,0,3,5,"68506830638546017,9191962833192559119894251405401"
```
I obtained the following result which matches [previously tested results](https://www.mersenne.org/report_exponent/?exp_lo=1063427&full=1&swversion=1) with mprime:
```
[2024-02-29 10:22:08] M1063427 Iter# = 1063427 [100.00% complete] clocks = 00:00:02.517 [  0.7346 msec/iter] Res64: DC727AB157FD5B77. AvgMaxErr = 0.162886046. MaxErr = 0.281250000. Residue shift count = 639848.
Suyama-PRP on cofactors of M1063427: using FFT length 52K = 53248 8-byte floats.
p1063427.G this gives an average   19.971210186298077 bits per digit
Fermat-PRP residue (A)     = **0x34F046857B8DEDB7**,140702053821552, 4314796808
Processed 158 bits in binary modpow; MaxErr = 0.281250000
3^(F-1) residue (B)        = 0x6EDA90980BA1A104,20717759102, 8128775763
(A - B) Res64 = 0xC615B5ED6FEC4CB3, C Res64 = 0xC893A566EEE74777
(A - B)/C: Quotient = 61782895853496272969963902813274391688714652455, Remainder Res64 = 0x32E74AF916FA4C92
Suyama Cofactor-PRP test of M1063427 / 68506830638546017 / 9191962833192559119894251405401 with FFT length 53248 = 52 K:
	(A - B) mod C has Res64,35m1,36m1: 0x32E74AF916FA4C92,25999072071,15353862932.
GCD(A[1063296 bits], B[1063269 bits]) = 1
Time for GCD = 00:00:00.113
Time for GCD = 00:00:00.113
Cofactor is not a prime power.
If using the manual results submission form at mersenne.org, paste the following JSON-formatted results line:
{"status":"C", "exponent":1063427, "known-factors":["68506830638546017","9191962833192559119894251405401"], "worktype":"PRP-3", "res64":"34F046857B8DEDB7", "residue-type":5, "fft-length":53248, "shift-count":0, "error-code":"00000000", "program":{"name":"Mlucas", "version":"20.1.1"}, "timestamp":"2024-02-28 23:22:08 UTC"}
```

The problem appears to be that Mlucas is unable to restart from a saved residue (which is required for the Fermat numbers as there are two `worktodo` formats for the two tests, the primality test using `Fermat,Test=`, and the cofactor test using `PRP=`).

**Note**.
The cofactor test tries a “sanity check” on each known factor prior to running the test, which has the effect of testing whether the residue _R_ has been correctly calculated _at any point of the computation_:

```
Computing 1060000-squaring residue R (mod known prime q = 68506830638546017)
	A: R == 13206716689333421 (mod q)
	B: R == 13206716689333421 (mod q)
Computing 1060000-squaring residue R (mod known prime q = 9191962833192559119894251405401)
	A: R == 3548111304294208065146791942045 (mod q)
	B: R == 3548111304294208065146791942045 (mod q)
```
Here if the _A_ and _B_ values are equal, then the full residue _R_ has been calculated correctly (_A_ is derived directly from the full residue, while _B_ is calculated solely from the prime factor _q_; they should always be in agreement, modulo _q_).

If the residue _is_ saved correctly and is saved at the right iteration, Mlucas restarts and the cofactor test proceeds correctly. The Mersenne cofactor test can be run separately if a residue is saved and a new factor is found later; for the Fermat test, the saved residue must be from a Pépin test (2<sup>_m_</sup>–1 iterations) or the value is rejected and the test refuses to run. For this reason, Fermat testing **must** consist of two separate `worktodo` entries, e.g.:
```
Fermat,Test=21
PRP=N/A,1,2,2097152,+1,99,0,3,5,"4485296422913"
```

### LINKED ISSUE #5: Pépin residue written with 2-PRP base instead of 3-PRP
This is a weird necessity of the Fermat modular squaring code: at about `Mlucas.c` L1212 Ernst set the PRP_BASE to 2, even though the starting base for the Pépin test is 3. When writing residues to disk though, the wrong base gets written and subsequently causes a problem for any ensuing cofactor testing.
https://github.com/tdulcet/Mlucas/blob/ba49d68b8e62f272bf682f51f3ca77c1bb0e89fa/src/Mlucas.c#L1212-L1215

The fix seems to be to intervene just at the last minute to write 3 as the PRP base to disk on the final iteration, with an `if` statement inserted at about L2223:
```c
if (ihi == maxiter && TEST_TYPE == TEST_TYPE_PRIMALITY && MODULUS_TYPE == MODULUS_TYPE_FERMAT) PRP_BASE = 3;
```

In theory, you may run the Pépin test with [any base](https://oeis.org/A129802) _a_ that has the correct Jacobi symbol (_a_/_F<sub>m</sub>_) = –1, however there is no ability within Mlucas’ code or the `worktodo` format to modify the Pépin test to use a different base, and the test format only allows the exponent _m_ of the Fermat number _F<sub>m</sub>_ to be chosen, _e.g._:
```
Fermat,Test=30
```
This is not the case with the PRP `worktodo` format, so it is quite possible to run Mersenne cofactor testing using a different base; but the results will obviously not match the default values, typically base 3.

### LINKED ISSUE #6: Request for enhanced residues
[Requested](https://www.mersenneforum.org/showthread.php?p=651562&postcount=106) by Gary Gostin.
The result block at about line 2511 of `Mlucas.c` has a long `snprintf` statement,
https://github.com/tdulcet/Mlucas/blob/ba49d68b8e62f272bf682f51f3ca77c1bb0e89fa/src/Mlucas.c#L2510-L2513
This might be better handled by adding a conditional `if (condition) {snprintf} else {snprintf}` that kicks in after the residue shift is printed, as the Mersenne exponents usually have a JSON string to submit to mersenne.org, while nothing is printed for Fermat exponents and the “paste the following” instruction just looks silly with no JSON to paste. Instead, it is customary to print an ‘enhanced’ residue devised by Selfridge and Hurwitz _circa_ 1964, consisting of (then) residues mod 2<sup>36</sup>, mod 2<sup>35</sup>–1, and mod 2<sup>36</sup>–1 (and now: the mod 2<sup>36</sup> is substituted by a hexadecimal mod 2<sup>64</sup> residue; you just read off the last 9 hex digits to recover the mod 2<sup>36</sup> residue). The –1 applied to the other moduli means for large values > 2<sup>36</sup> they function as a pseudo-checksum, since the chances of a coincidental alignment of all three values is ~ 1 in 2<sup>(64+35+36)</sup>.

```c
	snprintf_nowarn(cbuf,STR_MAX_LEN, "%s is not prime. Program: E%s. Final residue shift count = %llu.\n",PSTRING,VERSION,RES_SHIFT); 
	mlucas_fprint(cbuf,1);
	if (MODULUS_TYPE == MODULUS_TYPE_FERMAT) {snprintf_nowarn(cbuf,STR_MAX_LEN, "Selfridge-Hurwitz residues Res64,Res35m1,Res36m1 = %016llX,%11llu,%11llu.\n",Res64,Res35m1,Res36m1);}
	else {snprintf_nowarn(cbuf,STR_MAX_LEN, "If using the manual results submission form at mersenne.org, paste the following JSON-formatted results line:\n%s\n",cstr);}
	mlucas_fprint(cbuf,1);
```
**Edited to add**: In reviewing the code, we found several other places where it might be appropriate to print the “enhanced” residue with Selfridge–Hurwitz moduli. One is when the Suyama PRP-cofactor code is called, but this would require passing the values in, so we chose instead to modify the “restart” printout at [line 1670](https://github.com/tdulcet/Mlucas/blob/e63bdf1bf49d58a754bc80a80a4db05bcb725180/src/Mlucas.c#L1670) that precedes calling the cofactor code.

xanthe-cat:
Set of Fermat number related bug-fixes worked out between TD and CXC. Fixes:
* FFT override available for Fermat exponents in production mode. Fixes #3 
* <strike>Gerbicz error check supported for residue-shifted modular squaring (by commenting out a restriction). Update: (more investigation required)</strike>
* Ensures a final residue is saved to disk for PRP tests, preceding a possible cofactor test. Fixes #4 
* Prints full S–H residues for Fermat exponents both when restarting and completing a PRP test. Fixes #6 
* Ensures final Pépin test iteration is written with correct 3-PRP base. Fixes #5 

More information about each of these issues is available at the links above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
         grep 'warning:' build.log | sed 's/\x1B\[\([0-9]\+\(;[0-9]\+\)*\)\?m//g' | awk '{ print $NF }' | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
         echo '```' >> $GITHUB_STEP_SUMMARY
         for s in tt t s m l; do time ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" |& tee -a test.log; done
+        bash -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
     - uses: actions/upload-artifact@v4
       if: always()
       with:
@@ -81,6 +82,7 @@ jobs:
         cd obj
         make clean
         time ./Mlucas -s m -cpu "0:$(( $(nproc --all) - 1 ))"
+        bash -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
       continue-on-error: true
 
   TSan-Linux:
@@ -217,6 +219,10 @@ jobs:
         grep 'warning:' build.log | sed 's/\x1B\[\([0-9]\+\(;[0-9]\+\)*\)\?m//g' | awk '{ print $NF }' | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
         echo '```' >> $GITHUB_STEP_SUMMARY
         for s in tt t s m l; do time ./Mlucas -s $s -cpu "0:$(( $(sysctl -n hw.ncpu) - 1 ))" 2>&1 | tee -a test.log; done
+        if (( $(sysctl -n hw.optional.AdvSIMD) )); then
+            cd ../obj_NOSIMD
+        fi
+        bash -- ../config-fermat.sh -cpu "0:$(( $(sysctl -n hw.ncpu) - 1 ))"
     - uses: actions/upload-artifact@v4
       if: always()
       with:
@@ -243,6 +249,9 @@ jobs:
         cd obj
         make clean
         time ./Mlucas -s m -cpu "0:$(( $(sysctl -n hw.ncpu) - 1 ))"
+        if ! (( $(sysctl -n hw.optional.AdvSIMD) )); then
+            bash -- ../config-fermat.sh -cpu "0:$(( $(sysctl -n hw.ncpu) - 1 ))"
+        fi
 
   TSan-macOS:
     name: ThreadSanitizer macOS
@@ -315,6 +324,7 @@ jobs:
         done
         cd obj
         for s in tt t s m l; do time ./Mlucas -s $s -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))" |& tee -a test.log; done
+        bash -- ../config-fermat.sh -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))"
     - uses: actions/upload-artifact@v4
       if: always()
       with:
@@ -353,3 +363,4 @@ jobs:
         cd obj
         make clean
         time ./Mlucas -s m -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))" |& tee -a test.log
+        bash -- ../config-fermat.sh -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))"
--- a/README.md
+++ b/README.md
@@ -9,6 +9,33 @@ Mlucas and Mfactor are 100% open source programs. Mlucas is for [primality](http
 
 The original [Mlucas README](https://mersenneforum.org/mayer/README.html) is available for posterity and contains a lot of information, but note that it is no longer up to date. For more information about Mlucas v21, please see the [Ernst's Mlucas - the future](https://www.mersenneforum.org/showthread.php?t=28926) thread on the Mersenne Forum.
 
+Feature | | Mlucas | Prime95/MPrime
+--- | --- | ---: | ---:
+**Architectures** | x86 | ✔️ | ✔️
+\- | ARM | ✔️ | 
+\- | Other | ✔️ | 
+**Worktypes** | LL | ✔️ | ✔️
+\- | PRP | ✔️ | ✔️
+\- | P-1 | ✔️ | ✔️
+\- | P+1 | | ✔️
+\- | ECM | | ✔️
+\- | Pépin | ✔️ | 
+**PRP** | Proofs | | ✔️
+\- | Certs | | ✔️
+**Error Checking** | Jacobi | | ✔️
+\- | Gerbicz | ✔️ | ✔️
+**Random Shifts** | | ✔️ | ✔️
+**Interface** | CLI | ✔️ | MPrime only
+\- | GUI | | Prime95 only
+**Multiple Workers** | | Separate runs | ✔️
+**PrimeNet Support** | | Separate script | ✔️
+**Max FFT Length** | | 256M<br>(**512M** with 0 shift) | 32M (AVX) -<br>64M (AVX512)
+**Largest Exponent** | | 4,294,967,231<br>(**8,937,021,911** with 0 shift) | 595,700,000 (AVX) -<br>1,169,000,000 (AVX512)
+**Performance** | | ~50-90% | **100%**
+**Free** 🆓 | | **Yes**, GPL | No, EULA
+**100% Open Source** | | ✔️ | Mostly
+**Claim Full EFF Awards** | | ✔️ | 
+
 ## Usage
 
 ### Automatic method
--- a/config-fermat.sh
+++ b/config-fermat.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Shell script for generating fermat.cfg; mlucas output saved to config-fermat.log
+# Shell script for generating fermat.cfg; Mlucas output saved to config-fermat.log
 ITERS=10000     # Number of iterations; use 100, 1000, or 10000 to match pre-computed values
 MIN=14          # Minimum Fermat number (14 or greater)
 MAX=29          # Maximum Fermat number (33 or less)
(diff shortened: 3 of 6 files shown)


tdulcet (approved):
Thank you for fixing all of these bugs! It looks good to me.

You might want to add a few brief comments though with links to your issues, so that future contributors could see why these fixes were needed.

xanthe-cat:
Sorry for pulling out one of the bug-fixes in this pull at the last minute, however I need to do more testing of the GEC with residue shifting, and the minimum Fermat with a suitable exponent for this is _F_<sub>20</sub>. If it doesn’t cause a problem I’m happy to try another PR later.

tdulcet:
> Sorry for pulling out one of the bug-fixes in this pull at the last minute

No worries, please feel free to take as much time as you need to test everything. I can merge this PR whenever you are ready.

> I’ve found a definite failure with non-zero residue shift and GEC (with all other variables normal), so this “fix” is now a non-starter.

The zero shift is required for the largest exponents above the 256M FFT length. However, that `exit(1)` call is the most problematic here, as it causes Mlucas to silently exit when attempting to restart a Pépin test for any reason, even with the zero shift.

xanthe-cat:
I’m wondering, Teal, was the save file that you were trying to restart from using zero shift? (The GIMPS_status script will indicate whether it is.) I’ve reinstated that bit of code because, even for small exponents, there is no good reason _not_ to use GEC in production mode; a different roadblock appears if one attempts using non-zero residue shift on _F_<sub>20</sub> or greater, as the GEC tries to check the computation every million iterations by default, and even if Mlucas evaluates the correct residue at that point GEC can’t check that it _is_ correct.
It is possible to run the primality test with non-zero residue shift on the very small exponents where you only need to evaluate about half a million squarings for say _F_<sub>19</sub>, but for multi-million iterations in production mode it doesn’t seem wise to disable error checking.

tdulcet:
Thanks for the explanation. I did not realize that using a nonzero shift was implicitly disabling GEC. I definitely agree that users should be using GEC in production, especially on large exponents.

> I’m wondering, Teal; was the save file that you were trying to restart from using zero shift?

Yes, the savefile was using a zero shift, but that exit call caused it to silently exit. Anyway, your latest changes look good to me.

tdulcet:
No worries now as I just manually fixed it, but in the future when merging PRs, please use the "Squash and Merge" option. This will automatically squash all of the commits in the PR to a single commit before merging it, so that it does not add a bunch of unnecessary commits to the main branch.

xanthe-cat:
Apologies, I did not see that that was both an available, and the preferred, option.

Issue #15, opened 2024-03-29
Mluces incorrectly states that M37 is no prime


Hermann-SW:
I followed Linux build instructions, ugraded to latest and all dev prerequistites were installed already.
I did run <kbd>./Mlucas -s large</kbd> in obj directory.
Then I did run Mlucase on M37.
I run "Ubuntu 22.04.4 LTS" on "AMD Ryzen 9 7950X 16-Core Processor".
@kriesel mentioned in [this posting](https://www.mersenneforum.org/showthread.php?p=653894#post653907) that I should report, and that you might be interested in  p3021377.stat, so I attach that file here (gziping was needed to be able to attach here):
[p3021377.stat.gz](https://github.com/primesearch/Mlucas/files/14801594/p3021377.stat.gz)

This is complete execution stating <kbd>M3021377 is not prime</kbd>, but I see some warnings as well.
I will keep all files in obj directory in case you want other files as well.
```
hermann@7950x:~/Mlucas/obj$ taskset -c 0-7,16-23 ./Mlucas 

    Mlucas 20.1.1

    http://www.mersenneforum.org/mayer/README.html

INFO: testing qfloat routines...
System total RAM = 31190, free RAM = 28673
INFO: 28673 MB of free system RAM detected.
CPU Family = x86_64, OS = Linux, 64-bit Version, compiled with Gnu C [or other compatible], Version 11.4.0.
INFO: Build uses AVX512 instruction set.
INFO: Using prefetch.
INFO: Using inline-macro form of MUL_LOHI64.
INFO: Using FMADD-based 100-bit modmul routines for factoring.
INFO: MLUCAS_PATH is set to ""
INFO: using 64-bit-significand form of floating-double rounding constant for scalar-mode DNINT emulation.
INFO: testing IMUL routines...
INFO: Testing 64-bit 2^p (mod q) functions with 100000 random (p, q odd) pairs...
INFO: System has 32 available processor cores.
INFO: testing FFT radix tables...
User did not set LowMem in mlucas.ini ... allowing all test types.
User did not set CheckInterval in mlucas.ini ... using default.
No CPU set or threadcount specified ... running single-threaded.
Set affinity for the following 1 cores: 0.
Setting ITERS_BETWEEN_CHECKPOINTS = 10000.
 looking for worktodo.txt file...
 worktodo.txt file found...reading next assignment...
 worktodo.txt entry: PRP=1,2,3021377,-1,75,0

[...]

PR #20, opened 2024-04-07
Included the Res2048 in the JSON results

tdulcet:
* Updated to include the Res2048 in the JSON results.
  * It is only included for PRP tests, the same as Prime95/MPrime, but we could included it for LL and/or Pépin tests as well if needed.
  * It is included for both compose and prime results, which matches how Mlucas includes the Res64, but is different from Prime95/MPrime.
* Made agreed upon changes to the `config-fermat.sh` script.
* Updated mersenneforum.org URLs to use HTTPS.

I confirmed that the resulting Res2048 matched Prime95/MPrime for [M500029](https://www.mersenne.org/report_exponent/?exp_lo=500029&full=1).

### DIFF
--- a/config-fermat.sh
+++ b/config-fermat.sh
@@ -73,10 +73,10 @@ for fft in "${!FFTS[@]}"; do
 	elif [[ -n $MAX && $f -gt $MAX ]]; then
 		break
 	fi
-	printf '\n\tTesting F%s (%s),\tFFT length: %sK\n\n' "$f" $((1 << f)) "$fft"
+	printf '\n\tTesting F%s (2^%s + 1),\tFFT length: %sK\n\n' "$f" $((1 << f)) "$fft"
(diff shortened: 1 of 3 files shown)


xanthe-cat (approved):
These all look fine – thanks for updating those links to Ernst’s README, they were something I’d noticed ought to be changed but had overlooked actually doing! The Res2048 will be most helpful comparing my test composites (related to #15) that I also ran in mprime.

xanthe-cat:
For the record, I have a 12 for 12 exact match for Res2048 results between mprime 30.19 b6/b13 and Mlucas 21.0.1; 6 exponents 756853, 859447, 1257827, 1398281, 2976229, 3021407, each tested with a 3–PRP and a 7–PRP test for residue type 1.

tdulcet:
Thank you for testing it! That is good to know that the residues were correct.

Issue #10, opened 2024-03-03
Error `Assertion failed: Nonzero carryout of nonoverlapping vector add!` on x86 when compiled with GCC and LTO enabled


tdulcet:
Mlucas cannot be compiled on x86 with GCC and Link Time Optimization (LTO) enabled, as it triggers a runtime error:
```
INFO: Maximum recommended exponent for FFT length (2048 Kdbl) = 39606917; p[ = 39003229]/pmax_rec = 0.9847580159.
Initial DWT-multipliers chain length = [short] in carry step.
M39003229: using FFT length 2048K = 2097152 8-byte floats, initial residue shift count = 31137256
This gives an average   18.598188877105713 bits per digit
Using complex FFT radices      1024        32        32
mers_mod_square: Init threadpool of 1 threads
Using 1 threads in carry step
ERROR: at line 457 of file ../src/mi64.c
Assertion failed: Nonzero carryout of nonoverlapping vector add!
```
Here is the assert: https://github.com/primesearch/Mlucas/blob/d2dad70871af597919b51377251fbfcf55329d6c/src/mi64.c#L457
The issue seems to affect all FFT lengths. If LTO could be enabled, it could provide a significant performance improvement.

**Edit**: We got [a report](https://www.mersenneforum.org/showpost.php?p=654167&postcount=92) that LTO works on 32-bit ARM.

gbvalor:
On apple silicon  M3 Pro with armv8 (64-bits) and compiled with **-flto=auto** , **Mlucas -s a**  completed selftest without problems. 

OTOH, timings are the same that in no-lto case. No gain using LTO in my tests.

ldesnogu:
> OTOH, timings are the same that in no-lto case. No gain using LTO in my tests.

Depending on what you're running, most of the executed code can be in hand-written assembly loops so the compiler won't be able to improve run time a lot.

tdulcet:
> On apple silicon M3 Pro with armv8 (64-bits) and compiled with **-flto=auto** , **Mlucas -s a** completed selftest without problems.

Interesting, thanks for the information. I was able to confirm that LTO works on my 64-bit ARM system as well. We got another report that LTO works on 32-bit ARM, so the bug must only affect x86 systems.

> No gain using LTO in my tests.

[...]

PR #22, opened 2024-04-11
Fixed assorted compiler warnings

tdulcet:
* Fixed 20 `stringop-truncation` warnings.
* Removed use of the `snprintf_nowarn` macro, which revealed some hidden warnings.
  * Fixed 20 `format-truncation` warnings by doubling the size of the `cbuf` array.
* Fixed 10 `discarded-qualifiers` warnings.
* Updated the `ASSERT` function to (partially) use a macro instead, ~similar to the C standard library assert macro in `assert.h`~.
  * ~This should reduce the number of false positive warnings, as compilers and static analyzers can much more easily check if the assert is going to fail.~
  * I switched to Ernst's C99 `ASSERT` function, which shows the function name in error messages ~and I renamed it to `ABORT`~. I also added the assertion string to the error messages.
  * This allowed removing the `HERE` argument, which I added to the macro instead. This was a major change, as Ernst had over 4,000 assert statements.
  * ~This should result in a performance improvement, as it no longer needs to call a function to test the expression.~
* Updated to use the `int8_t`, `int16_t`, `int32_t` and `int64_t` integer types from `stdint.h` instead of redefining them.
  * This could help with the Windows issues. Unfortunately, to prevent more compiler warnings it required also updating over 700 of the `printf` calls to use the defines from `inttypes.h`.
  * I also used the boolean types from `stdbool.h` while I was at it.

This pull request is a work in progress, as I am not sure it is the best way to fix some of these warnings. I would recommend reviewing each commit individually. Feedback is welcome!


src/types.h
```diff
@@ -63,55 +68,71 @@ so we're not flying blind:
...
+	// #endif
+
+// #else
+	// typedef          long long	 int64;
+	// typedef          long long	sint64;
+	// typedef unsigned long long	uint64;
+	// typedef const          long long	 int64c;
+	// typedef const          long long	sint64c;
+	// typedef const unsigned long long	uint64c;
+// #endif
```
ldesnogu, line 135: I would remove all the commented out defines.

src/util.h
```diff
@@ -229,8 +229,12 @@ void	WARN	(long line, char*file, char*warn_string, char*warn_file, int copy2stde
 	__device__
 	void	ASSERT(long line, char*file, int expr, char*assert_string);
 #else
-	void	ASSERT	(long line, char*file, int expr, char*assert_string);
+	// void	ASSERT	(long line, char*file, int expr, char*assert_string);
+	/* _Noreturn */ void ABORT(const char*assertion, const char*file, long line, const char*func, const char*assert_string);
```
ldesnogu, line 233: Both gcc and clang should accept __attribute__ ((__noreturn__)).

src/types.h
tdulcet: OK, will do.

src/util.h
tdulcet: Yeah, I tried both `__attribute__ ((__noreturn__))` and `_Noreturn` from C11, but from some reason they both caused these unhelpful compiler errors with GCC:
```
../src/mi64.c: Assembler messages:
../src/mi64.c:1327: Error: symbol `loop_shrl_short' is already defined
../src/mi64.c:887: Error: symbol `loop_shl_short2' is already defined
```
I was not able to figure out what was causing the issue...

ldesnogu: The good news is that this means the noreturn helped the optimizer :-)

tdulcet: Yes, that is the main reason why I wanted to remove the `ASSERT` function. That is good to know that the `noreturn` was needed, so thanks for your help getting it to work.

The CI turned up a bunch of similar errors with other labels, so I redid my commit to make more fixes (22 so far): 6fc40900ffa4c843fed0b6cfcf5cf1f59681d46c. Hopefully this time I got them all...

xanthe-cat:
Just a general comment on one particular category you’ve looked at in this PR. Adding the tested condition for the `ABORT` message should be highly useful; Ernst may have had some idea what was going on but reading his code and the output it generates is often rather obscure to me. As much explanatory information as available is sometimes barely enough.

tdulcet:
> Adding the tested condition for the `ABORT` message should be highly useful

Thanks, that is good to know it would be useful for you. It is a feature of the standard library `assert` statement from `assert.h`, which I thought might be useful to add to Ernst's custom implementation. Unfortunately, until someone is able to find a fix for #10, I think I am going to have to rework this PR and remove some of the changes.

src/util.h
tdulcet: Unfortunately, I had to remove these changes from this PR, as they were triggering #10, but I moved them to a new separate PR: #24.

ldesnogu:
Random comments:
- The last commit of that branch (adding const) introduced new errors when doing a non multi-threaded build.
- I think we should have an ASSERT_ALWAYS (or some other name such as FATAL); this would allow us to use the gcc/clang noreturn attribute.

Regarding the first comment above, I'm unsure how to proceed: amend or add a patch. My preference is to have bisectable code, which would mean amending the offending git change, instead of adding a new commit after.

For the second comment, this can come as a later patch.

tdulcet:
> Regarding the first comment above, I'm unsure how to proceed: amend or add a patch.

Please feel free to amend my commit. I changed the base of your PR, so it should automatically update when this is fixed (might need a quick rebase).

> For the second comment, this can come as a later patch.

I previously moved the assert changes to #24, which does add the noreturn attribute, but this is currently blocked on fixing #10. Feel free to make additional changes in the PR if you want.

ldesnogu:
> > Regarding the first comment above, I'm unsure how to proceed: amend or add a patch.
> 
> Please feel free to amend my commit. I changed the base of your PR, so it should automatically update when this is fixed (might need a quick rebase).

I force pushed the change. I will use this as a basis to fix other errors for single-thread compilation (if that proves possible at all, which I'm not sure is the case).

ldesnogu (commented):
EDIT: still making my way through the flow. I thought this comment was only addressing the patch changing the printf format and types :(

LGTM

I guess apart from type changes this was mostly automated so I quickly looked through the diff.

PR #23, opened 2024-04-13
Update Mlucas.c, mi64.c for correct modular division (small scalars)

### LINKED ISSUE #1: Fermat-testing.md
# Introduction to Fermat number testing with Mlucas

Mlucas is a powerful Mersenne primality testing and factoring package which supports Lucas–Lehmer 
testing, Fermat probable primality testing, and Pollard _p_–1 factorisation; justly, most applications of the 
software have concentrated on examination of the Mersenne numbers, however Mlucas is also capable of 
testing the primality of Fermat numbers, testing the compositeness of Fermat cofactors, and running _p_–1
factorisation on Fermat numbers. This post and others hope to fill in several lacunae in the Mlucas 
documentation with respect to Fermat number testing.

## Mlucas build notes

The Fermat modular squaring code is architecture-specific, ideally requiring Mlucas to be built with one of 
the SIMD modes (AVX, AVX2, SSE2, etc). This should be handled automatically by the `makemake.sh` script 
for most hardware, including x86, however if Fermat modular squaring is not natively supported, there may 
be workarounds available.

For example, Apple Silicon is ASIMD, meaning a native Mlucas build cannot test Fermat numbers. The 
answer is to build an SSE2 Mlucas on an Intel-based Mac, which should be capable of running under 
Rosetta emulation on Apple Silicon, at the penalty of a significant performance hit for larger exponents.

(When compiling on Intel with the intention of building a compatible version for Apple Silicon, the Makefile 
requires static linking of the GMP library, by the following alteration to the linker flags:
`LDLIBS = -Bstatic ${LD_ARGS[@]} -Bdynamic`
Static linking of libraries is usually not recommended, and this may not be a solution in all possible cases.)

## Fermat self-testing and populating fermat.cfg

Assuming you have a working build, Mlucas allows any Fermat number _F<sub>m</sub>_ = 2<sup>2<sup>_m_</sup></sup> + 1 with exponent _m_ in 
the range [14,63] to be entered as an argument for self-testing, by using the `-f` flag like so:

`./Mlucas -f <exponent> -iters <number> [-fft <fft_length>]`

However, the software only provides fast Fourier transforms (FFTs) up to 512M in length, which further 
limits the largest Fermat number that may be tested, up to _F_<sub>33</sub>.

The `-iters` flag is mandatory and should be followed by a number less than a million; the recommended 
values for `-iters` to be set to are 100, 1000, or 10000, since the self-testing code has correct residues 
for all Fermat numbers [14,33] saved, and any error in the test result will be immediately discovered.

Mlucas does not support Fermat testing for all possible FFT lengths. Generally, FFTs for testing Mersenne 
numbers are available for any length _k_·2<sup>_n_</sup> K, where _k_ = [8,15], _n_ ≥ 0. 
In the case of Fermat numbers however, only the subset _k_ = {8, 14, 15} supports Fermat testing, along with
_k_ = 63 when _n_ ≥ 4.

When self-testing Mlucas for Mersenne numbers, a command such as `./Mlucas -s all` configures 
Mlucas for the majority of FFT lengths and saves the results in `mlucas.cfg`. A similar process must be 
carried out for the Fermat numbers, using the `./Mlucas -f` command above, which saves results in the 
file `fermat.cfg`.

A script [`config-fermat.sh`](https://github.com/tdulcet/Mlucas/blob/main/config-fermat.sh) has been written to automate the generation of the `fermat.cfg` file using 
the set of possible FFT lengths. After the license, the next few lines declare some variables which may 
be customised as desired (for example, 10000 iterations will require significant runtime at the largest 
possible FFT sizes):
```bash
# Mlucas
MLUCAS=./Mlucas

# Number of iterations
# use 100, 1000, or 10000 to match pre-computed values
ITERS=100

# Minimum Fermat number (14 or greater)
MIN=14

# Maximum Fermat number (33 or less)
MAX=29
```
The script should be invoked from the build directory, typically `obj`, with the following command, which 
may include as parameters any cpu-specific options, such as `-core` or `-cpu`:
```
../config-fermat.sh -cpu 0:3
```
Once self-testing has occurred, production work on Fermat numbers may be carried out using a `worktodo` 
file (Mlucas 20.x: `worktodo.ini`; Mlucas 21: `worktodo.txt`) with entries in one of the several formats 
below.

Only assignments for ECM factoring of Fermat numbers are distributed by Primenet, so work assignments 
cannot be obtained for Fermat numbers using the `primenet.py` script. However _p_–1 results may be 
submitted to [mersenne.org](https://www.mersenne.org/manual_result/) as Manual results.

## Primality testing: Pépin’s test

For the Pépin test, the `worktodo` format is simply:
```
Fermat,Test=<exponent>
```

Fermat numbers in the range [14,22] may select a non-optimal FFT length by default in production mode.
If this is the case, the FFT may be overridden using the command:
```
./Mlucas -fft FFT_LENGTH -shift 0
```
The optimal FFT lengths vary from 4K up to 512M as shown in the table below.

Currently, all Fermat numbers up to _F_<sub>30</sub> have received a [Pépin test](https://www.mersenneforum.org/showthread.php?t=18748), and moreover _F_<sub>31</sub> and _F_<sub>32</sub> are known 
to be composite. However, the character of their cofactors is unknown; and as for _F_<sub>33</sub>, this is yet to be 
tested for primality.

The Pépin test uses Gerbicz error checking (GEC) to assure the reliability of the computation, however 
residue shifting is not implemented for Fermat modular squaring with GEC. Thus when invoking Mlucas the 
flag `-shift 0` _must_ be added to the Mlucas command line:
```
./Mlucas -shift 0
```
If non-zero residue shifting is used, the Pépin test will not be able to progress beyond the millionth 
iteration (the default interval for GEC) as error checking will be unable to confirm whether or not the 
computation is correct.

## Cofactor compositeness testing: Suyama’s test
Suyama’s test is a Fermat probable primality test on the cofactor of a Fermat number. As a prerequisite, 
you _must_ already have run a Pépin test, which should have saved a final residue as a file named e.g. `f23` 
for a Pépin test of _F_<sub>23</sub>. Do **not** try to run a Suyama test as a single work type incorporating the preceding 
Pépin test.

The `worktodo` format for the Suyama test is:
```
PRP=N/A,1,2,<2^m>,+1,99,0,3,5,"known_factor_1[,known_factor_2,...]"
```

Note that this work format uses the numeric value of 2<sup>_m_</sup> of a Fermat number _F<sub>m</sub>_, rather than just the 
exponent _m_. At least one known prime factor must be supplied; composite factors are disallowed.

Prior to testing, Mlucas tries a “sanity check” on each known factor, which has the effect of testing 
whether the Pépin residue _R_ has been correctly calculated. This calculation can be performed at any point 
up to the final iteration. More information is available [here](https://github.com/tdulcet/Mlucas/issues/4).

## Pollard _p_–1 factoring

Work entries for _p_–1 factoring have two variations, however only the `Pminus1` format is supported for 
Fermat numbers:
```
Pminus1=N/A,1,2,<2^m>,+1,B1,B2[,trial_factoring_bits][,B2_start][,"known_factors"]
```

The same note as regards the exponent for the Suyama test applies here also; supplying known factors is 
optional. The `-shift 0` flag is again a necessary addition to the Mlucas command line.

## The testable Fermat numbers _F<sub>m</sub>_ = 2<sup>2<sup>_m_</sup></sup> + 1

The following table lists the default FFT lengths selected by Mlucas for the Fermat numbers in the range 
[14,33] and the known factors, required for some of the test types above.

 _m_|       2<sup>_m_</sup>|Default FFT |Optimal&nbsp;FFTs     |known_factor(s)
--|---------:|-----------:|----------------:|--------------------------------------------------------------------
14|     16384|        1K  |               4K|`"116928085873074369829035993834596371340386703423373313"`
15|     32768|        2K  |               4K|`"1214251009,2327042503868417,168768817029516972383024127016961"`
16|     65536|        3K\*|               4K|`"825753601,188981757975021318420037633"`
17|    131072|  6K\*, 7K  |           7K, 8K|`"31065037602817,7751061099802522589358967058392886922693580423169"`
18|    262144|       13K\*|    14K, 15K, 16K|`"13631489,81274690703860512587777"`
19|    524288|       26K\*|    28K, 30K, 32K|`"70525124609,646730219521,37590055514133754286524446080499713"`
20|   1048576|       52K\*|    56K, 60K, 64K|
21|   2097152|      104K\*| 112K, 120K, 128K|`"4485296422913"`
22|   4194304|      208K\*| 224K, 240K, 256K|`"64658705994591851009055774868504577"`
23|   8388608|      448K  | 448K, 480K, 512K|`"167772161"`
24|  16777216|      896K  | 896K, 960K, 1008K, 1M|
25|  33554432|     1792K  | 1792K, 1920K, 2M|`"25991531462657,204393464266227713,2170072644496392193"`
26|  67108864|     3584K  | 3584K, 3840K, 4032K, 4M|`"76861124116481"`
27| 134217728|  7M, 7.5M  | 7M, 7.5M, 7.875M, 8M|`"151413703311361,231292694251438081"`
28| 268435456|       15M  | 14M, 15M, 15.75M, 16M|`"1766730974551267606529"`
29| 536870912|       30M  |  30M, 31.5M, 32M|`"2405286912458753"`
30|1073741824|       60M  |    60M, 63M, 64M|`"640126220763137,1095981164658689"`
31|2147483648|      120M  | 120M, 126M, 128M|`"46931635677864055013377"`
32|4294967296|      256M  | 240M, 252M, 256M|`"25409026523137"`
33|8589934592|      512M  |       504M, 512M|

\* These default FFTs selected by Mlucas have radices that cannot be used for Fermat modular squaring;
thus _F_<sub>16</sub> to _F_<sub>22</sub> cannot be tested with Mlucas without overriding the default FFT selected, _e.g._:
```
./Mlucas -fft 224K -shift 0
```
As mentioned under “Mlucas build notes” above, building Mlucas is highly architecture-specific, and some 
build types will not support all of the optimal FFT lengths.
Finally, 4K appears to be the smallest usable FFT for the three smallest testable Fermat numbers.

### LINKED ISSUE #18: PRP modular division bug
As discovered in issue #15, a PRP test of a probable Mersenne prime was incorrectly tested, probably as a result of legacy code that was not correctly adapted from the case of Lucas–Lehmer or Pépin tests.

To PRP test a Mersenne prime efficiently with modular squaring, Mlucas evaluates _b_<sup>2<sup>_p_</sup></sup> ≡ _b_<sup>2</sup> (mod 2<sup>_p_</sup> – 1), and then as the final step undertakes a modular division by _b_<sup>2</sup> to obtain the usual form of Fermat’s little theorem, where _b_<sup>2<sup>_p_</sup>–2</sup> ≡ 1 (mod 2<sup>_p_</sup> – 1) indicates the Mersenne number is a probable prime to the base _b_. Any result other than 1 indicates the Mersenne number is composite.

The typical base for this PRP test is 3; of the other small primes, 2 is unsuitable as a base because _any_ Mersenne number with an odd prime exponent is a 2–PRP pseudoprime, while bases 5 and 7 do not present problems. An assert at line 698 of Mlucas.c disables using PRP bases other than 3 (which was put in place probably owing to this issue).

So for base 3, Mlucas performs a modular division by 9, taking the 3<sup>2<sup>_p_</sup></sup> residue from the a[ ] array and storing the result of the division in arrtmp[ ]. (Pull request #17 fixes the primality checking loop that examines the arrtmp[ ] for any non-zero values other than the final 0x1 result.)

Bases 5 and 7 require modular division by 25 and 49 respectively as the final step of the test, and in testing the 5–PRP results were correct (0x1) while the 7–PRP results were not (0x31 ÷ 7<sup>2</sup> ≠ 0x0, the result returned).

Therefore there exists a substantial bug in the modular division code for untypical PRP bases such as 7 (and possibly others).

### LINKED ISSUE #27: Suyama PRP call doesn’t have correctly calculated Selfridge-Hurwitz residues
A quick example to demonstrate, with $M_{130439}$ which is semiprime with a small factor $260879$. The `worktodo` line is:
```
PRP=N/A,1,2,130439,-1,99,0,3,5,"260879"
```
My customised Mlucas test build prints S–H residues before and after the final modular division step:
```
Before mod div: Selfridge-Hurwitz residues Res64,Res35m1,Res36m1 = 0xFF122E035D795FEB, 5596574508,36217851704.
After mod div:  Selfridge-Hurwitz residues Res64,Res35m1,Res36m1 = 0x553AE8AB0A62D1C4, 8242891979,65108185009.
```
The Res64 values are correct, but the “After” Res35m1 and Res36m1 are not. For comparison, the output from cofact 0.74c:
```
                                        |      Selfridge - Hurwitz residues
                       mod 2^64 (hex)   |   mod 2^36    mod 2^36-1   mod 2^35-1
Calculated A residue matches proof file residue;
 Undivided  residue: 0xFF122E035D795FEB |  14453137387  36217851704   5596574508
  Divided   residue: 0x553AE8AB0A62D1C4 |  47418888644  65108186657   8674905291
```
Mlucas’ output from the Suyama call reports the Suyama (A) value but again only the Res64 is correctly calculated:
```
Suyama-PRP on cofactors of M130439: using FFT length 8K = 8192 8-byte floats.
Fermat-PRP residue (A)     = 0x553AE8AB0A62D1C4,          0,          0
Processed 17 bits in binary modpow; MaxErr = 0.000244141
3^(F-1) residue (B)        = 0x553AE8AB0A62D1C4, 8674905291,65108186657
(A - B) Res64 = 0x0000000000000000, C Res64 = 0xED5CFA1BF90E6C11
(A - B)/C: Quotient = 0, Remainder Res64 = 0x0000000000000000
This cofactor is PROBABLE PRIME [PRP39261].
```
Again, to compare with cofact:
```
Testing the cofactor for primality using the following 1 supplied factor: 260879 
Cofactor is 39261 digits long
                                        |      Selfridge - Hurwitz residues
                       mod 2^64 (hex)   |   mod 2^36    mod 2^36-1   mod 2^35-1
Suyama    A residue: 0x553AE8AB0A62D1C4 |  47418888644  65108186657   8674905291
Suyama    B residue: 0x553AE8AB0A62D1C4 |  47418888644  65108186657   8674905291
(A-B) mod C residue: 0x0000000000000000 |            0            0            0
M130439 cofactor is (probably) prime!
```
I tried looking at how the (A) residue is passed from the main Mlucas code to see why Mlucas reports zeroes for the Res35m1 and Res36m1 values but couldn’t make head or tail of fixing it.

xanthe-cat:
Small fix, for PRP bases other than 3, to ensure the modular division by the square of the base at the end of a PRP test is not subject to a rounding error (bases 7, 14, 27, 28, and 29 are susceptible among the possible PRP bases less than 32). The rounding error for these bases appears in the `mi64_div_by_scalar64` function of `mi64.c`. We also modify the `Mlucas.c` code to permit PRP bases other than 3 by disabling the `ASSERT` at line 698 (which would apply to both Mersenne and Fermat exponents) and instituting a similar check later, but only for Fermat numbers (since Mlucas always uses base 3 for the Pépin test).
This fixes #18, fixes #27, as well as committing the Fermat testing documentation to the repository which fixes #1.


src/Mlucas.c
xanthe-cat, line 1: Another afterthought. We actually _do_ want to enforce this line 698 `ASSERT` in the case of the Suyama test, since there is no capability for running the Pépin test on Fermat numbers with bases other than 3. So I think the following modification would be preferable:
```c
i = (int)strtol(char_addr+1, &cptr, 10);	if (MODULUS_TYPE == MODULUS_TYPE_FERMAT) ASSERT(HERE, i == 3,"PRP-test base for Fermat numbers must be 3!");
```

tdulcet: Feel free to make this change, but it may be better to add this as a new `ASSERT` statement a few lines down, so it could be applied only in the case that the user also supplied some known factors. That would still allow users to run regular PRP tests of Fermat numbers with arbitrary bases.

xanthe-cat: Yes, but that’s the curious thing that Mlucas doesn’t run a PRP test properly (with the cofactor test following on afterwards) if you specify it with a `PRP=...` entry in `worktodo`. This was one of the oddities I found last month, that the `PRP` test for e.g. _F_<sub>18</sub> will terminate after 262,144 iterations, but the Suyama test for Fermat numbers assumes you have a residue from 262,143 iterations and will not proceed otherwise!
Weirdly, you have to do, e.g. `Fermat,Test=18` first, and then a separate `PRP=1,2,262144,1,...` item second.

tdulcet: Yes, I meant that you would only need to apply this `ASSERT` statement in the case that the user is performing a Suyama test, not a general PRP test of a Fermat number. One could differentiate these two cases by checking if the user provided a list of known factors, which unless I am missing something, would mean they must be performing a Suyama test. The known factors are not read until a few lines down, so the new `ASSERT` statement would need to be added somewhere after that:
```c
if (MODULUS_TYPE == MODULUS_TYPE_FERMAT && KNOWN_FACTORS[0] != NULL) ASSERT(HERE, i == 3,"PRP-CF test base for Fermat numbers must be 3!");
```

xanthe-cat: Ah, gotcha. Yes, that would probably make more sense.

src/mi64.c
xanthe-cat, line 1: More discussion on this fix (whether this is an appropriate way to resolve the float to int conversion issue) is over at issue #18.

src/Mlucas.c
xanthe-cat: One last little bit of weirdness: in the initial `Mlucas.c` comments where global variables are being defined Ernst canvasses the possibility of using other bases for Pépin tests – which flies in the face of how Mlucas actually works elsewhere (as line 698 is pretty unambiguously ruling it out).
https://github.com/primesearch/Mlucas/blob/93b703dca04b648e78bedd13ac3e3da3a662b373/src/Mlucas.c#L153-L154

tdulcet: From looking at the `git blame`, it looks like that assert statement [was added](https://github.com/primesearch/Mlucas/commit/0396583e96d3c12ff49d254ff96adc7c1ebadb3b#diff-4eb3a064be6340b86e444f667ce8df7eefe365b06b55ced61a86eab4387c894fL591-R590) in Mlucas v20.0, so before that any PRP bases greater than 1 and odd were supported. You may want to test v19, although I suspect Ernst realized some PRP bases were not working correctly (or someone reported a bug), but he did not have time to debug it and so added that assert statement as a temporary solution, but did not update his other comments.

xanthe-cat:
I’ve added in the Fermat documentation (and some small tweaks to the config script) from issue #1 as this is now ready to be added to the repository, whatever final disposition we reach on the status of whatever modification is desirable to `mi64.c`.

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+testing the primality of Fermat numbers, testing the compositeness of Fermat cofactors, and running _p_–1
+factorisation on Fermat numbers. This document is aimed at filling in several lacunae in the Mlucas 
+documentation with respect to Fermat number testing.
+
+## Mlucas build notes
+
+The Fermat modular squaring code is architecture-specific, ideally requiring Mlucas to be built with one of 
+the SIMD modes (AVX, AVX2, SSE2, etc). This should be handled automatically by the `makemake.sh` script 
+for most hardware, including x86, however if Fermat modular squaring is not natively supported, there may 
+be workarounds available.
```
tdulcet, line 15: 64-bit ARM is the only architecture where SIMD is unsupported for Fermat numbers.

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+
+For example, Apple Silicon is ASIMD, meaning a native Mlucas build cannot test Fermat numbers. The
```
tdulcet, line 17: Apple Silicon is ARM and supports ASIMD.

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+
+However, the software only provides fast Fourier transforms (FFTs) up to 512M in length, which further 
+limits the largest Fermat number that may be tested, up to _F_<sub>33</sub>.
+
+The `-iters` flag is mandatory and should be followed by a number less than a million; the recommended 
+values for `-iters` to be set to are 100, 1000, or 10000, since the self-testing code has correct residues 
+for all Fermat numbers [14,33] saved, and any error in the test result will be immediately discovered.
+
+Mlucas does not support Fermat testing for all possible FFT lengths. Generally, FFTs for testing Mersenne 
+numbers are available for any length _k_·2<sup>_n_</sup> K, where _k_ = [8,15], _n_ ≥ 0.
```
tdulcet, line 41: Did you mean to list `k` twice here?

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+```
+Fermat,Test=<exponent>
+```
+
+Fermat numbers in the range [14,22] may select a non-optimal FFT length by default in production mode.
+If this is the case, the FFT may be overridden using the command:
+```
+./Mlucas -fft FFT_LENGTH -shift 0
+```
+The optimal FFT lengths vary from 4K up to 512M, as shown in the table below under “testable Fermat numbers”.
```
tdulcet, line 92: Links to other sections are supported:
```suggestion
The optimal FFT lengths vary from 4K up to 512M, as shown in the table below under [testable Fermat numbers](#the-testable-fermat-numbers-fm--22m--1).
```

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Cofactor compositeness testing: Suyama’s test
+Suyama’s test is a Fermat probable primality test on the cofactor of a Fermat number. As a prerequisite, 
+you _must_ already have run a Pépin test, which should have saved a final residue as a file named e.g. `f23` 
+for a Pépin test of _F_<sub>23</sub>. Do **not** try to run a Suyama test as a single work type incorporating the preceding 
+Pépin test.
+
+The `worktodo` format for the Suyama test is:
+```
+PRP=N/A,1,2,<2^m>,+1,99,0,3,5,"known_factor_1[,known_factor_2,...]"
```
tdulcet, line 117: It might help to explain more of these parameters, especially if we now support other PRP bases:
```
PRP=N/A,1,2,<2^m>,1[,how_far_factored,tests_saved[,base,residue_type[,"known_factors"]]]
```
The plus in the `+1` is optional and assumed.

docs/Fermat-testing.md
```diff
@@ -0,0 +1,186 @@
...
+31|2147483648|      120M  | 120M, 126M, 128M|`"46931635677864055013377"`
+32|4294967296|      256M  | 240M, 252M, 256M|`"25409026523137"`
+33|8589934592|      512M  |       504M, 512M|
+
+\* These default FFTs selected by Mlucas have radices that cannot be used for Fermat modular squaring;
+thus _F_<sub>14</sub> to _F_<sub>22</sub> cannot be tested with Mlucas without overriding the default FFT selected, _e.g._:
+```
+./Mlucas -fft 224K -shift 0
+```
+As mentioned under “Mlucas build notes” above, building Mlucas is highly architecture-specific, and some
```
tdulcet, line 184:
```suggestion
As mentioned under [Mlucas build notes](#mlucas-build-notes) above, building Mlucas is highly architecture-specific, and some 
```

tdulcet (commented):
Thanks for your work on the Fermat number documentation. I just had some minor feedback. Feel free to add a link to the main README as well.

docs/Fermat-testing.md
xanthe-cat: By `k` twice, do you mean _k_ and K in line 41? If so, they are different items, as _k_ is an integer value and K is a kilobyte (as the FFTs are measured in multiples of K or M).

docs/Fermat-testing.md
xanthe-cat: I’m a bit on the fence about this; I don’t want to raise the issue of different bases in this context since the `worktodo` format for the Pépin test does not support bases other than 3; and choosing a different base for composite numbers always involves obtaining a completely different set of final residues.
(Also I like the plus in `+1` to clearly indicate what it’s doing following the exponent.)

docs/Fermat-testing.md
xanthe-cat: Now that I read it the directions of ‘under’ and ‘above’ are a little clumsy. I’ll update this with the link and a slight amendment to the text.

docs/Fermat-testing.md
tdulcet: Yes, thanks for the explanation. I do not believe the K units are needed here, as K is just 2<sup>10</sup>, so your formula would still hold, although _n_ then must be ≥ 10.

docs/Fermat-testing.md
tdulcet: (Yes, I like the `+1` as well, I just thought I should note that it is not strictly required by the format.)

docs/Fermat-testing.md
xanthe-cat: I’d rather leave the formula untouched, so I’ve explicitly stated that the length is in multiples of kilobytes.

docs/Fermat-testing.md
xanthe-cat: Thanks for this. (Obviously I haven’t had the opportunity to test all the possible variants.) I hope the rewriting of this paragraph (and the start of the next) meets your approval.

tdulcet: No problem. Yes, it looks great to me! (Of course, this is your documentation and I know very little about Fermat numbers, so feel free to ignore my feedback.)

docs/Fermat-testing.md
tdulcet: Some users may want to run a PRP test on one of the (three) supported Fermat numbers without known factors, in which case maybe some of the parameters could potenchally be useful.

xanthe-cat: Hmm, PRP tests would then deserve their own section separate of the Suyama test. Let me do some testing and I’ll consider adding it.

First problem, an assert that a PRP test should actually be PRP-CF but no factors have been supplied.
```
ERROR: at line 1217 of file ../src/Mlucas.c
Assertion failed: Fermat-mod PRP test implies a PRP-CF run, but no known-factors provided!
```
https://github.com/primesearch/Mlucas/blob/d5faeff78f390b70576b4a97d2388808541608f8/src/Mlucas.c#L1218
Second problem, an assert that a Fermat Suyama test should start from _p_–1 squarings, when the PRP test passes in the _p_ squarings residue.
```
ERROR: at line 3305 of file ../src/Mlucas.c
Assertion failed: Fermat-mod cofactor-PRP test requires p-1 mod-squarings!
```
https://github.com/primesearch/Mlucas/blob/d5faeff78f390b70576b4a97d2388808541608f8/src/Mlucas.c#L3306
Now it _is_ possible to bypass the following code which squares the _p_–1 residue, so we should be starting at _p_, but then the (A) printout has wrong Selfridge–Hurwitz residues:
```
Fermat-PRP residue (A)     = 0xA52CF038FA870F39,          0,          0
```
It looks like it runs the rest of the cofactor test correctly. However any attempt to restart fails, because the Gerbicz integrity test falsely assumes the base is 3.
```
Computing 30000-squaring residue R (mod known prime q = 1214251009)
	A: R == 865617297 (mod q)
	B: R == 1138999586 (mod q)
ERROR: at line 5272 of file ../src/Mlucas.c
Assertion failed: Full-residue == 3^nsquares (mod q) check fails!
```
This looks like a can of worms that I would not want to broach at this time; it appears as though we can allow PRP tests for bases ≠ 3 for Mersennes. Unfortunately it is not possible to do the same for Fermat numbers without a number of fiddly rewrites.

xanthe-cat: I’ve pushed a commit to resolve the issue with Gerbicz’s nsquares integrity check (the assertion at 5272), since it assumes the PRP base is always 3, and the construction of the test would cause a roadblock for restarting PRP tests with bases ≠ 3. There are still plenty of other issues with PRP tests for Fermats using bases ≠ 3 that would make me not want to document this, because Ernst clearly moulded the Mlucas code to only support the default Pépin test with base 3 and a whole raft of code changes would be needed to properly support other bases (which is not the case with the Mersennes).

tdulcet: Interesting, thanks for testing it and for the detailed write up. Also, thank you for fixing the GEC issue. As you know, when Ernst originally added support for the PRP test for Mersenne numbers in v19, it did support bases ≠ 3, but this was disabled in v20 (I suspect because he realized it did not work). If I had to guess, when he was adding support for PRP cofactor tests in v21, he did not bother adding support bases ≠ 3 knowing already that they did not work. Anyway, feel free to open an issue if you want bases ≠ 3 to be supported for Fermat numbers.

xanthe-cat: It isn’t the GEC – it’s another neat trick suggested by Robert Gerbicz that Ernst worked into the restart code, based on this discussion [here](https://www.mersenneforum.org/showthread.php?t=18748&page=8), and which I mention in the long footnote to issue #4.

tdulcet: Oh, interesting, thanks for the clarification.

Just a random aside from reading #4, since you write a lot of complex math formulas, GitHub [supports LaTeX syntax](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions), if you have used that before. See https://github.com/preda/gpuowl/pull/278 for some examples.

xanthe-cat: The last time I seriously used TeX or LaTeX was in my Honours year, which is now thirty years in the past… I used to have a copy of Lamport’s book which got lost somewhere over various house moves.
I’m investigating how the Pépin test could be adapted for other bases (not all possible bases give a determinative result), so have been running PRP tests with the cofactor test following on; the main problem is the lack of specification for the base in the `worktodo` format. I coded other bases into my customised version of Gary’s [cofact](https://www.mersenneforum.org/showthread.php?p=636276&postcount=30) utility and it does appear the Mlucas code works correctly for Fermat modular squaring and cofactor testing in other bases, as I obtained identical Suyama results for base 5 tests of $F_{15}$ and $F_{19}$.

tdulcet: Interesting, thanks again for testing all of this. Regarding support for other bases, probably the easiest solution would be to update the existing `PRP=`/`PRPDC=` formats to fully support Fermat numbers, but as you noted above, this could still require a lot of changes.

I had not heard of Gary’s cofact utility, but we could always create a repository in this GitHub organization to store it, along with your customized version, if that would be helpful.

tdulcet:
I made a simplification to the `config-fermat.sh` script, so avoid duplicating the Mlucas command.

For future reference, when you are just updating the documentation or code comments and do not need to run the CI to test the changes, you can add `[skip ci]` (or one of [these other strings](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs)) to your commit message to skip it.

src/Mlucas.c
```diff
@@ -5262,10 +5264,13 @@ int read_ppm1_savefiles(const char*fname, uint64 p, uint32*kblocks, FILE*fp, uin
...
-			// Raise 3 to the just-computed power; result in 4-limb local-array pow[]:
-			mi64_scalar_modpow_lr(3ull, exp, KNOWN_FACTORS+i, j, pow);
+			// Raise PRP base (usually but not always 3) to the just-computed power; result in 4-limb local-array pow[]:
+			mi64_scalar_modpow_lr(PRP_BASE, exp, KNOWN_FACTORS+i, j, pow);
 			sprintf(cstr,"\tB: R == %s (mod q)\n",&cbuf[convert_mi64_base10_char(cbuf, pow, j, 0)] ); mlucas_fprint(cstr,1);
-			ASSERT(HERE, mi64_getlen(pow,4) == k && mi64_cmp_eq(pow,rem,k), "Full-residue == 3^nsquares (mod q) check fails!");
+			if (mi64_getlen(pow,4) != k) {
+				snprintf_nowarn(cbuf,STR_MAX_LEN,"Full-residue == %lu^nsquares (mod q) check fails!", PRP_BASE); mlucas_fprint(cbuf,0);
+				ASSERT(HERE, mi64_getlen(pow,4) == k && mi64_cmp_eq(pow,rem,k), cbuf);
+			}
```
tdulcet, line 5273: To avoid repeating the `mi64_getlen()` function call, I do not think you need the conditional here, as the `snprintf()` call is not expensive. If you want to keep this conditional, I would go with your approach in 585a79c592a4125f0be6aa113fbc7bc1c344a775, just reverse each of the comparisons:
```c
if (mi64_getlen(pow,4) != k || !mi64_cmp_eq(pow,rem,k)) {
```
Also, the `PRP_BASE` variable is not a long, so you would need to use the `%u` printf specifier instead. If you remove the `_nowarn`, I suspect you would get a warning about that.

xanthe-cat: This is all fine. I must have been very sleepy when I made those changes (besides failing to reverse the logic of the assert); in future I will try not to edit GitHub without brain in gear.

src/Mlucas.c
```diff
@@ -2432,7 +2433,7 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 			// by base and check that remainder 0. Note that we do want the quotient now, as that is our reside/base:
 			mi64_div(c_uint64_ptr, &itmp64, j+1,1, arrtmp,&rmodb);	ASSERT(HERE, rmodb == 0ull,"After short-div, R != 0 (mod B)");
 			// And recompute the S-H residues:
-			res_SH(arrtmp,i,&Res64,&Res35m1,&Res36m1);
+			res_SH(arrtmp,j,&Res64,&Res35m1,&Res36m1);
```
xanthe-cat, line 2436: I think this value should be j, judging by the data I’ve looked at j usually is equal to i+1, so the calculation of the mod 2^35-1 and mod 2^36-1 residues is not encompassing the entire length of the arrtmp array. This partially resolves #27.

src/Mlucas.c
```diff
@@ -3310,6 +3311,8 @@ uint32 Suyama_CF_PRP(uint64 p, uint64*Res64, uint32 nfac, double a[], double b[]
 /*A*/	ierr = func_mod_square(a, (int*)ci, n, ilo,ihi, 0ull, p, scrnFlag, tdiff, TRUE, 0x0);
 		convert_res_FP_bytewise(a, (uint8*)ci, n, p, Res64, &Res35m1, &Res36m1);	// Overwrite passed-in Pepin-Res64 with Fermat-PRP one
 		snprintf_nowarn(cbuf,STR_MAX_LEN,"MaxErr = %10.9f\n",MME); mlucas_fprint(cbuf,1);
+	} else if (MODULUS_TYPE == MODULUS_TYPE_MERSENNE) {	// Mersenne PRP-CF doesn't have the Res35m1 or Res36m1 values passed in,
+		res_SH(ci,n,&itmp64,&Res35m1,&Res36m1);			// so we refresh these; see https://github.com/primesearch/Mlucas/issues/27
```
xanthe-cat, line 3315: This resolves the remaining part of #27, as the Suyama (A) value should print with correct `Res35m1` and `Res36m1` values. (`Res64` is already correct so we don’t “re_update” it by referencing `itmp64` instead.)

src/Mlucas.c
```diff
@@ -5262,10 +5266,13 @@ int read_ppm1_savefiles(const char*fname, uint64 p, uint32*kblocks, FILE*fp, uin
...
 				ASSERT(HERE, 0, "Only known-factors < 2^256 supported!");
-			// Raise 3 to the just-computed power; result in 4-limb local-array pow[]:
-			mi64_scalar_modpow_lr(3ull, exp, KNOWN_FACTORS+i, j, pow);
+			// Raise PRP base (usually but not always 3) to the just-computed power; result in 4-limb local-array pow[]:
+			mi64_scalar_modpow_lr(PRP_BASE, exp, KNOWN_FACTORS+i, j, pow);
 			sprintf(cstr,"\tB: R == %s (mod q)\n",&cbuf[convert_mi64_base10_char(cbuf, pow, j, 0)] ); mlucas_fprint(cstr,1);
-			ASSERT(HERE, mi64_getlen(pow,4) == k && mi64_cmp_eq(pow,rem,k), "Full-residue == 3^nsquares (mod q) check fails!");
+			if (mi64_getlen(pow,4) != k || !mi64_cmp_eq(pow,rem,k)) {
+				snprintf_nowarn(cbuf,STR_MAX_LEN,"Full-residue == %u^nsquares (mod q) check fails!", PRP_BASE); mlucas_fprint(cbuf,0);
+				ASSERT(HERE, mi64_getlen(pow,4) == k && mi64_cmp_eq(pow,rem,k), cbuf);
```
tdulcet, line 5274: You can now remove the redundant condition:
```suggestion
				ASSERT(HERE, 0, cbuf);
```

xanthe-cat: Done. As I said before, the ASSERT statements scare me.

docs/Fermat-testing.md
xanthe-cat: I am amenable to adding a repository for GitHub but I would like Gary’s approval before doing so – and he was responsible for all versions before 0.6, which is the earliest I have a copy of. It’s a relatively simple Makefile and single C source file, but it would be nice to have all of the previous versions for populating blame with useful stuff; my fork splits from 0.6 and is now at 0.74c (the c stands for “Cathy”) while Gary is somewhere beyond 0.8.1.

tdulcet: I can send Gary a message and ask if he would be OK moving the development of cofact to GitHub. We could then host your fork either in a separate repository or in a separate branch in the same repository, which ever you preferred. Unlike with Mlucas which is currently community maintained, you and Gary would have complete control over the cofact repository.

Even if Gary does not want to use GitHub, we could always host a mirror, as I do for the other Prime Search related programs. However, in this case I do not believe it could automatically be updated, as the downloads are only on the forum and not a website that could be scrapped, so it would be better to move the development to GitHub if possible.

ldesnogu (commented):
The change looks good to me.

tdulcet:
> The change looks good to me.

Just note that you would need to check the "Approve" radio button when writing the review to allow @xanthe-cat to merge this. Each PR needs at least one approving review.

ldesnogu:
> Just note that you would need to check the "Approve" radio button when writing the review to allow @xanthe-cat to merge this. Each PR needs at least one approving review.

I would if I could see it :-) Sorry, I'm a newbie.

ldesnogu (approved):
OK let's approve it.

gary641:
This is in a section of the code I am not familiar with. So I will trust Cathy's changes!

PR #24, opened 2024-04-14
Replaced ASSERT function with a macro and enabled LTO

tdulcet:
* Updated the `ASSERT` function to use a macro instead, similar to the C standard library assert macro in `assert.h`.
  * This should reduce the number of false positive warnings, as compilers and static analyzers can much more easily check if the assert is going to fail.
  * This should also result in a performance improvement, as it no longer needs to call a function to test the expression.
  * It seems to enable more loop unrolling, as evident by the required changes to the inline assembly. Thanks to @ldesnogu for help with that in #22.
* Updated the `makemake.sh` script to enable Link Time Optimization (LTO) by default.

These `ASSERT` function changes were originally part of #22, but I had to remove them due it causing #10 on x86 systems when compiled with GCC:
```
INFO: Maximum recommended exponent for FFT length (3 Kdbl) = 66742; p[ = 66431]/pmax_rec = 0.9953402655.
M66431: using FFT length 3K = 3072 8-byte floats, initial residue shift count = 16763
This gives an average   21.624674479166668 bits per digit
Using complex FFT radices        12         8        16
mers_mod_square: radix0/2 not exactly divisible by NTHREADS - This will hurt performance.
Using 4 threads in carry step
ERROR: Function mi64_shlc, at line 457 of file ../src/mi64.c
Assertion 'cy == 0ull' failed: Nonzero carryout of nonoverlapping vector add!
```

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -838,7 +838,7 @@ uint64	mi64_shl_short(const uint64 x[], uint64 y[], uint32 nshift, uint32 len)
 		"vmovd	%[__n] ,%%xmm14		\n\t"/* shift count - since imm-operands only take compile-time consts, this costs a vector register */\
 		"vmovd	%[__nc],%%xmm15		\n\t"/* complement-shift count, 64-n */\
 		"vmovdqa	     (%%rax),%%ymm0	\n\t"/* preload x[i0-(0:3)] */\
-	"loop_shl_short:		\n\t"\
+	"loop_shl_short%=:		\n\t"\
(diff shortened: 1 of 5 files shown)


src/mi64.c
ldesnogu, line 1: Nice catch!
An alternative might have been to clear cy in the inline assembly I guess.

tdulcet:
I am going to take @ldesnogu's comment as an approving review in order to get this merged.

PR #26, opened 2024-04-21
Updated to save Res2048 before PRP cofactor test

### LINKED ISSUE #25: Suyama cofactor test trashes Res2048 value
At lines 2480 and 2481 the lower 32 words of the arrtmp[] array is read, however this needs to precede the call to Suyama_CF_PRP at line 2467, since the computation of the cofactor by dividing known factors uses arrtmp[].
https://github.com/primesearch/Mlucas/blob/93b703dca04b648e78bedd13ac3e3da3a662b373/src/Mlucas.c#L2455-L2483

tdulcet:
* Updated to save Res2048 before PRP cofactor test. Fixes #25

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -2461,6 +2461,11 @@ PM1_STAGE2:	// Stage 2 invocation is several hundred lines below, but this needs
 	// the list of known factors used, and the basic "is PRP?" result for the corresponding cofactor.
 	/************************************************************************************************************************/
 
+		char Res2048[513];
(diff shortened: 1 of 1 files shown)


xanthe-cat (approved):

src/Mlucas.c
xanthe-cat, line 1: Thanks – the Res2048 test values match mprime’s, which is always nice :)

tdulcet: No problem. Thank you for testing it and reporting the bug!

PR #34, opened 2024-07-28
Ld warnings

ldesnogu:
Check build.


docs/ConvertInlineAsm.txt
tdulcet, line 1: Where did this file come from? It seem we would need to perform the opposite conversion, GCC --> MSVC, if we wanted to add MSVC support.

src/Mlucas.c
```diff
@@ -6186,58 +6191,58 @@ void generate_JSON_report(
...
+			snprintf(p_cstr,strlen(p_cstr)-1,"{\"status\":\"%c\", \"exponent\":%" PRIu64 ", \"worktype\":\"%s\", \"res64\":\"%016" PRIX64 "\", \"fft-length\":%u, \"shift-count\":%" PRIu64 ", \"error-code\":\"00000000\", \"program\":{\"name\":\"Mlucas\", \"version\":\"%s\"}, \"timestamp\":\"%s\", \"aid\":\"%s\"}\n",prp_status[isprime],p,ttype,Res64,n,RES_SHIFT,VERSION,timebuffer,aid);
 		} else {
-			snprintf(cstr,STR_MAX_LEN,"{\"status\":\"%c\", \"exponent\":%" PRIu64 ", \"worktype\":\"%s\", \"res64\":\"%016" PRIX64 "\", \"fft-length\":%u, \"shift-count\":%" PRIu64 ", \"error-code\":\"00000000\", \"program\":{\"name\":\"Mlucas\", \"version\":\"%s\"}, \"timestamp\":\"%s\"}\n",prp_status[isprime],p,ttype,Res64,n,RES_SHIFT,VERSION,timebuffer);
+			snprintf(p_cstr,strlen(p_cstr)-1,"{\"status\":\"%c\", \"exponent\":%" PRIu64 ", \"worktype\":\"%s\", \"res64\":\"%016" PRIX64 "\", \"fft-length\":%u, \"shift-count\":%" PRIu64 ", \"error-code\":\"00000000\", \"program\":{\"name\":\"Mlucas\", \"version\":\"%s\"}, \"timestamp\":\"%s\"}\n",prp_status[isprime],p,ttype,Res64,n,RES_SHIFT,VERSION,timebuffer);
```
tdulcet, line 6196: Per @ldesnogu's [comment on the forum](https://www.mersenneforum.org/node/22731?p=1084815#post1084815), these and the below `strlen` calls would need to be removed:
> The proper fix is to pass the buffer size to the generate_JSON_report function that needs it.

MarkRose:
At the point the great majority of it is already in `main`, and `main` has since moved ahead of the branch, so I think this can be closed.

## What landed

Of the 147 commits here, **144 are already in `main` verbatim**. `main` additionally carries **3 commits that this branch does not** — "Fix AArch64 duplicate assembly error", "Fix duplicate declaration of `addr` in radix16_ditN_cy_dif1.c", and "Fix initialization of carries for single thread in radix24_ditN_cy_dif1.c" — i.e. follow-up fixes to this same work.

Content-level check agrees with the commit-level one: of 1590 hunks, **1338 are already in `main`**, and a three-way merge (merge-base / `main` / this branch) reproduces `main` byte-for-byte for **73 of 127 files**. Of the 54 with any delta, resolving to `main`'s side leaves only 15 hunks in 11 files that would actually change anything.

## What has not landed

**`src/threadpool.c` and `src/threadpool.h`** — the only substantive remainder. These are carried by open PR #110 (Windows/macOS CPU affinity), whose `threadpool.h` is **byte-identical** to this branch's, and whose `threadpool.c` has the same `xmalloc`/`xcalloc` move, the same `&& defined(CPU_IS_X86)` gate on the force-align attribute, and the same scoping of `cbuf`/`str`/`cpu_set_t`/`thread_id` into the branches that use them. So this part is not lost, it is in review elsewhere.

**`docs/ConvertInlineAsm.txt`** — the one file unique to this branch. No content is lost by dropping it: it is a verbatim relocation of the "Recipe for MSVC → GCC inline ASM conversion" block, which `main` still has inline in `src/radix16_dif_dit_pass.c`. If the split-out file is wanted, it is a trivial standalone change against current `main`.

## Why merging as-is would now regress `main`

The branch predates several fixes that have since landed, so a merge today would undo them. The clearest:

* **`atol` → `atoll`.** `main` now uses `atoll()` for `-iters`, `-shift`, `-b1`, `-nthread`, `-m`, `-base`, `-f` (13 call sites, 0 remaining `atol`); this branch has the mirror image (0 `atoll`, 13 `atol`). That is merged PR #160, the Windows/LLP64 overflow fix — it would be reintroduced.
* **`src/pm1.c`** — adds `(void)nq; // currently unused`, but `main` has since **deleted the `nq` declaration entirely**. This would be a compile error.
* **`src/radix15_ditN_cy_dif1.c`** — adds a bare `#endif`; `main` resolved the same spot as a balanced `#if 0 && PFETCH … #endif`, so this would leave the file preprocessor-unbalanced.
* **`src/radix32_main_carry_loop.h`** — comments out `addi = (double *)cy_i;` and `++addi` in the GENFFTMUL block. `main` solved this in `carry.h` with `(void)sizeof(cy)` and an explicit comment warning that dropping the caller's `cy_i` setup *silently breaks the V1 variant*. This hunk is precisely that.

A few smaller ones point the same way: `Mlucas.c::gcd()` here uses `gmp_snprintf(gcd_str, sizeof(gcd_str), …)` where `gcd_str` is a `char*` parameter, so `sizeof` is 8 — `main` uses `STR_MAX_LEN` with a comment explaining exactly that; `util.c::mkdir_p()` here makes a nonzero `system("mkdir …")` abort the program, which (with plain `mkdir`, no `-p`) fires whenever the directory already exists; and `main` uses the portable `SYS_gettid` where this uses `syscall(__NR_gettid)`.

For completeness, one place this branch is nicer than `main`: `prefetch.h`'s `do { (void)(_pd); } while (0)` versus `main`'s bare `(void)(_pd)`. Cosmetic.

## Suggestion

Close this, with the threadpool changes tracked by #110. Thoughts?

ldesnogu:
I agree, let's close it.

PR #35, opened 2024-08-02
Singlethread

ldesnogu:
This branch is for fixing compilation errors in single thread mode.

It hasn't been tested.


tdulcet (dismissed):
This looks good to me, although I am not qualified to review 4f3b2aa693932b5eac1e3fde0ddafe63a113cca7.

To prevent future regressions, we should probably add some automated testing. Let me know what you think would be the priority to test and I can add that before merging.

tdulcet (approved):

PR #38, opened 2024-10-20
Fast CPUs have gotten faster: reduce self-test times by factor 5

corneliusroemer:
I just downloaded Mlucas on my M1 MBP (32GB RAM) and ran the tiny and small self tests. They were at least a factor of 5 faster than what was stated.

This PR divides all self-test run-times by around a factor of 5.

Self-test runtimes:
tiny: 8s (instead of 1 minute)
small: 1m23s (instead of 10 minutes)

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -123,7 +123,7 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
 **************** THIS IS THE ONLY SELF-TEST ORDINARY USERS ARE RECOMMENDED TO DO: *************
 *                                                                                             *
 * -s medium Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering *
-* -s m      FFT lengths from 2M to 7.5M. This will take around an hour on a fast CPU.         *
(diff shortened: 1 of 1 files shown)


tdulcet:
Thanks for the PR!

How many threads were you using to benchmark? It looks like you forgot to update the medium and egregious/xxl self-test times.

Note that more of this help documentation file needs to be updated. For example, it is missing the new teensy self-test and a few of other the FFT length ranges were also updated by Ernst in v21, which could partially explain the time discrepancies you noticed. If possible, I would recommend running some of the larger self-tests on your system to confirm the new runtimes.
https://github.com/primesearch/Mlucas/blob/4db9fbc42ba00298c123623549bd36dd47573d45/src/Mlucas.c#L3455-L3467

corneliusroemer:
I now also ran the medium (recommended default) test (`Mlucas -s m`) and it took 10m9s (single CPU), I forgot to update that one actually, done now!

I didn't forget to update the egregious one - I just didn't know whether several days would now be just one day or not, so I played it safe and kept it at several days.

PR #39, opened 2024-10-21
Added error counts to JSON results

tdulcet:
* Added error counts to JSON results for LL and PRP tests.
* Moved the smallest 4 FFT lengths from the large to the medium self-test to cover expected growth in wavefront first time primality testing.
* Added new Mersenne prime to arrays of known primes!
* Changed P-1 results to report as `P-1` instead of `PM1` in the JSON results for constancy with Prime95/MPrime.
* Updated README to reflect that the PrimeNet program is now called AutoPrimeNet.
* Bumped version to 21.0.2.


KarlLudwig3485 (approved):
Looks good.

I noticed that Prime95's JSON output includes `Roundoff` and `Jacobi` errors for LL, and either `dblchk` or `gerbicz` for PRP tests. Apart from the inconsistent capitalisation the most notable part to me is that Prime95 doesn't provide a seperate `Roundoff` key for PRP tests, only including it in the error code. It probably doesn't make any difference on the server side, but for readability it seems like an improvement to include it as a seperate field as you have done.

tdulcet:
Thanks for the review.

> I noticed that Prime95's JSON output includes `Roundoff` and `Jacobi` errors for LL, and either `dblchk` or `gerbicz` for PRP tests.

Yeah, Mlucas does not yet support the Jacobi error check for LL tests or `dblchk` error checking for PRP tests. I figured the separate values would be good to include in the JSON results, as they do not have an arbitrary limit on the maximum number of errors, in addition to the improved readability, as you mentioned.

PR #41, opened 2024-11-20
Add support for ARM Windows

KarlLudwig3485:
Some simple fixes to allow Mlucas to be compiled with MinGW for armv7 and aarch64.

I've run `-s m` for both architectures with LTO enabled, both work fine.

My 2012 Surface RT may take a few years to run [a PRP-DC](https://www.mersenne.org/M85809707), but modern ARM Windows laptops are genuinely useful.

You may remember my earlier foibles from April, which resulted in the "Cygwin" branch, which I can only describe as an artifact of the follies of my youth. This is the same thing, but done properly.

### DIFF
--- a/src/platform.h
+++ b/src/platform.h
@@ -896,7 +896,7 @@ states that the compiler defines __64BIT__ if compiling in 64-bit mode.
 
 	   The symbol __ARM_EABI__ is not defined if compiles under the old EABI.
 	*/
-	#if !(defined(__ARMEL__) && defined(__ARM_EABI__))
+	#if !defined (__MINGW32__) && !(defined(__ARMEL__) && defined(__ARM_EABI__))
 		#error __ARM_ARCH predefine seqeunce expects both __ARMEL__ and __ARM_EABI__ to be defined! Please check your platforms predefine list using 'gcc -dM -E - < /dev/null' and forward the results to the program author/maintainer(s) listed on the README page.
 	#endif
 
--- a/src/getRealTime.c
+++ b/src/getRealTime.c
@@ -11,7 +11,7 @@
 
 #if defined(_WIN32)
 
-	#include <Windows.h>
+	#include <windows.h>
 
 #elif defined(__unix__) || defined(__unix) || defined(unix) || (defined(__APPLE__) && defined(__MACH__))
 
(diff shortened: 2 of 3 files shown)


src/platform.h
```diff
@@ -896,7 +896,7 @@ states that the compiler defines __64BIT__ if compiling in 64-bit mode.
 
 	   The symbol __ARM_EABI__ is not defined if compiles under the old EABI.
 	*/
-	#if !(defined(__ARMEL__) && defined(__ARM_EABI__))
+	#if !defined (__MINGW32__) && !(defined(__ARMEL__) && defined(__ARM_EABI__))
```
tdulcet, line 899: Just curious, are either of those `__ARMEL__` or `__ARM_EABI__` symbols defined?

tdulcet (approved):
Thanks for the PR! This looks good to me, but I will give the other reviewers a chance to look at it before merging.

I suppose I can now go ahead and remove that Cygwin.

src/platform.h
KarlLudwig3485: `__ARMEL__` is defined by armv7-w64-mingw32-clang but not aarch64-w64-mingw32-clang.
`__ARM_EABI__` isn't defined by either.

tdulcet: Thanks for the information. @ldesnogu - Do you know if either of these are important on ARM?

ldesnogu: I think __ARMEL__ and __ARM_EABI__ are only applicable to ARMv7.

KarlLudwig3485: I'm specifically targeting both 32-bit ARMv7 and the modern "ARM64" (aarch64) Windows. Windows' ARM ABI is based on the EABI, so there is no point to a check for OABI vs EABI, which is why there's no mechanism to do so i.e. the `__ARM_EABI__` predefine.

I know that I'm probably the only person who will ever run Mlucas on 32-bit ARM Windows ([Windows RT](https://en.wikipedia.org/wiki/Windows_RT) or [Windows 10 build 15035](https://betawiki.net/wiki/Windows_10_build_15035)), but it works fine on my Surface RT. It's incredibly slow though, so not something I would recommend to anyone without ridiculous amounts of patience.

PR #42, opened 2025-01-11
Fix carry-out in x86-64 assembly version of mi64_add()

### LINKED ISSUE #10: Error `Assertion failed: Nonzero carryout of nonoverlapping vector add!` on x86 when compiled with GCC and LTO enabled

KarlLudwig3485:
Also removes assert statement for cy init, as it gets optimized out anyway.
(`cy` is always zero, so the assert is never triggered, so it gets optimized out. Then, prior to this fix, the `cy` init gets optimized away as the inline assembly incorrectly states that it doesn't read `cy`. So the assert is never triggered, even in the situation it was intended for.)

Fixes #10

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -2327,8 +2327,9 @@ uint64	mi64_add_cyin(const uint64 x[], const uint64 y[], uint64 z[], uint32 len,
 		// SdyBr: 2.20; Haswell: 1.84
 		// Jun 2016: bizarre ... GCC builds with opt > 0 on Haswell/Broadwell init this != 0 ...
 		//  making static not a reliable workaround, so try put cy = 0 init on separate line from declaration:
-		uint64 cy;
-		cy = 0ull;	ASSERT(cy == 0, "Init (cy = 0) fails!");
+		// Jan 2025: The operand __cy was incorrectly set as write-only ('=') in the inline assembly, causing
+		//  the cy = 0 init to be optimized out. Setting __cy to read/write ('+') fixes this.
+		uint64 cy = 0ull;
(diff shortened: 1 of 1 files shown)


tdulcet:
Thanks for debugging this and for the PR! I copied your commit over to #24, so we can see if the CI now succeeds when enabling LTO by default.

ldesnogu:
As I said in the other commit: nice catch. Thanks!

KarlLudwig3485:
You're most welcome!
It was quite an entertaining and educational exercise.
I had fun :D

Issue #8, opened 2024-03-03
Error with native Windows builds: `Assertion failed: thread-local memcheck fail!`


tdulcet:
Error with native Windows builds:
```
INFO: Maximum recommended exponent for FFT length (16 Kdbl) = 343407; p[ = 341749]/pmax_rec = 0.9951719097.
Initial DWT-multipliers chain length = [hiacc] in carry step.
M341749: using FFT length 16K = 16384 8-byte floats, initial residue shift count = 341536
This gives an average   20.858703613281250 bits per digit
Using complex FFT radices        16        16        32
mers_mod_square: Init threadpool of 1 threads
executing worker thread id 0, #cores = 12
ERROR: at line 1444 of file ../src/radix16_ditN_cy_dif1.c
Assertion failed: thread-local memcheck fail!
```
The issue seems to only affect the powers of 2 FFT lengths, such as 16K, 32K, 64K, 128K, 256K, 512K, 1M, 2M, 4M, 8M, 16M, 32M, etc. Interestingly, when Mlucas is run in GDB or compiled with ASan and UBSan enabled, this error usually goes away, so it is difficult to debug. It must be more Windows compatibility issues, similar to the datatype issues I fixed in https://github.com/tdulcet/Mlucas/commit/011549e027777577e5fd119145e852a612b8d4ba. It currently prevents users from being able to use the native Windows builds.

tdulcet:
I just had the assertion fail on a 64-bit ARM ASIMD build on Linux using GCC, while doing P-1 stage 1. Here was the output:
```
pm1_set_bounds: Stage 2 needs at least 24+5 buffers ... each buffer needs 60 MB; avail-RAM allows 83 such.
Setting default p-1 stage bounds b1 = 1200000, b2_start = 1200000, b2 = 43000000.
INFO: precomputed p-1 stage 1 primes-product file p143809313.s1_prod not found...computing from scratch.
INFO: Opened precomputed p-1 stage 1 primes-product file p143809313.s1_prod for writing...
Product of Stage 1 prime powers with b1 = 1200000 is 1733006 bits (27079 limbs), vs estimated 1776540. Setting PRP_BASE = 3.
INFO: Maximum recommended exponent for FFT length (7680 Kdbl) = 144258522; p[ = 143809313]/pmax_rec = 0.9968860834.
Initial DWT-multipliers chain length = [hiacc] in carry step.
INFO: primary restart file p143809313 not found...looking for secondary...
[...]

PR #43, opened 2025-01-14
Fix support for 32-bit x86 Windows

KarlLudwig3485:


tdulcet:
What is the issue that this function causes on Windows? Are 32-bit Windows builds affected by #8?

We probably should add some CI jobs for 32-bit systems, although my understanding is that Ernst was removing support. For full 32-bit x86 support with SIMD, I believe one would need to resurrect all of the `gcc32` files.

KarlLudwig3485:
Sorrry for not adding a description earlier. I was a bit tired.

MSYS2 is currently in the process of dropping their 32-bit environments, but LLVM-MinGW (my MinGW toolchain of choice) still fully supports 32-bit x86.

This function increases the stack limit on 32-bit Linux (see the comments above it), but because the MinGW Windows builds use the Linux codepath (with overrides) it gets included on 32-bit x86 Windows builds, resulting in a bunch of errors relating to the non-existance of the rlimit structs/functions on Windows.

```
../src/util.c:1905:16: error: variable has incomplete type 'struct rlimit'
 1905 |         struct rlimit stack_limits;
      |                       ^
../src/util.c:1905:9: note: forward declaration of 'struct rlimit'
 1905 |         struct rlimit stack_limits;
      |                ^
../src/util.c:1907:6: error: call to undeclared function 'getrlimit'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1907 |         if (getrlimit(RLIMIT_STACK, &stack_limits)) {
      |             ^
../src/util.c:1907:16: error: use of undeclared identifier 'RLIMIT_STACK'; did you mean 'SIID_STACK'?
 1907 |         if (getrlimit(RLIMIT_STACK, &stack_limits)) {
      |                       ^~~~~~~~~~~~
      |                       SIID_STACK
/opt/llvm-mingw/llvm-mingw-ucrt/i686-w64-mingw32/include/shellapi.h:671:5: note: 'SIID_STACK' declared here
  671 |     SIID_STACK = 55,
      |     ^
../src/util.c:1912:55: error: use of undeclared identifier 'RLIM_INFINITY'
 1912 |                stack_limits.rlim_cur, stack_limits.rlim_max, RLIM_INFINITY);
      |                                                              ^
../src/util.c:1918:6: error: call to undeclared function 'setrlimit'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1918 |         if (setrlimit(RLIMIT_STACK, &stack_limits)) {
      |             ^
../src/util.c:1918:16: error: use of undeclared identifier 'RLIMIT_STACK'; did you mean 'SIID_STACK'?
 1918 |         if (setrlimit(RLIMIT_STACK, &stack_limits)) {
      |                       ^~~~~~~~~~~~
      |                       SIID_STACK

```

Assembly code for 32-bit x86 doesn't seem to be present in current versions, but a pure nosimd C build works fine.

In my attempts to debug #8 I tried to make 32-bit x86 build to see if it got triggered in those (it doesn't), which is why I made this patch.

tdulcet (approved):
Thanks for the information and the fix!

> Assembly code for 32-bit x86 doesn't seem to be present in current versions, but a pure nosimd C build works fine.

The assembly code was removed by Ernst (in v19.1?), but was added back in v21 in 77749764e9f49906e0208e05bc5ba0c375a044a9 due to still being present in the files provide by George from Ernst's system. See all of the files that end with `gcc32.h`. If someone wanted SIMD support on 32-bit x86, they would need to hook it up again.

Issue #29, opened 2024-04-28
Undefined behavior on Windows: "call to function X through pointer to incorrect function type"


tdulcet:
Undefined behavior detected when Mlucas is compiled on Windows with Clang and UBSan enabled:
```
../src/threadpool.c:452:5: runtime error: call to function mers_process_chunk through pointer to incorrect function type 'void (*)(void *, int)'
../src/threadpool.c:452:5: runtime error: call to function fermat_process_chunk through pointer to incorrect function type 'void (*)(void *, int)'
```

Issue #7, opened 2024-03-03
Segmentation fault in the "teensy" self-test


tdulcet:
A segmentation fault occurs in the "teensy" self-test: `./Mlucas -s tt`. Here is the output when Mlucas is run in GDB:
```
INFO: Maximum recommended exponent for FFT length (1 Kdbl) = 22788; p[ = 22679]/pmax_rec = 0.9952167808.
Initial DWT-multipliers chain length = [hiacc] in carry step.
M22679: using FFT length 1K = 1024 8-byte floats, initial residue shift count = 20848
This gives an average   22.147460937500000 bits per digit
Using complex FFT radices        32        16
[New Thread 33016.0x38a4]
mers_mod_square: Init threadpool of 1 threads
executing worker thread id 0, #cores = 12

Thread 5 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 33016.0x38a4]
0x00007ff75b14a060 in radix16_wrapper_square (a=a@entry=0x5b4400, arr_scratch=arr_scratch@entry=0x64a5b0, n=n@entry=1024, radix0=<optimized out>, rt0=0x655c00, rt1=0x655f80, nradices_prim=9, radix_prim=0x7ff75b4f63a0, ws_i=0, ws_j1=736, ws_j2=768, ws_j2_start=992,
    ws_k=30, ws_m=56, ws_blocklen=128, ws_blocklen_sum=128, init_sse2=0, thr_id=14, fwd_fft_only=<optimized out>, c_arr=0x0) at ../src/radix16_wrapper_square.c:1706
1706                    SSE2_RADIX16_CALC_TWIDDLES_LOACC(cc0,add0,add1,rt0,rt1);
```
Here is the output when Mlucas is compiled with ASan and UBSan enabled:
```
INFO: Maximum recommended exponent for FFT length (1 Kdbl) = 22788; p[ = 22679]/pmax_rec = 0.9952167808.
Initial DWT-multipliers chain length = [hiacc] in carry step.
M22679: using FFT length 1K = 1024 8-byte floats, initial residue shift count = 20848
This gives an average   22.147460937500000 bits per digit
Using complex FFT radices        32        16
mers_mod_square: Init threadpool of 1 threads
../src/radix16_wrapper_square.c:1679:74: runtime error: left shift of negative value -68424725
../src/radix16_wrapper_square.c:1680:5: runtime error: signed integer overflow: -1094795586 + -1094795586 cannot be represented in type 'int'
[...]

PR #45, opened 2025-07-09
Fix occasional thread-local memcheck fail

### LINKED ISSUE #8: Error with native Windows builds: `Assertion failed: thread-local memcheck fail!`

KarlLudwig3485:
(Always occurs on x86-64 Windows, one report on ARM Linux)

~Note that I branched from before ARM Windows was added to the CI, as that broke Windows CI entirely.
No one seems to have noticed, as Windows builds were already broken. Probably ought to be fixed...~

Fixes #8

### DIFF
--- a/src/radix1008_ditN_cy_dif1.c
+++ b/src/radix1008_ditN_cy_dif1.c
@@ -386,7 +386,7 @@ int radix1008_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 
 	if(p != psave || n != nsave
 	#ifdef USE_PTHREAD	// Oct 2021: cf. radix176_ditN_cy_dif1.c for why I added this
-		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si))
+		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si || tdat[0].rn0 != rn0))
 	#endif
 	) {	/* Exponent or array length change triggers re-init */
 		first_entry=TRUE;
--- a/src/radix1008_ditN_cy_dif1.c
+++ b/src/radix1008_ditN_cy_dif1.c
@@ -386,7 +386,7 @@ int radix1008_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 
 	if(p != psave || n != nsave
 	#ifdef USE_PTHREAD	// Oct 2021: cf. radix176_ditN_cy_dif1.c for why I added this
-		|| (tdat != 0x0 && tdat[0].wt1 != wt1)
+		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si))
 	#endif
 	) {	/* Exponent or array length change triggers re-init */
 		first_entry=TRUE;
--- a/src/radix1024_ditN_cy_dif1.c
+++ b/src/radix1024_ditN_cy_dif1.c
@@ -331,7 +331,7 @@ int radix1024_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 
 	if(p != psave || n != nsave
 	#ifdef USE_PTHREAD	// Oct 2021: cf. radix176_ditN_cy_dif1.c for why I added this
-		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si))
+		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si || tdat[0].rn0 != rn0))
 	#endif
 	) {	/* Exponent or array length change triggers re-init */
 		first_entry=TRUE;
--- a/src/radix1024_ditN_cy_dif1.c
+++ b/src/radix1024_ditN_cy_dif1.c
@@ -331,7 +331,7 @@ int radix1024_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 
 	if(p != psave || n != nsave
 	#ifdef USE_PTHREAD	// Oct 2021: cf. radix176_ditN_cy_dif1.c for why I added this
-		|| (tdat != 0x0 && tdat[0].wt1 != wt1)
+		|| (tdat != 0x0 && (tdat[0].wt0 != wt0 || tdat[0].wt1 != wt1 || tdat[0].si != si))
 	#endif
 	) {	/* Exponent or array length change triggers re-init */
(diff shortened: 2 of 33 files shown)


tdulcet:
> Sorry to be the bearer of bad news, but ARM Windows CI is still broken.

Yeah, I saw that about the Windows ARM runners actually building x86-64. However, based on the CI results for this PR, the Windows ARM builds are OK. It is the Windows x86-64 builds with Clang that are seg faulting. I thought that was also related to #8, but maybe it is #29 or something else.

> Also, why does the ARM Windows CI not build with GMP and hwloc? They're in the repos.

The limitations with the Windows ARM builds are due to the runners provided by GitHub Actions, which include fewer tools than the equivalent x86-64 runners, so there is not much we can (easily) do about it. Specifically, MSYS2 is not installed (and does not seem supported on ARM), which means the pacman package manger is not available and thus we cannot install GMP or hwloc. Clang is also not installed. See [this list](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) for what is available. I do not have access to any Windows ARM systems, so I am not sure if there is anything we could do short of compiling the tools and libraries ourselves.

KarlLudwig3485:
> > Sorry to be the bearer of bad news, but ARM Windows CI is still broken.
> 
> Yeah, I saw that about the Windows ARM runners actually building x86-64. However, based on the CI results for this PR, the Windows ARM builds are OK. It is the Windows x86-64 builds with Clang that are seg faulting. I thought that was also related to #8, but maybe it is #29 or something else.
> 
> > Also, why does the ARM Windows CI not build with GMP and hwloc? They're in the repos.
> 
> The limitations with the Windows ARM builds are due to the runners provided by GitHub Actions, which include fewer tools than the equivalent x86-64 runners, so there is not much we can (easily) do about it. Specifically, MSYS2 is not installed (and does not seem supported on ARM), which means the pacman package manger is not available and thus we cannot install GMP or hwloc. Clang is also not installed. See [this list](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) for what is available. I do not have access to any Windows ARM systems, so I am not sure if there is anything we could do short of compiling the tools and libraries ourselves.

MSYS2's ARM support is a little confusing. MSYS2 itself only runs on ARM under x64 emulation, but the CLANGARM64 environment is native aarch64.

If the ARM Windows image (which I see is still in beta) doesn't include MSYS2, I don't quite see the point of building the x64 version on ARM. Still, I suppose we can just leave it as-is until MSYS2 gets added, at which point we can test ARM Windows builds in addition to x64-under-emulation.

Back to the issue at hand, I myself don't encounter the segfault with the Clang build, but I am yet to try it with MSYS2.
Does the Clang build use the CLANG64 environment (i.e. a pure LLVM toolchain), or do you just install Clang in the MINGW64 environment, and still use the rest of the GNU toolchain?

The segfault is definitely a separate issue.

tdulcet:
> If the ARM Windows image (which I see is still in beta) doesn't include MSYS2, I don't quite see the point of building the x64 version on ARM.

Yeah, I did not realize until after I had already added the Windows ARM runner that one could not actually build it for native ARM on Windows or I probably would not have bothered.

> Does the Clang build use the CLANG64 environment (i.e. a pure LLVM toolchain), or do you just install Clang in the MINGW64 environment, and still use the rest of the GNU toolchain?

The latter, it uses MINGW64, as I was trying to mirror how it is built on Linux. From looking closer at the output, it seems to be #7. If we update the CI to skip the "teensy" self-test on Windows, it would likely resolve the problem.

More concerning is the Windows ASan build, which does use the CLANG64 environment. It is still producing the error (in the Fermat code), for example:
```
ERROR: Function radix32_ditN_cy_dif1, at line 1234 of file ../src/radix32_ditN_cy_dif1.c
Assertion 'tdat[ithread].rn0 == rn0' failed: thread-local memcheck fail!
```

tdulcet:
> The Clang build didn't segfault. Now that I've pushed a commit and the CI has run again, the Clang build failed again.

Some of the GitHub Actions runners are slightly different, which can sometimes affect the reproducibility of volatile bugs like seg faults. I am able to reproduce it on both Windows and Linux. However, as you said, it is a separate issue, so it should not block us from merging this PR.

MarkRose:
_The coverage audit below was performed by Claude Code (Anthropic's AI CLI) on my behalf; I've reviewed and am posting it._

I had all 58 `radix*_ditN_cy_dif1.c` carry files gone through to double-check this PR's coverage, and it's complete — every file that can actually trigger the assert is patched here.

The assert can only fire where a file **(a)** caches the weight pointers per-thread (`tdat[i].wt0 = wt0`, etc.) **and (b)** is actually built multithreaded. Of the 58:

- **33 are exposed** (threaded *and* cache the pointers) — **all 33 are patched in this PR.**
- **radix63** has the caching/memcheck code physically, but disables threading itself near the top (`#undef MULTITHREAD` / `#undef USE_PTHREAD`, _"No parallel impl of this experimental code yet"_), so that code is dead and can't fire.
- The rest (72, 80, 88, 96, 104, 112, 120, 512, and the small radices) have no per-thread weight-pointer caching at all — nothing to go stale.

So there are no additional files needing the guard; extending it further would just be dead code. The Mersenne (`wt0`/`wt1`/`si`) + Fermat (`rn0`) split across your two commits lines up with which radices are Fermat-capable.

One useful note for verification: on current `main`, the Windows x86-64 **gcc** CI job reproduces this exact assert — once the AVX-512 build mode is skipped on MinGW gcc (whose assembler can't build it), the build reaches the `-s t` self-test and fires `Assertion 'tdat[ithread].wt0 == wt0' failed` at `radix16_ditN_cy_dif1.c:1444` (16K FFT, radices 16/16/32). So this is now directly reproducible/verifiable in Windows CI.

tdulcet (approved):
Thanks for debugging and fixing this. Sorry it has taken so long to be merged.

Since this has now been tested by @MarkRose (and Claude AI), yourself and Ken on the forum, I am going to go ahead and approve it.

PR #47, opened 2025-07-14
Make MinGW Windows builds report as Windows

### LINKED ISSUE #46: MinGW Windows builds report OS as Linux
Due to the MinGW builds using the Linux codepath with overrides, the OS name is reported as Linux.

The question is, should (potential future) MSVC builds and MinGW builds report differently (e.g. `Windows` and `Windows (MinGW)`), or should they both just be `Windows`?

I'm leaning towards the latter, as the compiler identifies which one was used (MSVC [or compatible, e.g. clang-cl] for MSVC, GCC-compatible for MinGW), but I'd like hear to everyone elses thoughts before making a PR.

KarlLudwig3485:
Fixes #46

### DIFF
--- a/src/platform.h
+++ b/src/platform.h
@@ -1140,7 +1140,7 @@ states that the compiler defines __64BIT__ if compiling in 64-bit mode.
 #endif
 
 /* Time to name names: */
-#if(defined(OS_TYPE_WINDOWS))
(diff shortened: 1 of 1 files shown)


tdulcet (approved):
Thanks. It is too bad we have to mess up Ernst's nice list of OS defines, but I do not see a cleaner way to do this.

KarlLudwig3485:
The only way I can see to keep the list of OS defines tidy would be to rework the MinGW codepath to use the `OS_TYPE_WINDOWS` define, and that would be rather bit more difficult than this. Perhaps a project for another day.

PR #55, opened 2025-11-04
Untested fix for PRP optional how_far_factored,tests_saved

### LINKED ISSUE #30: Mlucas 21.0.1 - Assertion failed: Expected ',' not found after TF_BITS field in assignment-specifying line!
HI All,

I installed today Mlucas 21.0.1 from main branch, no issues to compile.
Then I chose worktype 160 with the current Primenet 2.0.1
This version is faster than 20.x from last month.

But I got an error message: 

This is the output: Assertion failed: Expected ',' not found after TF_BITS field in assignment-specifying line!

Mlucas 21.0.1

System total RAM = 23236, free RAM = 13175
INFO: 13175 MB of free system RAM detected.
INFO: Using prefetch.
NTHREADS = 4
 worktodo.txt entry: PRP=2B37D274B3B7D8A18DC3D3E17E27EEAC,1,2,16582879,-1,"1628121653153521"

ERROR: at line 658 of file ../src/Mlucas.c
Assertion failed: Expected ',' not found after TF_BITS field in assignment-specifying line!

[2]+  Exit 1                  ./Mlucas -cpu 0:3

Artoria2e5:
Is it alright if I just let the CI do it?

Should fix #30, gotta figure out if there's aplace to add test cases though!


Artoria2e5:
Huh. Might as well try the agent mode since I still want to catch more sleep. Actually forget about going back to sleep (that almost never worked! worse chances than writing code without tests) I've gotta change the makemake

tdulcet:
> Someone is gonna need to fix the 1,048,576 or however many clang-tidy errors sometime.

Many of those errors should already be fixed in #34. We just need help reviewing that massive PR.

> Also, the current `main` baseline seems to have left something dubious in the form of an undefined `twopmodq100_2WORD_DOUBLE_q4`

This is a longstanding known issue with Mfactor (build from the same Mlucas codebase), which Ernst was aware of. Many of the Mfactor build combinations currently fail to compile on all OSs.

> In the meantime windows 11 arm gcc just has no gmp. Very funny.

Yes, see https://github.com/primesearch/Mlucas/pull/45#issuecomment-3056861444 for the reason. The Windows builds are largely broken, but that should be fixed by #45. As long as your changes successfully compile on Windows without additional warnings, feel free to ignore the runtime errors.

Artoria2e5:
It's going well so far. I am going to mark "ready" again. In the meantime I should open another branch to get Windows CI sorted.

tdulcet:
> I can confirm that the Windows build failures have nothing to do with me.

Thanks for looking into it. I have the CI scheduled to run once a month, to catch issues like this. It looks like the Windows CI broke some time in September (it was successful on September 1, but failed October 1), most likely because GitHub upgraded the version of GCC, which turned one of the previous warnings into an error. This should be fixed by #34, which resolves all of the remaining compiler warnings. We may need to extract the relevant commit out from #34 if this needs to be fixed sooner.

Regarding your PR, this looks like a lot of added complexity for a single use of `memrchr`. Could we possibly refactor this to use a different function instead?

Artoria2e5:
Of course, when it comes to detecting whether GNU-extension funcs are available and taking steps to replace them -- as well as writing the most contrived script to work around ancient Bourne shell bugs -- very few things can beat GNU's own solution: gnulib + autotools. It might even end up easier for whoever wants to build the project just because the process is more familiar and standardized. It's definitely not less complex for us though. Assuming bash is present makes shell scripts much easier to write. Worst case scenario, the user needs to build bash first.

src/config.analyzer.h
tdulcet: This file can now be removed.

tdulcet (changes requested):

src/Mlucas.c
```diff
@@ -6440,6 +6454,53 @@ uint32 extract_known_factors(uint64 p, char*fac_start) {
...
+		line_end = line_start + strlen(line_start);
+	}
+	const char* last = line_end - 1;
+	if (startq_out != NULL) {
+		*startq_out = line_end;
+	}
+	if (last > line_start && *last == '\"') {
+		// This is an inlined memrchr(). Not using it since not in POSIX.
+		const char* startq = last;
+		int found = 0;
```
tdulcet:
```suggestion
		bool found = FALSE;
```

src/util.h
```diff
@@ -26,6 +26,7 @@
 #ifndef util_h_included
 #define util_h_included
 
+#include "config.h"
```
tdulcet: This can be removed, as the `config.h` file no longer exists. Same in `.gitignore` above.

tdulcet (commented):
This looks good to me. We will just need to manually test it with all variations of the optional values for both the `PRP=`/`PRPDC=` and `Pminus1=` assignment formats.

src/Mlucas.c
tdulcet: Yes, Ernst had a custom boolean type, but I changed it to use `stdbool.h` in #22, since this is part of C99.

PR #56, opened 2025-11-04
Parallel lto codegen

Artoria2e5:
Lto was turned on ten months ago but with single-thread codegen, which made builds much slower than before. At a little cost to code quality (due to partitioning) we can get back much of the speed by enabling parallel codegen.

Yes, my copy of Apple Clang understands what this is, though it does not partition in this case. Clang wants a specific -flto-partitions=N flag, since it interprets -flto=auto as -flto=full, but then GCC wouldn't know what flag that is.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,7 @@ jobs:
     - name: Script
       run: |
         set -x 
+        export XCFLAGS=-fsanitize=address,undefined
         bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
@@ -182,6 +183,7 @@ jobs:
     - name: Script
       run: |
         set -x
+        export XCFLAGS=-fsanitize=thread
         bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
@@ -380,6 +382,7 @@ jobs:
     - name: Script
       run: |
         set -x
+        export XCFLAGS=-fsanitize=address,undefined
         bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
@@ -405,6 +408,7 @@ jobs:
     - name: Script
       run: |
         set -x
+        export XCFLAGS=-fsanitize=thread
         bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
@@ -519,6 +523,7 @@ jobs:
       shell: bash
       run: |
         set -x
+        export XCFLAGS=-fsanitize=address,undefined
         bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,8 +146,9 @@ jobs:
         sudo apt-get update -y
         sudo apt-get install -y libhwloc-dev
     - name: Before script
+      # With [^\s]* we also remove possible -flto= stuff
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        sed -i 's/-O3 -flto[^\s]*/-Og -fsanitize=address,undefined/' makemake.sh
         $CC --version
     - name: Script
       run: |
@@ -179,7 +180,7 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=thread/' makemake.sh
+        sed -i 's/-O3 -flto[^\s]*/-Og -fsanitize=thread/' makemake.sh
         $CC --version
     - name: Script
       run: |
@@ -378,7 +379,7 @@ jobs:
         brew install hwloc
     - name: Before script
       run: |
-        sed -i '' 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        sed -i '' 's/-O3 -flto[^\s]*/-Og -fsanitize=address,undefined/' makemake.sh
         clang --version
     - name: Script
       run: |
@@ -404,7 +405,7 @@ jobs:
         brew install hwloc
     - name: Before script
       run: |
-        sed -i '' 's/-O3 -flto/-Og -fsanitize=thread/' makemake.sh
+        sed -i '' 's/-O3 -flto[^\s]*/-Og -fsanitize=thread/' makemake.sh
         clang --version
     - name: Script
       run: |
@@ -518,7 +519,7 @@ jobs:
     - name: Before script
       shell: bash
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        sed -i 's/-O3 -flto[^\s]*/-Og -fsanitize=address,undefined/' makemake.sh
         $CC --version
     - name: Script
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,14 +146,12 @@ jobs:
         sudo apt-get update -y
         sudo apt-get install -y libhwloc-dev
     - name: Before script
-      # With [^\s]* we also remove possible -flto= stuff
       run: |
-        sed -i 's/-O3 -flto[^[:space:]]*/-Og -fsanitize=address,undefined/' makemake.sh
         $CC --version
     - name: Script
       run: |
-        set -x
-        bash -e -o pipefail -- makemake.sh use_hwloc
+        set -x 
+        bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
         time ./Mlucas -s m -cpu "0:$(( $(nproc --all) - 1 ))"
@@ -180,12 +178,11 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        sed -i 's/-O3 -flto[^\s]*/-Og -fsanitize=thread/' makemake.sh
         $CC --version
     - name: Script
       run: |
         set -x
-        bash -e -o pipefail -- makemake.sh use_hwloc
+        bash -e -o pipefail -- makemake.sh use_hwloc debug
         cd obj
         make clean
         time ./Mlucas -s m -cpu "0:$(( $(nproc --all) - 1 ))"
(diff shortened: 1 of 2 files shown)


makemake.sh
```diff
@@ -370,7 +370,7 @@ fi
 # stack trace of the issue. If one wishes, one can run 'strip -g Mlucas' to remove the debugging symbols:
 cat <<EOF >Makefile
 CC ?= gcc
-CFLAGS = -fdiagnostics-color -Wall -g -O3 -flto # =auto
+CFLAGS = -fdiagnostics-color -Wall -g -O3 -flto=auto
```
tdulcet: The reason I commented this out is that older versions of Clang do not support `-flto=auto`. We would either need to pass an explicit number of threads or make the `=auto` conditional if supported by the compiler.

.github/workflows/ci.yml
```diff
@@ -146,8 +146,9 @@ jobs:
         sudo apt-get update -y
         sudo apt-get install -y libhwloc-dev
     - name: Before script
+      # With [^\s]* we also remove possible -flto= stuff
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        sed -i 's/-O3 -flto[^\s]*/-Og -fsanitize=address,undefined/' makemake.sh
```
tdulcet: Backslash character classes like `\s` are not portable. You would need to use POSIX character classes instead:
```suggestion
        sed -i 's/-O3 -flto[^[:space:]]*/-Og -fsanitize=address,undefined/' makemake.sh
```

tdulcet (changes requested):
Thanks for the PR.

makemake.sh
Artoria2e5: Personally I think we should bump the minimum compiler version up a bit to also help with the ancient mac `-march=native` issue. But in this case though we do know the processor count for setting `-j` from before, so might as well do the core count.

Artoria2e5: Heck, since we are generating the Makefile anew anyways, we can do some shell magic to figure out what CC it is and talk appropriately to it. No reliance on properties makefile "if" extensions, just shell. (Yeah I failed to fall asleep oops

makemake.sh
```diff
@@ -78,25 +89,37 @@ if ! command -v $MAKE >/dev/null; then
...
 
+MAKE_ARGS+=(-j "$CPU_THREADS")
+
+CCVER=$($CC --version | head -n1)
+if [[ $CCVER == *"clang"* ]]; then
+	echo "Clang detected as C compiler. Enabling ThinLTO for multithreaded codegen."
+	C_ARGS_LTO+=(-flto=thin)
+elif [[ $CCVER == *"gcc"* ]]; then
+	echo "GCC detected as C compiler. Enabling multithreaded LTO."
+	C_ARGS_LTO+=(-flto="$CPU_THREADS")
```
tdulcet: I do not believe we would want ThinLTO for Clang, as it provides less performance. Also, `-flto=auto` should work with GCC:
```suggestion
if [[ $CCVER == *clang* ]]; then
	echo "Clang detected as C compiler. Enabling multithreaded LTO."
	C_ARGS_LTO+=(-flto="$CPU_THREADS")
elif [[ $CCVER == *gcc* ]]; then
	echo "GCC detected as C compiler. Enabling multithreaded LTO."
	C_ARGS_LTO+=(-flto=auto)
```

makemake.sh
```diff
@@ -78,25 +89,37 @@ if ! command -v $MAKE >/dev/null; then
...
 if [[ ! $OSTYPE == darwin* ]]; then
 	MAKE_ARGS+=(-O)
 	LD_ARGS+=(-lm -lpthread)
 	if [[ $OSTYPE != msys && $OSTYPE != cygwin ]]; then
 		LD_ARGS+=(-lrt)
 	fi
+	# Only supported on ELF platforms, but is harmless on not-mac since the compiler is not ancient:
+	CC_ARGS+=(-gz)
+	# Could make output smaller
+	LD_ARGS+=("-Wl,-O1")
```
tdulcet: Could you explain more about what you are trying to achieve with these new options? If binary size in a concern, we could make the `-g` option optional, but it is helpful for debugging a crash.

makemake.sh
```diff
@@ -106,30 +129,33 @@ echo "Total number of input parameters = $#"
 # order fashion, [details snipped]
 arglist=("$@") # Local array into which we copy cmd-line args in order to be able to manipulate them
 for i in "${!arglist[@]}"; do
-	echo "Arg[$i] = ${arglist[i]}"
+	printf "Arg[$i] = %q\n" "${arglist[i]}"
```
tdulcet: For consistency:
```suggestion
	printf 'Arg[%s] = %q\n' "$i" "${arglist[i]}"
```

tdulcet (changes requested):

makemake.sh
```diff
@@ -34,16 +34,24 @@ shopt -s nocasematch
...
 DIR=obj
 Mlucas=Mlucas
 Mfactor=Mfactor
 TARGET=$Mlucas
-ARGS=(-DUSE_THREADS) # Optional compile args
+
+# Optional compile [optimize] args
+C_ARGS=(-fdiagnostics-color -Wall -g)
+C_ARGS_OPT=(-g -O3)
+C_ARGS_DEBUG=(-g -Og '-fsanitize=address,undefined')
```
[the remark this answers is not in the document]
Artoria2e5: Might be more appropriate to add a positional argument for sanitizer then.

makemake.sh
Artoria2e5: As for why not =n on clang: clang 8.4.0 has =thin but not =n, nor any partitioning options. Yes I'm for nudging the user towards getting a newer clang, but I'm also a bit of a coward…

makemake.sh
Artoria2e5: Good idea!

makemake.sh
tdulcet: If you find anything that is not valid C99 (with the GNU extensions) on Unix, it probably should be considered a bug. All of the dependencies (GMP, hwloc, pthreads) should already be optional, although the CI does not test every possible build combination.

The `makemake.sh` script was originally written by me to make it easier to build Mlucas on Linux. Ernst added macOS support and most of the comments. I later added Mfactor and Windows support. I also added that `-fdiagnostics-color` compiler option to make the output easier to read, since there can be thousands of compiler warnings. If you find a version of GCC or Clang that does not support this option, we could remove it.

Artoria2e5: Testing with an older compiler hits about 90% of the issues we want to test with an older OS, save for actual system-level bugs like... I don't know, wrong handling of widened vector registers, but the compiler and assembler are not going to understand the new instruction sets anyways. 10-year-old compiler binaries do largely work on current OS.

That said, I don't expect something that has worked in an old version and a current version to start failing somewhere in between. It *could* happen, but it would likely not be our fault but someone else's bug.

In the case of mac things: apple rightfully gets a lot of flak for comparmentalizing everything, but in this case it also makes the solution for getting an old compiler simple: grab an old xcode dmg, extract the `.app`, and use the SDK prefix from within. https://xcodereleases.com/?q=10.9 takes you to a version from 2015 (XCode 6.2).

(When I had my "core2duo mac" I just did `ports install gcc` and got the new compiler. When clang became established I did the same. But welp, macports does not support macOS 10.7 any more, I think.)

* * *
About the job-count issue: I desire a two-tiered CI check: an abbreviated one [without needing approval] for quick checks (perhaps just 5-year-old Linux, current Linux, Windows, Mac) and the full thing. The free GitHub action runner only has this much job concurrency and wait times have become a bit too high. I notice that early-fail is disabled; perhaps it should be on for the abbreviated version (unless it's disabled for log-preserving reasons, then I say we need to make our own early-fail logic).

tdulcet: Interesting, I do not have access to any macOS systems myself, so I rely entirely on the CI for testing.

We could create a two tiered CI check, by making jobs [dependent on other jobs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs). In the successful case, it may make the CI take longer since fewer jobs could run in parallel, but it would allow the CI to fail quicker.

The approval is required by GitHub and I do not believe there is anything we could do about that. However, once your first PR is merged and you are marked as a "contributor" for this repository, it should no longer be required.

The fast fail feature is disabled because it can be useful to see which combinations of OS, architecture and compiler are affected by an issue, to reduce the amount of manual debugging required. Some of the crashes/seg faults can also be intermittent or nondeterministic.

makemake.sh
```diff
@@ -34,16 +34,26 @@ shopt -s nocasematch
...
 Mfactor=Mfactor
 TARGET=$Mlucas
-ARGS=(-DUSE_THREADS) # Optional compile args
+
+# Optional compile [optimize] args
+C_ARGS=(-fdiagnostics-color -Wall)
+C_ARGS_OPT=(-g -O3)
+C_ARGS_DEBUG=(-g -Og)
+mapfile -t C_ARGS_ADD <<< "$XCFLAGS"
+C_ARGS+=("${C_ARGS_ADD[@]}")
```
tdulcet: `XCFLAGS` is only a single line:
```suggestion
C_ARGS+=("$XCFLAGS")
```

makemake.sh
```diff
@@ -346,7 +386,8 @@ EOF
 		echo "Error: Unable to detect the SIMD build mode" >&2
 		exit 1
 	fi
-	ARGS+=($output)
+	mapfile -t output_array <<<"$output"
+	C_ARGS+=("${output_array[@]}")
```
tdulcet: This is not equivalent, as `mapfile` splits by newlines, but this was splitting by all whitespace. Since we are pasting the value in the Makefile anyway, it should be fine to just quote the variable and disable the splitting, as I suggested above.

makemake.sh
```diff
@@ -78,25 +91,37 @@ if ! command -v $MAKE >/dev/null; then
...
+fi
+
 if [[ ! $OSTYPE == darwin* ]]; then
 	MAKE_ARGS+=(-O)
 	LD_ARGS+=(-lm -lpthread)
 	if [[ $OSTYPE != msys && $OSTYPE != cygwin ]]; then
 		LD_ARGS+=(-lrt)
 	fi
+	# Only supported on ELF platforms, but is harmless on not-mac since the compiler is not ancient:
+	CC_ARGS+=(-gz)
```
tdulcet: Did you mean:
```suggestion
	C_ARGS+=(-gz)
```

tdulcet (changes requested):

MarkRose:
Small thing spotted while auditing open PRs for description/code drift — `makemake.sh:121`:

```sh
	# Only supported on ELF platforms, but is harmless on not-mac since the compiler is not ancient:
	CC_ARGS+=(-gz)
```

That's the only `CC_ARGS` in the file; every other flag array in this PR is `C_ARGS` (21 uses). Since bash creates an array on first `+=`, there's no error — `-gz` just accumulates into a variable nothing ever expands, so compressed debug sections are silently not enabled. Looks like `C_ARGS+=(-gz)` was intended.

Worth mentioning that the surrounding comment may want a second look too: `-gz` defaults to `-gz=zlib`, which needs the assembler and linker to agree on the compression, and Apple's toolchain doesn't accept ELF-compressed sections at all. The `if` this sits in is the non-msys/cygwin branch, so macOS reaches it — if `-gz` does get wired up, that path may need excluding.

Not blocking anything, and no action needed from me — just didn't want it to land silently as a no-op.

PR #64, opened 2026-06-13
Fixed GCC 15 support and added support for musl

tdulcet:
* Added support for GCC 15 on x86 by fixing incompatible pointer type errors.
  * This was contributed by @ldesnogu and cherry picked from #34.
* Added support for the musl C standard library implementation.
  * Enabled the CI to test building on Alpine Linux, which uses musl.


ldesnogu (commented):
That looks good to me, thanks!

Issue #65, opened 2026-06-16
Segmentation fault with Clang 18+/GCC 15+ and AVX512


tdulcet:
CI runs on Ubuntu 24.04 and 26.04 using Clang 18 and 21 respectfully with Intel CPUs supporting AVX512 are consistently seg faulting. For example:
```
INFO: Maximum recommended exponent for FFT length (8 Kdbl) = 174264; p[ = 173431]/pmax_rec = 0.9952198962.
Initial DWT-multipliers chain length = [hiacc] in carry step.
M173431: using FFT length 8K = 8192 8-byte floats, initial residue shift count = 73199
This gives an average   21.170776367187500 bits per digit
Using complex FFT radices         8        16        32
/home/runner/work/_temp/72235bae-f826-4c68-8a91-382ab47e6682.sh: line 30:  8262 Segmentation fault         (core dumped) ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" 2>&1
```
Run on Ubuntu 22.04 using Clang 14 are unaffected. AVX2 builds also seem to be unaffected.

CI runs on Ubuntu 26.04 using GCC 15 with AVX512 are also occasionally seg faulting. For example:
```
INFO: Maximum recommended exponent for FFT length (128 Kdbl) = 2628150; p[ = 2614999]/pmax_rec = 0.9949960999.
Initial DWT-multipliers chain length = [hiacc] in carry step.
M2614999: using FFT length 128K = 131072 8-byte floats, initial residue shift count = 1944584
This gives an average   19.950859069824219 bits per digit
Using complex FFT radices       128        16        32
Using 4 threads in carry step
/home/runner/work/_temp/0a9bab47-fc6d-424b-9ef0-63b7557d0a81.sh: line 30:  9894 Segmentation fault         (core dumped) ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" 2>&1
```

We need help from someone with an AVX512 system to run this under GDB to determine where specifically the problem is. It could be related to #7 or #16.

ldesnogu:
I wonder if the old warning branch works with the CI (and it can be rebased). Did you give it a try?

tdulcet:
> I wonder if the old warning branch works with the CI (and it can be rebased). Did you give it a try?

Yes, I wonder that as well. We could give it a try, but first we would need to merge #64 and then rebase #34.

[...]

PR #67, opened 2026-07-09
Probe toolchain capabilities instead of hardcoding version cutoffs (#60)

MarkRose:
## Summary

resolves #60

Replaces several hardcoded compiler/tool version-cutoff assumptions in `makemake.sh` with actual capability probes, fixes a new build break on current glibc, and updates `ci.yml` to match.

- **New build break fixed:** `CPU_ZERO`/`CPU_SET`/`sched_setaffinity` failed to declare on glibc 2.43/GCC 16.1.1 (current Arch/EndeavourOS) — the in-file `#define _GNU_SOURCE` before `<sched.h>` in `platform.h` no longer suffices, because some earlier header in the chain now locks in `__USE_GNU=0` first. Same root cause Ernst documented from Fedora 16 in the comment right above that line, and his own note there already identified the fix: define `_GNU_SOURCE` on the compile command line. Wired that up via `CPPFLAGS` in `makemake.sh`, and guarded the in-file `#define` so it no longer emits a redundant "redefined" warning.
- `-fdiagnostics-color`: now probed via a `try_flag()` compile-and-link helper instead of assumed.
- `-flto`: probed via a genuine two-object compile-and-link (`try_lto()`) rather than a single-file probe — LTO breakage on old/flaky toolchains (old binutils, mismatched ar/nm/ranlib plugin support) only shows up once the linker actually combines LTO objects from more than one translation unit, which is exactly what building Mlucas's ~90 source files does.
- GNU Make's `-O` flag: gated on the actual Make version (>=4) rather than excluding only macOS.
- AVX-512 build mode: gated on a probe that the assembler accepts the "extended" register names (zmm16-31/xmm16-31/k0-7), since some older Clang releases reject them even when otherwise AVX-512-aware. Hard error for an explicit `avx512` request; silent fallback to AVX2 in auto-detect.
- ARM ASIMD: probes `-mcpu=native` support, falling back to `-march=native` for older Clang aarch64 toolchains.
- hwloc `< 1.11`: shims `HWLOC_OBJ_PACKAGE` to the older `HWLOC_OBJ_SOCKET` name.

**`ci.yml` updated to match:** the `-fdiagnostics-color`/`-flto` probing turns `CFLAGS` from a static string into a computed value, which broke every CI step that `sed`/`-replace`-patched that literal string:
- 2 old-toolchain LTO-strip compat hacks are now obsolete (the probe self-detects) — deleted.
- 5 sanitizer-build jobs (ASan/TSan on Linux/macOS/Windows) swapped in `-Og -fsanitize=...` via the same mechanism — these were silently no-op'ing, so those jobs were reporting green while no longer running under any sanitizer at all. Replaced with a `CFLAGS` environment-variable override: the generated Makefile already uses `CFLAGS ?= ...`, which Make defers to a pre-set environment variable — that's the actual override mechanism the script was designed around (every other var follows the same `?=` pattern), not a new workaround.

Also removed two long-dead commented-out code blocks adjacent to the edits (a superseded SIMD-detection loop, an unused build.log-backup snippet).

## Not included (flagging rather than guessing)

- #60's GCC 5.4 AVX2 "impossible constraints" sub-issue is not addressed — needs the actual CI build.log to pin the failing macro, or an old-GCC repro environment, neither of which I had access to.
- The AVX-512 clobber-list segfault family (#65/#52/#16, tracked separately) is out of scope here — we're deferring that until AVX-512 test hardware is available.

## Test plan

Verified locally (Intel i9-10885H, AVX2, glibc 2.43/gcc 16.1.1 — no AVX-512 hardware for that portion):
- [x] Default auto-detect build, explicit `avx512` (compile-only), and `use_hwloc` — all build clean, 0 errors; `-s tt` self-test passes
- [x] `try_lto()` and the `CFLAGS ?=` environment-override mechanism verified directly (a pre-set `CFLAGS` env var correctly takes precedence over the computed default)
- [x] `ci.yml` re-validated as syntactically correct YAML after edits

CI on this PR caught real issues during review (confirmed against the last green run on `main` for comparison):
- Confirmed pre-existing, unrelated to this PR: `Windows (windows-11-arm, gcc)`, `Windows (windows-latest, clang)`, `Windows (windows-latest, gcc)` — already fail on `main`.
- Regressions from the CFLAGS refactor (now fixed by the `ci.yml` update in this PR): `AddressSanitizer Windows`, `Mlucas Linux old (ubuntu-24.04-arm, alpine:3.12, clang)`, and the silent sanitizer-coverage loss on `AddressSanitizer Linux (gcc/clang)`, `AddressSanitizer macOS`, `ThreadSanitizer Linux (clang)`.
- `Mlucas Linux (ubuntu-26.04, clang)` hit the pre-existing, already-tracked Task 1 AVX-512 clobber-list segfault (#65/#52/#16) — passes on `main`, but that bug is known to be compiler/register-allocation-sensitive, and adding `-D_GNU_SOURCE` may have shifted codegen enough to newly trigger it here. Out of scope for this PR; will be addressed separately once AVX-512 test hardware is available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,10 +177,6 @@ jobs:
       run: |
         $CC --version
         ldd --version || true
-    - name: Fix
-      if: ${{ (matrix.container == 'ubuntu:16.04' || matrix.container == 'alpine:3.8' || matrix.container == 'alpine:3.12' || matrix.container == 'alpine:3.16') && matrix.cc == 'clang' }}
-      run: |
-        sed -i 's/ -flto//' makemake.sh
     - name: Script
       shell: bash
       run: |
@@ -224,7 +220,7 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=address,undefined" >> "$GITHUB_ENV"
         $CC --version
     - name: Script
       run: |
@@ -256,7 +252,7 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=thread/' makemake.sh
+        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=thread" >> "$GITHUB_ENV"
         $CC --version
     - name: Script
       run: |
@@ -455,7 +451,7 @@ jobs:
         brew install hwloc
     - name: Before script
       run: |
-        sed -i '' 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=address,undefined" >> "$GITHUB_ENV"
         clang --version
     - name: Script
       run: |
@@ -481,7 +477,7 @@ jobs:
         brew install hwloc
     - name: Before script
       run: |
-        sed -i '' 's/-O3 -flto/-Og -fsanitize=thread/' makemake.sh
+        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=thread" >> "$GITHUB_ENV"
         clang --version
     - name: Script
       run: |
@@ -525,7 +521,6 @@ jobs:
       if: ${{ matrix.cc == 'clang' }}
       run: |
         pacman -S --noconfirm "${env:PACKAGE_PREFIX}clang"
-        (Get-Content makemake.sh) -replace ' -flto', '' | Set-Content makemake.sh
     - name: Before script
       shell: bash
       run: |
@@ -595,7 +590,7 @@ jobs:
     - name: Before script
       shell: bash
       run: |
-        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
+        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=address,undefined" >> "$GITHUB_ENV"
         $CC --version
     - name: Script
       shell: bash
--- a/makemake.sh
+++ b/makemake.sh
@@ -88,13 +88,11 @@ fi
 # Windows (MinGW) gcc does not understand the POSIX path /dev/null as an -o target (it becomes a
 # nonexistent C:\dev\null), so the /dev/null form spuriously fails every probe on that toolchain:
 try_flag() {
-	local tmpdir rc
+	local tmpdir
 	tmpdir=$(mktemp -d) || return 1
+	trap 'rm -rf "$tmpdir"' RETURN
 	printf 'int main(void){return 0;}\n' >"$tmpdir/t.c"
 	"${CC:-gcc}" "$@" "$tmpdir/t.c" -o "$tmpdir/t.out" >/dev/null 2>&1
-	rc=$?
-	rm -rf "$tmpdir"
-	return $rc
 }
 
 # Returns success iff $CC's assembler accepts the AVX-512 constructs Mlucas's inline-asm kernels use:
@@ -103,8 +101,9 @@ try_flag() {
 # reject one or both even when otherwise AVX-512-aware - e.g. clang 3.8 rejects the extended register
 # names, clang 5.0 assembles those but rejects vrcp28pd - so probe both, exactly as the CI does (#73):
 try_avx512_asm() {
-	local tmpdir rc
+	local tmpdir
 	tmpdir=$(mktemp -d) || return 1
+	trap 'rm -rf "$tmpdir"' RETURN
 	cat >"$tmpdir/t.c" <<'EOF'
 int main(void)
 {
@@ -122,9 +121,6 @@ EOF
 	# toolchain that genuinely can't assemble these constructs still fails; but compiling a real file to a
 	# real .o avoids the MinGW '-o /dev/null' pitfall that made this probe a false-negative on Windows:
 	"${CC:-gcc}" -mavx512f -c "$tmpdir/t.c" -o "$tmpdir/t.o" >/dev/null 2>&1
-	rc=$?
-	rm -rf "$tmpdir"
-	return $rc
 }
 
 # Returns success iff $CC can compile two separate translation units with the given -flto[=...] variant
@@ -146,18 +142,17 @@ try_lto() {
 	) >/dev/null 2>&1
 }
 
-# GNU Make's -O (synchronize parallel-job output) flag needs Make >= 4.0 - probe for the flag itself
-# rather than assuming by version number (which drifts, and varies by distro/backport):
-MAKE_SUPPORTS_dashO=0
-"$MAKE" --help 2>/dev/null | grep -wq -- '-O' && MAKE_SUPPORTS_dashO=1
-
 if [[ ! $OSTYPE == darwin* ]]; then
 	LD_ARGS+=(-lm -lpthread)
 	if [[ $OSTYPE != msys && $OSTYPE != cygwin ]]; then
 		LD_ARGS+=(-lrt)
 	fi
 fi
-((MAKE_SUPPORTS_dashO)) && MAKE_ARGS+=(-O)
+# GNU Make's -O (synchronize parallel-job output) flag needs Make >= 4.0 - probe for the flag itself
+# rather than assuming by version number (which drifts, and varies by distro/backport):
+if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
+	MAKE_ARGS+=(-O)
+fi
 MAKE_ARGS+=(-j "$CPU_THREADS")
 
 # $0 contains script-name, but $@ starts with first ensuing cmd-line arg, if it exists:
@@ -314,7 +309,7 @@ elif [[ $OSTYPE == darwin* ]]; then
 	if (($(sysctl -n hw.optional.avx512f))) && try_avx512_asm; then
 		echo -e "The CPU supports the AVX512 SIMD build mode.\n"
 		ARGS+=(-DUSE_AVX512 -march=native -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
-	elif (($(sysctl -n hw.optional.avx512f))) || (($(sysctl -n hw.optional.avx2_0))); then
+	elif (($(sysctl -n hw.optional.avx512f) || $(sysctl -n hw.optional.avx2_0))); then
 		if (($(sysctl -n hw.optional.avx512f))); then
 			echo "Warning: CPU supports AVX-512 but ${CC:-gcc}'s assembler rejects the extended register names needed ... falling back to AVX2." >&2
 		fi
@@ -474,6 +469,15 @@ EOF
 				ARGS+=(-march=native)
 			fi
 			;;
+		none_x86)
+			echo -e "The CPU supports no Mlucas-recognized SIMD build mode ... building in scalar-double mode.\n"
+			echo "Warning: This is a 64-bit x86 system without even SSE2, which likely means there is a bug in this script. Please report!" >&2
+			ARGS+=(-march=native)
+			;;
+		none)
+			echo -e "The CPU architecture is not recognized by this script ... building in scalar-double mode.\n"
+			try_flag -march=native && ARGS+=(-march=native)
+			;;
 		*)
 			echo -e "The CPU supports no Mlucas-recognized SIMD build mode ... building in scalar-double mode.\n"
 			echo "Warning: If this is a 64-bit x86 or ARM system, this likely means there is a bug in this script. Please report!" >&2
@@ -502,19 +506,12 @@ fi
 # defers to a pre-set environment CFLAGS instead of the computed value below. Prefer -flto=auto (parallel
 # LTO codegen, see #56) over plain -flto when supported:
 CFLAGS_PROBED=(-Wall -g -O3)
-# The Mlucas sources use C99 features (for-loop-scope declarations, mixed declarations-and-code). Old
-# gcc (e.g. 4.8 on Ubuntu 14.04) defaults to gnu89/gnu90 and rejects these with a hard error; newer
-# toolchains default to gnu11+ and accept them. Probe the compiler's default and request -std=gnu99 only
-# when the default won't compile a C99 for-loop-scope declaration, so modern builds keep their default:
-c99_probe() {	# $1: optional -std flag
-	local tmpdir rc; tmpdir=$(mktemp -d) || return 1
-	printf 'int main(void){for(int i=0;i<1;++i){int j=i;(void)j;}return 0;}\n' >"$tmpdir/t.c"
-	"${CC:-gcc}" ${1:+"$1"} -c "$tmpdir/t.c" -o "$tmpdir/t.o" >/dev/null 2>&1; rc=$?
-	rm -rf "$tmpdir"; return $rc
-}
-if ! c99_probe && c99_probe -std=gnu99; then
-	CFLAGS_PROBED+=(-std=gnu99)
-fi
+# The Mlucas sources use C99 features (for-loop-scope declarations, mixed declarations-and-code) plus GNU
+# extensions (statement expressions in the checked-alloc macros). Old gcc (e.g. 4.8 on Ubuntu 14.04)
+# defaults to gnu89/gnu90 and rejects the C99 constructs with a hard error, so request -std=gnu99
+# unconditionally: every toolchain then compiles the same dialect, and gnu99 (vs plain c99) keeps the GNU
+# extensions enabled. -D_GNU_SOURCE (set in CPPFLAGS below) still exposes the POSIX/GNU library surface:
+CFLAGS_PROBED+=(-std=gnu99)
 try_flag -fdiagnostics-color && CFLAGS_PROBED=(-fdiagnostics-color "${CFLAGS_PROBED[@]}")
 if try_lto -flto=auto; then
 	CFLAGS_PROBED+=(-flto=auto)
--- a/makemake.sh
+++ b/makemake.sh
@@ -83,9 +83,18 @@ fi
 
 # Returns success iff $CC (default gcc) accepts the given flag(s) for a full compile-and-link of a
 # trivial program - used below to auto-detect toolchain-version-dependent flag/feature support instead
-# of hardcoding version-number cutoffs (which drift out of date and vary by distro/backport):
+# of hardcoding version-number cutoffs (which drift out of date and vary by distro/backport).
+# Compile a real temp file to a real output rather than piping stdin to '-o /dev/null': a native
+# Windows (MinGW) gcc does not understand the POSIX path /dev/null as an -o target (it becomes a
+# nonexistent C:\dev\null), so the /dev/null form spuriously fails every probe on that toolchain:
 try_flag() {
-	printf 'int main(void){return 0;}\n' | "${CC:-gcc}" "$@" -x c -o /dev/null - >/dev/null 2>&1
+	local tmpdir rc
+	tmpdir=$(mktemp -d) || return 1
+	printf 'int main(void){return 0;}\n' >"$tmpdir/t.c"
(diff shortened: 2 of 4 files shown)


makemake.sh
```diff
@@ -89,13 +81,56 @@ elif ! command -v gcc >/dev/null; then
...
+# of hardcoding version-number cutoffs (which drift out of date and vary by distro/backport):
+try_flag() {
+	printf 'int main(void){return 0;}\n' | "${CC:-gcc}" "$@" -x c -o /dev/null - >/dev/null 2>&1
+}
+
+# Returns success iff $CC's assembler accepts the AVX-512 "extended" register names (zmm16-31,
+# xmm16-31, k0-k7) used throughout the AVX-512 inline-asm kernels. Some older Clang releases (pre-9ish)
+# reject these names even when otherwise AVX-512-aware:
+try_avx512_asm() {
+	printf 'int main(void){__asm__ __volatile__("vpxord %%%%zmm31,%%%%zmm31,%%%%zmm31\\n\\tvmovdqa64 %%%%xmm16,%%%%xmm17\\n\\tkmovw %%%%k1,%%%%eax" ::: "zmm31","xmm16","xmm17","k1","eax"); return 0;}\n' \
```
tdulcet: Maybe put this in a Bash heredoc for readability and to avoid needing to escape the percent characters and backslashes.

makemake.sh
```diff
@@ -89,13 +81,56 @@ elif ! command -v gcc >/dev/null; then
...
+	printf 'int mm_lto_probe_helper(void);\nint main(void){return mm_lto_probe_helper();}\n' >"$tmpdir/b.c"
+	(
+		cd "$tmpdir" && \
+		"$cc" -flto -c a.c -o a.o && \
+		"$cc" -flto -c b.c -o b.o && \
+		"$cc" -flto a.o b.o -o out
+	) >/dev/null 2>&1
+	local rc=$?
+	rm -rf "$tmpdir"
+	return $rc
```
tdulcet: You could replace this with an exit handler at the beginning of the function, after the `mktemp` call:
```bash
trap 'rm -rf "$tmpdir"' RETURN
```

makemake.sh
```diff
@@ -89,13 +81,56 @@ elif ! command -v gcc >/dev/null; then
...
+}
+
+# GNU Make's -O (synchronize parallel-job output) flag needs Make >= 4.0 - probe the actual version
+# rather than assuming by OS/distro (e.g. Ubuntu 14.04 and macOS's bundled Make both predate it):
+MAKE_SUPPORTS_dashO=0
+if make_ver_line=$("$MAKE" --version 2>/dev/null | head -n1) && [[ $make_ver_line =~ ([0-9]+)\.[0-9]+(\.[0-9]+)?[[:space:]]*$ ]]; then
+	(( ${BASH_REMATCH[1]} >= 4 )) && MAKE_SUPPORTS_dashO=1
+fi
```
tdulcet: Per the title of the PR about hardcoding version cutoffs, it would be simpler and more robust to just check if the `-O` option exists in Make's `--help` output:
```bash
if "$MAKE" --help | grep -wq -- '-O'; then
    MAKE_ARGS+=(-O)
fi
```

makemake.sh
```diff
@@ -256,7 +302,11 @@ elif [[ $OSTYPE == darwin* ]]; then
 		ARGS+=(-DUSE_SSE2 -march=core2 -msse2)
 	elif (($(sysctl -n hw.optional.neon))); then
 		echo -e "The CPU supports the ASIMD build mode.\n"
-		ARGS+=(-DUSE_ARM_V8_SIMD -mcpu=native) # -march=native
+		if try_flag -mcpu=native; then
+			ARGS+=(-DUSE_ARM_V8_SIMD -mcpu=native)
+		else
+			ARGS+=(-DUSE_ARM_V8_SIMD -march=native)
+		fi
```
tdulcet, line 340:
```suggestion
		ARGS+=(-DUSE_ARM_V8_SIMD)
		if try_flag -mcpu=native; then
			ARGS+=(-mcpu=native)
		else
			ARGS+=(-march=native)
		fi
```
The same simplification could be made below.

makemake.sh
```diff
@@ -266,10 +316,13 @@ elif [[ $OSTYPE == darwin* ]]; then
 elif [[ $OSTYPE == linux* ]]; then
 
 	# Linux:
-	if grep -iq 'avx512' /proc/cpuinfo; then
+	if grep -iq 'avx512' /proc/cpuinfo && try_avx512_asm; then
 		echo -e "The CPU supports the AVX512 SIMD build mode.\n"
 		ARGS+=(-DUSE_AVX512 -march=native -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
-	elif grep -iq 'avx2' /proc/cpuinfo; then
+	elif grep -iq 'avx512' /proc/cpuinfo || grep -iq 'avx2' /proc/cpuinfo; then
```
tdulcet:
```suggestion
	elif grep -iq 'avx512\|avx2' /proc/cpuinfo; then
```

makemake.sh
```diff
@@ -370,8 +440,8 @@ fi
 # stack trace of the issue. If one wishes, one can run 'strip -g Mlucas' to remove the debugging symbols:
 cat <<EOF >Makefile
 CC ?= gcc
-CFLAGS ?= -fdiagnostics-color -Wall -g -O3 -flto # =auto
-CPPFLAGS ?= -I/usr/local/include -I/opt/homebrew/include
+CFLAGS ?= ${CFLAGS_PROBED[*]} # =auto
```
tdulcet: Maybe consider testing for `-flto=auto` support as well. For context, see #56. Regardless, it could now be removed from here:
```suggestion
CFLAGS ?= ${CFLAGS_PROBED[*]}
```

makemake.sh
```diff
@@ -236,15 +271,26 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 			;;
 	esac
 
+	# The AVX-512 kernels use "extended" register names (zmm16-31/xmm16-31/k0-7) that some older
+	# Clang releases reject even when otherwise AVX-512-aware; probe rather than let the user hit a
+	# wall of asm errors deep in the build:
+	if [[ $arg == avx512* ]] && ! try_avx512_asm; then
+		echo "Error: ${CC:-gcc}'s assembler does not support the AVX-512 extended register names (zmm16-31/xmm16-31/k0-7) needed for this build mode ... aborting. Try a newer compiler, or build with 'avx2' instead." >&2
+		exit 1
+	fi
```
tdulcet, line 302: It would not need to be in this PR, but we should also update the CI to skip the AVX512 builds in this case, as several jobs are now failing with this error.

tdulcet (changes requested):
Wow, thank you for all four of these PRs and looking into these long standing issues! It is a major contribution.

This PR looks good to me. I just have few minor suggestions for improvement. I would also like to request that @ldesnogu review the inline assembly in that one probe.

> GCC 5.4 AVX2 "impossible constraints" sub-issue is not addressed — needs the actual CI build.log to pin the failing macro, or an old-GCC repro environment, neither of which I had access to.

You should have access to everything that I do, but I would be happy to provide anything that is needed. From the CI log, the full error is:
```
../src/dft_macro.c:5050:4: error: 'asm' operand has impossible constraints
    SSE2_RADIX16_DIT_0TWIDDLE(
    ^
../src/dft_macro.c:5101:3: error: 'asm' operand has impossible constraints
   SSE2_RADIX16_DIT_0TWIDDLE(
   ^
```
If needed, it is easy to spin up a Docker container with one of these older Linux distributions.

makemake.sh
MarkRose: Done - switched to a heredoc. (Still needs the GCC-level `%%` escaping for the asm template itself, independent of the shell-quoting fix - fixed a self-inflicted regression from dropping that during the rewrite, caught by rebuilding locally before pushing.)

makemake.sh
MarkRose: Done - `trap 'rm -rf "$tmpdir"' RETURN` right after `mktemp`, function now just falls through to the subshell's exit status.

makemake.sh
MarkRose: Done - replaced the version-parsing with `"$MAKE" --help | grep -wq -- '-O'`. Simpler and matches the PR's own theme, thanks.

makemake.sh
MarkRose: Done in both the Darwin and Linux ASIMD branches.

makemake.sh
MarkRose: Done.

makemake.sh
MarkRose: Done on both - probing `-flto=auto` first (falling back to plain `-flto`, then to nothing) and dropped the stale `# =auto` comment.

makemake.sh
MarkRose: Agreed this would help, but leaving it out of this PR as you suggested - it's a `ci.yml`-only change orthogonal to the probing work here. For context in case it's picked up separately: the `windows-11-arm` runner doesn't take the `ASIMD`-only branch in the CI script's own `if [[ $HOSTTYPE == aarch64 ]]` check, so it still attempts the x86-only `AVX512` mode there and hits this. Worth checking why $HOSTTYPE isn't reporting aarch64 on that runner's bash, separately from whether the probe here should also gate it.

MarkRose: Checked - confirmed: `$HOSTTYPE` reports `x86_64` on that runner's bash even though the host is ARM64 (`+ [[ x86_64 == aarch64 ]]` in the trace). The installed MSYS2/bash there is itself an x86_64 build running under Windows-on-ARM emulation, so `$HOSTTYPE` reflects bash's own architecture, not the CPU's. A CI-side fix would need a different detection method there (e.g. `uname -m` via the underlying Windows arch, or $PROCESSOR_ARCHITECTURE).

MarkRose:
Thanks for the build.log excerpt, @tdulcet — that was exactly what was missing. Dug into it:

Both `dft_macro.c:5050`/`:5101` (your GCC 5.4 + AVX2 log, macro `SSE2_RADIX16_DIT_0TWIDDLE`, AVX2/FMA3 branch in `sse2_macro_gcc64.h:9638`) and a second instance we hit live in our own CI (`mi64.c:6629`, `sse2_macro_gcc64.h:9640`, under `-Og -fsanitize=address,undefined`) have the **same root cause**: both macros clobber *all 14* non-reserved x86_64 GPRs (`rax`-`r15` minus `rsp`/`rbp`), leaving GCC's register allocator zero spare registers to compute operand addresses or do reload bookkeeping. That's fine under a strong/modern allocator (any recent GCC/Clang at the normal `-O3 -flto`), but fails when register pressure is already elevated from another source - an old, weaker allocator (GCC 5.4) or ASan/UBSan's own instrumentation overhead.

This is already a known, documented GCC limitation in-tree twice (`carry_gcc64.h:22-31`, `radix16_dif_dit_pass_asm.h:29-36`, citing GCC bug 23200) with a "use -O1 or higher" workaround - but I confirmed empirically that workaround doesn't apply to the sanitizer case: `-O1` and `-O2` both still fail on a modern GCC 16 under `-fsanitize=address,undefined`, since ASan's shadow-memory checks are the actual pressure source, not the optimization level.

It's systemic rather than isolated: `sse2_macro_gcc64.h`'s AVX2/FMA3 `SSE2_RADIX16_DIT/DIF_0TWIDDLE`, `mi64.c`'s AVX2-MULX 4-way loops, and `radix11/13_sse_macro.h` all use the same all-14-GPR-clobber style, whereas `SSE2_RADIX8_DIF_0TWIDDLE` (smaller macro, same file) deliberately leaves 4 GPRs free - so a real fix (rebalancing clobber lists to leave the allocator some slack, the way the radix-8 macro already does) looks possible, but it's a nontrivial audit across the hand-tuned asm macro library - similar risk profile to the AVX-512 clobber work in #68, and out of scope for this build-portability PR. `mi64.c`'s asm path alone has an escape hatch (`-DNO_ASM` falls back to portable C, `mi64.h:37-45`), but the SIMD DFT macros have no non-asm fallback within the same build mode, so that only covers half of it.

I've logged this analysis in the campaign's planning doc so it's not lost; recommend tracking it as its own follow-up (either rebalance the clobber lists, or land on a documented minimum-GCC-version policy for AVX2+ builds) rather than folding it into this PR.

makemake.sh
tdulcet: Also see: https://github.com/primesearch/Mlucas/pull/45#issuecomment-3056938160

.github/workflows/ci.yml
tdulcet, line 139: We should update this `continue-on-error` line based on which jobs are no longer expected to fail. Alternatively, you could make this change in #73 to avoid merge conflicts.

tdulcet (changes requested):
Thanks for the detailed analysis and all of the changes. It looks like this is about ready to be merged.

You might want to consider making the improvements to the `makemake.sh` script to the Windows code path as well, when applicable. The current Windows CI failures should be fixed by #45 and #69.

.github/workflows/ci.yml
MarkRose: Several of the PRs would want to change that line, so it's a separate PR to be merged after.

MarkRose:
There is a separate PR for fixing the Windows builds, outside of the fixes that affect multiple OSes.

MarkRose:
Pushed a follow-up (7cf3f13) that applies the probing to the Windows/fallback detect path as suggested. That branch (`else`, used when there's no `/proc/cpuinfo` or `sysctl` — i.e. MSYS2/Cygwin) previously compiled a probe that emitted `-DUSE_AVX512 …` flags whenever the CPU advertised AVX-512, without ever running `try_avx512_asm`. That's the root cause of the Windows AVX-512 build failures noted in #83. The probe now reports only the CPU's highest SIMD level as a keyword and the shell maps it to flags, so the same `try_avx512_asm` (AVX-512→AVX2 fallback) and `try_flag` (-mcpu/-march) probes used by the Linux and Darwin branches now cover Windows too.

makemake.sh
```diff
@@ -89,13 +81,83 @@ elif ! command -v gcc >/dev/null; then
...
+# Windows (MinGW) gcc does not understand the POSIX path /dev/null as an -o target (it becomes a
+# nonexistent C:\dev\null), so the /dev/null form spuriously fails every probe on that toolchain:
+try_flag() {
+	local tmpdir rc
+	tmpdir=$(mktemp -d) || return 1
+	printf 'int main(void){return 0;}\n' >"$tmpdir/t.c"
+	"${CC:-gcc}" "$@" "$tmpdir/t.c" -o "$tmpdir/t.out" >/dev/null 2>&1
+	rc=$?
+	rm -rf "$tmpdir"
+	return $rc
```
tdulcet: This should use the same `trap 'rm -rf "$tmpdir"' RETURN` pattern as before. The same in the below functions.

makemake.sh
```diff
@@ -89,13 +81,83 @@ elif ! command -v gcc >/dev/null; then
...
+		"$cc" "$flag" -c a.c -o a.o && \
+		"$cc" "$flag" -c b.c -o b.o && \
+		"$cc" "$flag" a.o b.o -o out
+	) >/dev/null 2>&1
+}
+
+# GNU Make's -O (synchronize parallel-job output) flag needs Make >= 4.0 - probe for the flag itself
+# rather than assuming by version number (which drifts, and varies by distro/backport):
+MAKE_SUPPORTS_dashO=0
+"$MAKE" --help 2>/dev/null | grep -wq -- '-O' && MAKE_SUPPORTS_dashO=1
```
tdulcet: The `MAKE_SUPPORTS_dashO` variable is only used once, so this could be simplified to just set `MAKE_ARGS` directly:
```suggestion
if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
	MAKE_ARGS+=(-O)
fi
```

makemake.sh
```diff
@@ -236,15 +298,26 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
 
 elif [[ $OSTYPE == darwin* ]]; then
 
 	# MacOS:
-	if (($(sysctl -n hw.optional.avx512f))); then
+	if (($(sysctl -n hw.optional.avx512f))) && try_avx512_asm; then
 		echo -e "The CPU supports the AVX512 SIMD build mode.\n"
 		ARGS+=(-DUSE_AVX512 -march=native -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
-	elif (($(sysctl -n hw.optional.avx2_0))); then
+	elif (($(sysctl -n hw.optional.avx512f))) || (($(sysctl -n hw.optional.avx2_0))); then
```
tdulcet:
```suggestion
	elif (($(sysctl -n hw.optional.avx512f) || $(sysctl -n hw.optional.avx2_0))); then
```

makemake.sh
```diff
@@ -340,13 +425,61 @@ EOF
...
+				ARGS+=(-mcpu=native)
+			elif try_flag -march=native; then
+				ARGS+=(-march=native)
+			fi
+			;;
+		*)
+			echo -e "The CPU supports no Mlucas-recognized SIMD build mode ... building in scalar-double mode.\n"
+			echo "Warning: If this is a 64-bit x86 or ARM system, this likely means there is a bug in this script. Please report!" >&2
+			ARGS+=(-march=native)
+			;;
```
tdulcet, line 493: Please separate out the `none_x86` and `none` branches, as the `simd.c` program did, so we can provide the correct output messages.

makemake.sh
```diff
@@ -362,6 +495,35 @@ if [[ -x $TARGET ]]; then
...
+# when the default won't compile a C99 for-loop-scope declaration, so modern builds keep their default:
+c99_probe() {	# $1: optional -std flag
+	local tmpdir rc; tmpdir=$(mktemp -d) || return 1
+	printf 'int main(void){for(int i=0;i<1;++i){int j=i;(void)j;}return 0;}\n' >"$tmpdir/t.c"
+	"${CC:-gcc}" ${1:+"$1"} -c "$tmpdir/t.c" -o "$tmpdir/t.o" >/dev/null 2>&1; rc=$?
+	rm -rf "$tmpdir"; return $rc
+}
+if ! c99_probe && c99_probe -std=gnu99; then
+	CFLAGS_PROBED+=(-std=gnu99)
+fi
```
tdulcet: For consistency, we should probably always use the `-std=gnu99` option. I do not see any harm in doing so, and it would eliminate all of this added complexity.

tdulcet (changes requested):
Thanks for the additional improvements. I just have some suggestions to simplify it.

tdulcet (commented):
Thanks for addressing all of the feedback. It looks good now.

I will wait to approve it until we resolve the AVX-512 ER issue from https://github.com/primesearch/Mlucas/pull/73#discussion_r3596398595, in case any further changes are required.

MarkRose:
The AVX-512-ER point from #73 is now resolved: #73 gates the unused `vrcp28pd`/`mi64_modmul53_batch` path behind `__AVX512ER__` (so it only compiles on KNL) and drops `vrcp28pd` from the CI assembler probe. No further changes should be needed here from that thread — ready for re-review whenever you have a chance.

makemake.sh
```diff
@@ -236,27 +293,53 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
-	if (($(sysctl -n hw.optional.avx512f))); then
+	# MacOS: sysctl -n prints nothing (and exits nonzero) for an absent key - e.g. all the
+	# hw.optional.avx* keys on Apple Silicon - so capture each with a 0 default. This also lets the
+	# combined AVX-512||AVX2 test below use plain arithmetic without an empty operand tripping
+	# '((: || : syntax error' (as `(( || $(...) ))` would when the first sysctl yields nothing):
+	avx512f=$(sysctl -n hw.optional.avx512f 2>/dev/null || echo 0)
+	avx2_0=$( sysctl -n hw.optional.avx2_0  2>/dev/null || echo 0)
+	avx1_0=$( sysctl -n hw.optional.avx1_0  2>/dev/null || echo 0)
+	sse2=$(   sysctl -n hw.optional.sse2    2>/dev/null || echo 0)
+	neon=$(   sysctl -n hw.optional.neon    2>/dev/null || echo 0)
```
tdulcet, line 316: (Note that we format the shell scripts `shfmt`, so these likely would not stay aligned.)

tdulcet (approved):
Thanks for all your work on this. I will leave the PR open for a few days in case anyone else wants to review.

Issue #12, opened 2024-03-07
Segmentation fault in the "medium" self-test on ARM


tdulcet:
A segmentation fault sometimes occurs in the "medium" self-test: ./Mlucas -s m` on the macOS ARM CI runners:
```
INFO: Maximum recommended exponent for FFT length (3072 Kdbl) = 58884428; p[ = 57984131]/pmax_rec = 0.9847107796.
Initial DWT-multipliers chain length = [short] in carry step.
M57984131: using FFT length 3072K = 3145728 8-byte floats, initial residue shift count = 54682346
This gives an average   18.432658831278484 bits per digit
Using complex FFT radices        48        32        32        32
Using 2 threads in carry step
AddressSanitizer:DEADLYSIGNAL
=================================================================
==11579==ERROR: AddressSanitizer: SEGV on unknown address 0x00015b0d5300 (pc 0x000105d253d0 bp 0x00016b8bcb30 sp 0x00016b8bbf60 T0)
==11579==The signal is caused by a READ memory access.
    #0 0x105d253d0 in radix48_dit_pass1+0x2b2c (Mlucas:arm64+0x1017e53d0)
    #1 0x1047ca600 in mers_mod_square+0xdbb4 (Mlucas:arm64+0x10028a600)
    #2 0x10480a308 in ernstMain+0x60a4 (Mlucas:arm64+0x1002ca308)
    #3 0x10482891c in main+0x41d0 (Mlucas:arm64+0x1002e891c)
    #4 0x188fd50dc  (<unknown module>)

==11579==Register values:
 x[0] = 0x000000016b8bc040   x[1] = 0x000000016b8bbf80   x[2] = 0x000000015b0d5300   x[3] = 0x0000000000202002  
 x[4] = 0x0000000106efa000   x[5] = 0x0000000106efa000   x[6] = 0x0000000106efa000   x[7] = 0x0000000106ef9000  
 x[8] = 0x0000000000000000   x[9] = 0x0000000000000001  x[10] = 0x0000000000000000  x[11] = 0x0000000106efa000  
x[12] = 0x0000007000020000  x[13] = 0x0000000106efa000  x[14] = 0x0000000106f2aa28  x[15] = 0x0000000106efa000  
x[16] = 0x0000000106efa000  x[17] = 0x0000000107ba45e0  x[18] = 0x0000000000000000  x[19] = 0x000000016b8bc300  
[...]

PR #68, opened 2026-07-09
Declare AVX-512 opmask/high-vector clobbers; harden carry-loop index arithmetic

### LINKED ISSUE #16: Segmentation fault for "small" and "medium" self-tests on AMD 7950X
I was asked to run "./Mlucas -s tiny" here:
https://github.com/primesearch/Mlucas/issues/15#issuecomment-2027020472

Just wanted to create this issue that small and medium dump core.
```
hermann@7950x:~/Mlucas/obj$ uname -a
Linux 7950x 6.5.0-26-generic #26~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Mar 12 10:22:43 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
hermann@7950x:~/Mlucas/obj$ head -1 /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.4 LTS"
hermann@7950x:~/Mlucas/obj$ lscpu |grep "Model "
Model name:                         AMD Ryzen 9 7950X 16-Core Processor
hermann@7950x:~/Mlucas/obj$ 
```

```
hermann@7950x:~/Mlucas/obj$ ./Mlucas -s small

    Mlucas 20.1.1

    http://www.mersenneforum.org/mayer/README.html

System total RAM = 31190, free RAM = 29671
INFO: 29671 MB of free system RAM detected.
CPU Family = x86_64, OS = Linux, 64-bit Version, compiled with Gnu C [or other compatible], Version 11.4.0.
INFO: Build uses AVX512 instruction set.
INFO: Using prefetch.
INFO: using 64-bit-significand form of floating-double rounding constant for scalar-mode DNINT emulation.
INFO: System has 32 available processor cores.

           Mlucas selftest running.....

/****************************************************************************/

INFO: Unable to find/open mlucas.cfg file in r+ mode ... creating from scratch.
No CPU set or threadcount specified ... running single-threaded.
Set affinity for the following 1 cores: 0.
 worktodo.txt file not found...using user-supplied command-line exponent p = 5152643
INFO: Maximum recommended exponent for FFT length (256 Kdbl) = 5178863; p[ = 5152643]/pmax_rec = 0.9949371126.
M5152643: using FFT length 256K = 262144 8-byte floats, initial residue shift count = 1159220
This gives an average   19.655773162841797 bits per digit
Using complex FFT radices       256        16        32
Using 1 threads in carry step
M5152643 Roundoff warning on iteration        2, maxerr =   0.420149805045
M5152643 Roundoff warning on iteration        6, maxerr =   0.414123535156
M5152643 Roundoff warning on iteration       10, maxerr =   0.406272888184
M5152643 Roundoff warning on iteration       11, maxerr =   0.411499023438
Segmentation fault (core dumped)
hermann@7950x:~/Mlucas/obj$ 
```

and
```
hermann@7950x:~/Mlucas/obj$ ./Mlucas -s medium

    Mlucas 20.1.1

    http://www.mersenneforum.org/mayer/README.html

System total RAM = 31190, free RAM = 29671
INFO: 29671 MB of free system RAM detected.
CPU Family = x86_64, OS = Linux, 64-bit Version, compiled with Gnu C [or other compatible], Version 11.4.0.
INFO: Build uses AVX512 instruction set.
INFO: Using prefetch.
INFO: using 64-bit-significand form of floating-double rounding constant for scalar-mode DNINT emulation.
INFO: System has 32 available processor cores.

           Mlucas selftest running.....

/****************************************************************************/

INFO: Unable to find/open mlucas.cfg file in r+ mode ... creating from scratch.
No CPU set or threadcount specified ... running single-threaded.
Set affinity for the following 1 cores: 0.
 worktodo.txt file not found...using user-supplied command-line exponent p = 39003229
INFO: Maximum recommended exponent for FFT length (2048 Kdbl) = 39606917; p[ = 39003229]/pmax_rec = 0.9847580159.
Initial DWT-multipliers chain length = [short] in carry step.
M39003229: using FFT length 2048K = 2097152 8-byte floats, initial residue shift count = 31137256
This gives an average   18.598188877105713 bits per digit
Using complex FFT radices      1024        32        32
Using 1 threads in carry step
Segmentation fault (core dumped)
hermann@7950x:~/Mlucas/obj$ 
```

### LINKED ISSUE #52: Segmentation fault with F33 P-1 stage 2
This error was reported by [Ken on the forum](https://www.mersenneforum.org/node/22731?p=1084537#post1084537) when attempting to run P-1 stage 2 on F33 with an AVX512 system using GCC 13.3:
```
Using 64 threads in carry step
Buffer-init done; clocks = 00:13:24.757, MaxErr = 0.062500000.
Stage 2 q0 = 699999960, k0 = 2121212
Computing Stage 2 loop-multipliers...
Stage 2 loop-multipliers: clocks = 00:01:58.143, MaxErr = 0.070312500.

Thread 68 "Mlucas" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fd1f67006c0 (LWP 248644)]
0x0000555555719405 in radix32_dyadic_square (a=<optimized out>, arr_scratch=<optimized out>, 
    n=<optimized out>, radix0=<optimized out>, rt0=<optimized out>, rt1=<optimized out>, ii=59, 
    nradices_prim=28, radix_prim=0x555555937180 <radix_prim>, incr=64, init_sse2=0, thr_id=59, 
    fwd_fft_only=<optimized out>, c_arr=0x0) at ../src/radix32_dyadic_square.c:1121
1121			SSE2_RADIX32_WRAPPER_DIT(add0,add1,add2,add3,add4,add5,add6,add7
```
Ken reports that it works as expected on a different system with GCC 11.4.

https://github.com/primesearch/Mlucas/blob/b6dda5253c9edfc34d5fbcdad12704bb61941370/src/radix32_dyadic_square.c#L1114-L1122

### LINKED ISSUE #65: Segmentation fault with Clang 18+/GCC 15+ and AVX512

MarkRose:
## Summary

Two related fixes for the AVX-512-only segfault family (#65, #52, #16, #12). The first is the root cause; the second is defense-in-depth for the same corruption.

### 1. Declare AVX-512 opmask (`k`) and high-vector clobbers in inline asm

No inline-asm block in the codebase declared the AVX-512 opmask registers `k0`–`k7` as clobbered (`grep '"k[0-7]"' src/` → 0 matches), yet **51 blocks across 16 files write them** via `vpcmpd`/`vcmppd`/`kshiftrw`/`kandw`/`kmov` etc. (`carry_gcc64.h` alone uses `%%k1`–`%%k7` ~2200 times). Under `-O3 -flto -march=native`, modern GCC/Clang auto-vectorize the surrounding C using opmask and `zmm16`–`zmm31` registers and keep those values live across the `asm volatile` blocks, which then silently destroy them.

This is a clean explanation for the observed pattern: the crash is **AVX-512-only** (AVX2 builds allocate no `k` registers), it **moves with compiler version and CPU vendor** (register-allocation dependent), and building with `-fno-tree-vectorize -fno-slp-vectorize` is expected to make it disappear.

This PR adds exactly the `k` registers each block writes (including `k0`, which `twopmodq80/100` write via `vcmppd`/`kandw`). While auditing the same blocks it also fills in undeclared high-vector clobbers:
- `xmm8` at the two `radix16/32_dyadic_square.c` sites (blocks use `xmm8`/`ymm8` but the clobber list stopped short),
- `zmm16`–`zmm31` omissions in `carry_gcc64.h` (zmm18/19), `mi64.c` (zmm30), `radix32_wrapper_square_gcc64.h` (zmm16/17) and `twopmodq100.c` (zmm30/31).

All additions are to blocks compiled only under `USE_AVX512`. Every operand in these blocks is memory/GPR/immediate (`"m"`/`"g"`/`"e"`), so no clobber can collide with an operand constraint. Over-declaring a clobber costs a little performance, never correctness.

### 2. Cast the carry-chain offset to `int` in `wt1[]` indexing

In the LOACC chained-weights carry loops, the per-chain offset `ii` is `uint32` while `col`/`co2`/`co3` are `int`, so `&wt1[co2-1-ii]` and `&wt1[co3-1-ii]` are evaluated in **unsigned 32-bit** arithmetic. If upstream corruption drives `co2`/`co3` negative, the index wraps to ~2³² and produces a wild pointer ~34 GB past `wt1` — matching the fault address decoded in the #16 7950X reports.

Casting `ii` to `int` at the three `wt1[]` index sites keeps the arithmetic signed: a corrupted index now lands near `wt1` (debuggable / promptly faulting) instead of at a random high address, and cannot silently read valid-but-wrong weights from mapped memory. An `ASSERT` on the three indices aborts cleanly on underflow; it runs only in the per-`j`-block weights-init path (a few int compares per 8–16 carries), so the cost is negligible. This also covers the SSE2/scalar/ARM branches.

## Verification

- `"k0"`–`"k7"` accepted as clobbers by **gcc 16.1.1** and **clang 22**.
- AVX-512 tree builds clean (0 errors) with both compilers.
- AVX2 `-s s` self-test passes all 24 radix-sets (256K–1920K) with no assert false-fire.

⚠️ The build host is AVX2-only (Comet Lake), so I could **only compile-check** AVX-512. Runtime confirmation of the segfault fix is expected from the CI self-test matrix and AVX-512 (Zen 4) hardware — the `-fno-tree-vectorize`/`-fno-slp-vectorize` falsification check is the cleanest way to confirm the mechanism.

**Update (2026-07-11):** the runtime segfault was subsequently reproduced on GitHub's AVX-512 runners, where this PR's AVX-512 self-test passes — runtime confirmation of the fix, upgrading the issue links below.

Fixes #65. Fixes #16. Likely fixes #52 (same clobber family; F33-scale verification impractical locally). Helps #12 (shares the index-wrap hardening; the ARM-side clobber audit remains open).

### DIFF
--- a/src/carry_gcc64.h
+++ b/src/carry_gcc64.h
@@ -10637,7 +10637,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		,	[__p4]   "m" (Xp4)\
 		/* Mar 2018: Needed to support PRP testing: */\
 		,	[__prp_mult]   "m" (Xprp_mult)\
-		: "cc","memory","k1","k2","k3","k4","k5","xmm18","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","k3","k4","k5","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm18","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25"	/* Clobbered registers */\
 	);\
 	}
 
@@ -12318,7 +12318,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		,	[__p4]   "m" (Xp4)\
 		/* Mar 2018: Needed to support PRP testing: */\
 		,	[__prp_mult]   "m" (Xprp_mult)\
-		: "cc","memory","k1","k2","k3","k4","xmm18","xmm19","rax","rbx","rcx","rdx","rdi","rsi","r14","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","k3","k4","rax","rbx","rcx","rdx","rdi","rsi","r14","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm18","xmm19","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
 	);\
 	}
 
--- a/src/carry_gcc64.h
+++ b/src/carry_gcc64.h
@@ -5126,7 +5126,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		, [__sinwtm1]	"m" (Xsinwtm1)		\
 		, [__sse_bw]	"m" (Xsse_bw)		\
 		, [__sse_nm1]	"m" (Xsse_nm1)		\
-		: "cc","memory","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm17","xmm18","xmm20","xmm21","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","k3","k4","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm17","xmm18","xmm20","xmm21","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
 	);\
 	}
 
@@ -5567,7 +5567,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		, [__sinwtm1]	"m" (Xsinwtm1)		\
 		, [__sse_bw]	"m" (Xsse_bw)		\
 		, [__sse_nm1]	"m" (Xsse_nm1)		\
-		: "cc","memory","rax","rbx","rcx","rdx","rdi","rsi","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm30","xmm31"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","rax","rbx","rcx","rdx","rdi","rsi","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm30","xmm31"	/* Clobbered registers */\
 	);\
 	}
 
@@ -6084,7 +6084,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		, [__sinwtm1]	"m" (Xsinwtm1)		\
 		, [__sse_bw]	"m" (Xsse_bw)		\
 		, [__sse_n]		"m" (Xsse_n)		\
-		: "cc","memory","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm17","xmm18","xmm20","xmm21","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","k3","k4","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm17","xmm18","xmm20","xmm21","xmm24","xmm25","xmm30","xmm31"	/* Clobbered registers */\
 	);\
 	}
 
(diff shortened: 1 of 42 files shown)


src/carry_gcc64.h
```diff
@@ -10637,7 +10637,7 @@ Check the compile optimization level - If -O0, try upping to at east -O1.
 		,	[__p4]   "m" (Xp4)\
 		/* Mar 2018: Needed to support PRP testing: */\
 		,	[__prp_mult]   "m" (Xprp_mult)\
-		: "cc","memory","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25"	/* Clobbered registers */\
+		: "cc","memory","k1","k2","k3","k4","k5","xmm18","rax","rbx","rcx","rdx","rdi","rsi","r15","xmm0","xmm1","xmm2","xmm3","xmm4","xmm5","xmm6","xmm7","xmm8","xmm9","xmm10","xmm11","xmm12","xmm13","xmm14","xmm15","xmm16","xmm17","xmm20","xmm21","xmm22","xmm23","xmm24","xmm25"	/* Clobbered registers */\
```
tdulcet: Maybe keep the clobber list in order (i.e. put `xmm18` after `xmm17`).

This occurs a few times below as well.

tdulcet (commented):
Thanks for debugging and fixing this! It looks good to me, but I will defer to @ldesnogu for the final review.

src/carry_gcc64.h
MarkRose: Addressed in follow-on commit.

MarkRose:
Heads-up: this PR also resolves the **intermittent `Mlucas Linux` / `Mlucas Singlethreaded` / `Mlucas macOS (intel)` SIGSEGVs (exit 139)** that show up nondeterministically across the other open PRs' CI runs.

They aren't flaky infrastructure — they're exactly the `#65`/`#16` AVX-512 register-clobber bug this PR fixes, surfacing by hardware lottery:

- GitHub's `ubuntu-24.04` (and `macos-15-intel`) runner labels sometimes land on an **AVX-512 CPU** and sometimes on AVX2. With `-march=native`, the self-test builds the AVX-512 FFT path only on the AVX-512 runners.
- On branches **without** this PR (everything currently based on `main`), that AVX-512 self-test segfaults. Verified on one such run: the crashing job built `__AVX512F__` and died in the AVX-512 self-test at FFT length 240K; the passing jobs on sibling PRs had built the AVX2 path.
- It hit a completely trivial PR (a one-line cfg-scan bound) and skipped others, which is the tell that it's runner-CPU-dependent, not code from those PRs.

Crucially, on **this** PR the `Mlucas Linux (ubuntu-22.04, clang)` job built the **AVX-512** path and **passed** — so the clobber fixes here resolve the runtime segfault on real AVX-512 hardware. In other words, GitHub's AVX-512 runners are effectively a free runtime repro of `#65`, and this PR is confirmed to fix it. Once this merges to `main`, those intermittent AVX-512 self-test segfaults should stop across the board.

ldesnogu (approved):
It's quite difficult to say if this fully covers all the cases of missing clobbered registers.

I guess we will have to trust Claude on this or we would have to write some script to check nothing was missed.

On the other hand, adding too many clobbered registers shouldn't generate any functional regression (though if it's too pessimistic there could be speed regression).

As a side note, I discussed several times with Ernst about clobbered register lists for Arm and Apple Arm in particular. So hopefully these should be correct.

Given that it also fixes some existing bugs, I propose that we merge this without further scrutiny.

PR #70, opened 2026-07-09
Free threadpools on re-init instead of leaking them (fixes #44)

### LINKED ISSUE #44: Thread leak per assignment
Mlucas leaks a thread per assignment, so that the total number of threads increases over time. This can be seen from tools like `htop` or `ps`:
Image
as well as from looking at the `/proc` filesystem:
```
$ for p in $(pgrep -x Mlucas); do grep -H '^State\|^Threads' /proc/$p/status; grep -H '^State' /proc/$p/task/[0-9]*/status; done
/proc/3426995/status:State:     R (running)
/proc/3426995/status:Threads:   18
/proc/3426995/task/3426995/status:State:        R (running)
/proc/3426995/task/3427009/status:State:        S (sleeping)
/proc/3426995/task/3427012/status:State:        S (sleeping)
/proc/3426995/task/3427013/status:State:        S (sleeping)
/proc/3426995/task/3452374/status:State:        S (sleeping)
/proc/3426995/task/3489695/status:State:        S (sleeping)
/proc/3426995/task/3543544/status:State:        S (sleeping)
/proc/3426995/task/3581303/status:State:        S (sleeping)
/proc/3426995/task/3621917/status:State:        S (sleeping)
/proc/3426995/task/3656074/status:State:        S (sleeping)
/proc/3426995/task/3708141/status:State:        S (sleeping)
/proc/3426995/task/3745384/status:State:        S (sleeping)
/proc/3426995/task/3806488/status:State:        S (sleeping)
/proc/3426995/task/3844755/status:State:        S (sleeping)
/proc/3426995/task/3894608/status:State:        S (sleeping)
/proc/3426995/task/3933565/status:State:        S (sleeping)
/proc/3426995/task/3986662/status:State:        R (running)
/proc/3426995/task/4023402/status:State:        S (sleeping)
```
In this case, there are 18 threads when there should only be two threads.

MarkRose:
## Summary

`threadpool_free()` is complete and correct but **was never called anywhere** in the codebase, so worker threads accumulate:

- **`mers_mod_square` / `fermat_mod_square`**: every `first_entry` re-init overwrites the static `tpool` pointer with a fresh `threadpool_init(...)`, leaking `NTHREADS` worker threads per runlength/radix-set switch (this matches the reporter's "18 threads with NTHREADS=1 after ~16 re-inits").
- **`pm1` Stage 2**: `threadpool_init` runs on every invocation and the pool is never freed; the per-thread arrays are freed at `ERR_RETURN` but the several early `return ierr` modmul-error paths bypass that cleanup.

## Fix

- **mers/fermat**: free the previous pool before re-creating it at the top of a `mod_square` call. Safe because the prior config's pool is **quiescent** there — the preceding call already drained it via the dispatch-completion wait (`free_tasks_queue.num_tasks == pool_work_units`).
- **pm1 Stage 2**: free the pool at `ERR_RETURN` (prompt cleanup on the normal + `goto`-error exits) **and** free-before-realloc the pool + per-thread arrays at the top of the alloc block, so the early-`return` error paths that bypass `ERR_RETURN` can't leak across repeated Stage-2 calls. Also removed the dead `pthread_attr_t` (init'd every call, never used — `threadpool_init` creates its threads with a `NULL` attr — so it was itself a small per-call pthread-attr leak).

## Verification

AVX2 build, 4 carry threads (`-cpu 0:3`), multi-length self-test `-s s` (127 radix-set re-inits), process thread count sampled from `/proc/<pid>/task`:

| | peak threads | behavior |
|---|---|---|
| baseline (`main`) | **621** | climbs monotonically with re-init count |
| this PR | **117** | **plateaus** — bounded, no per-config growth |

24/24 radix-sets still pass with **zero crashes**, confirming the pre-init free is race-free. Instrumentation confirmed `threadpool_free` is now called once per re-init (25 frees / 26 mers pools in `-s tt`), each `pthread_join` returning 0.

## Note on the residual (separate, bounded)

The remaining ~117 threads are **not** a per-config leak: each `radix*_ditN_cy_dif1.c` holds its own static carry-step `tpool`, created **once per leading radix** (guarded by `if(tdat==0x0)`, and `tdat` is never reset) and reused for the life of the process. That's a bounded one-pool-per-radix allocation (≈ distinct-radix count × `NTHREADS`), which is why the fixed curve *plateaus* instead of growing. Freeing those cleanly needs a different mechanism (they have no re-init point to hook, and no teardown hook exists), so I've left it out of this PR — happy to do it as a follow-up if you'd like.

`pm1` Stage 2 is compile-verified and follows the same (verified) pattern as mers; it isn't exercised by the Mersenne self-test, so its runtime path wasn't driven here.

Fixes #44.

### DIFF
--- a/src/fermat_mod_square.c
+++ b/src/fermat_mod_square.c
@@ -1147,6 +1147,11 @@ int fermat_mod_square(double a[], int arr_scratch[], int n, int ilo, int ihi, ui
 
 		main_work_units = 0;
 		pool_work_units = radix0;
+		// Free any threadpool left over from a prior runlength/radix-set before creating the new one,
+		// so a radix/runlength switch does not leak NTHREADS worker threads each time. Safe here because
+		// this re-init only happens at the top of a mod_square call, at which point the previous config's
+		// pool is quiescent - the preceding call drained it via the dispatch-completion wait below:
+		if(tpool) { threadpool_free(tpool); tpool = 0x0; }
 		ASSERT(0x0 != (tpool = threadpool_init(NTHREADS, MAX_THREADS, pool_work_units, &thread_control)), "threadpool_init failed!");
 		printf("%s: Init threadpool of %d threads\n",func,NTHREADS);
 
--- a/src/mers_mod_square.c
+++ b/src/mers_mod_square.c
@@ -1352,6 +1352,11 @@ for(i=0; i < NRT; i++) {
 
 		main_work_units = 0;
 		pool_work_units = nchunks;
+		// Free any threadpool left over from a prior runlength/radix-set before creating the new one,
+		// so a radix/runlength switch does not leak NTHREADS worker threads each time. Safe here because
+		// this re-init only happens at the top of a mod_square call, at which point the previous config's
+		// pool is quiescent - the preceding call drained it via the dispatch-completion wait below:
+		if(tpool) { threadpool_free(tpool); tpool = 0x0; }
 		ASSERT(0x0 != (tpool = threadpool_init(NTHREADS, MAX_THREADS, pool_work_units, &thread_control)), "threadpool_init failed!");
 		printf("%s: Init threadpool of %d threads\n",func,NTHREADS);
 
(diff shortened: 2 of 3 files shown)


tdulcet (approved):
Thanks for fixing this!

It looks good to me. I will just leave it open for a few days before merging, in case anyone else wants to review.

PR #71, opened 2026-07-09
Fix 'asm operand has impossible constraints' in hand-tuned asm on ancient compilers (part of #60)

MarkRose:
## Summary

Part of #60 (older GCC/Clang support). This fixes the `error: 'asm' operand has impossible constraints or there are not enough registers` failures in the hand-tuned inline-asm — the sub-item of #60 that bites **ancient compilers** (reproduced on GCC 5.4 / Ubuntu 16.04 at `-flto`) as well as AddressSanitizer builds on modern toolchains.

### Root cause
Several hand-tuned asm blocks clobber **all 14** non-reserved x86-64 GPRs (`rax`–`r15` minus `rsp`/`rbp`) while also taking memory operands. That leaves the register allocator no spare GPR to hold an operand's base address, which is fatal the moment register pressure is elevated from another source — a weak old allocator (GCC 5.4 at `-flto`) or ASan's shadow-memory / fake-stack base register (modern GCC & Clang). It's the long-documented GCC PR23200 limitation; the in-tree "use -O1 or higher" note does not cover these cases.

### Fix, split by whether a non-asm fallback exists

- **`dft_macro.c`'s radix-16 `SSE2_RADIX16_DIT_0TWIDDLE` / `DIF_0TWIDDLE`** — the only site that fails under **GCC 5.4 `-flto`**, and it has no fallback, so the register pressure is fixed in the asm itself: the Im-part output "shadow" registers (`rdi = rsi + off`, `r9 = r8 + off`, `off` = 0x20 for AVX2 / 0x40 for AVX-512) are folded into compile-time displacement operands, and the isrt2/cc0 trig pointers are re-homed. **The AVX2/FMA3 `DIF`+`DIT` and the AVX-512 `DIT` are fixed in commit 1; the AVX-512 `DIF` is fixed in commit 4** (see below — it is not needed by the toolchains this PR targets, but GCC 13 allocates differently and does need it). The "Linux old" CI job compiles every SIMD mode on the old toolchains; AVX-512 was the last mode still failing there. The folds are pure address rewrites — e.g. `%c[__oN](%%r9)` == `%c[__oN]+0x40(%%r8)` since `r9 == r8+0x40` — so the emitted vector math is unchanged: residues are byte-identical (verified on the radix-256 DFT path) and it benchmarks no slower.

- **`mi64.c` / `twopmodq.c` / `twopmodq64_test.c` integer-modmul asm** — fails **only under ASan** (verified: UBSan-only, TSan-only, and plain GCC 5.4 all compile fine) and already has a `YES_ASM`-gated portable-C fallback, so `YES_ASM` is disabled under `__SANITIZE_ADDRESS__`. The shipping asm is left **completely unchanged**.

- **`SSE2_RADIX_11/13_DFT` (radix 44/52/176/208/352)** — fails only under ASan, has no in-SIMD-mode fallback, and genuinely uses all 14 GPRs, so the enclosing carry routines are annotated with `no_sanitize_address`. This suppresses ASan *instrumentation* only; the asm is untouched.

ASan detection uses a nested `#if defined(__SANITIZE_ADDRESS__)` / `#elif defined(__has_feature)` form — the flat `|| (defined(__has_feature) && __has_feature(...))` form makes GCC 5.4's preprocessor choke, since it lacks `__has_feature` and tokenizes it as `0(0)`.

### Verification
- Whole tree compiles clean under `-fsanitize=address,undefined` in **every** SIMD mode (SSE2/AVX/AVX2/AVX512) — was 7+ `impossible constraints` errors (plus more masked, since GCC reports only the first per compile).
- **GCC 5.4.0 (Ubuntu 16.04) `-O3` and `-O3 -flto`, every SIMD mode**, link and run `./Mlucas -s tt` with correct residues — reproduced with the exact CI invocation (`bash makemake.sh use_hwloc <MODE>`). The AVX-512 LTO build previously failed to link at `dft_macro.c:5050/5101`.
- `./Mlucas -s tt` residues, and the radix-256 DFT path (`-fftlen 1024 -radset 2`, which exercises both edited macros), are byte-identical to before the change.
- `-fftlen 1024` radix-256 benchmark is at or below baseline in every trial (the DFT edits only remove address-setup instructions).

Performance of normal (non-sanitizer) builds is unaffected: the integer-asm and radix-11/13 asm are byte-for-byte unchanged, and the radix-16 DFT edits change only address computation, not the vector arithmetic.

### The AVX-512 `SSE2_RADIX16_DIF_0TWIDDLE`, needed by newer GCC

`src/sse2_macro_gcc64.h` holds one copy of each radix-16 `_0TWIDDLE` macro per SIMD mode. Commit 1
edits the AVX-512 **DIT** (line ~4311) and both AVX2 variants (~9206 / ~9635). The AVX-512 **DIF**
(~3880) was left alone, and on GCC 5.4 and GCC 11.4 that is fine — neither allocator needs the extra
register there.

GCC 13.3 does. Per-TU compile sweep of all 101 `src/*.c` with
`-DUSE_AVX512 -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma -O3 -fsanitize=address`:

| tree | TUs failing `'asm' operand has impossible constraints` |
|---|---|
| `main` | 13 |
| this PR through commit 3 | 3 — `radix48`, `radix144`, `radix1008`, all at `sse2_macro_gcc64.h:3882` |
| **+ commit 4** | **0 / 101** |

Same treatment as commit 1 applies to the AVX2 DIF: drop the `rdi` shadow pointer holding
`rsi + 0x40` and fold the offset into each displacement; reuse `r14`, which already holds `isrt2` and
is written exactly once in the macro, for the two `cc0` accesses `rdi` was carrying and for a
redundant `isrt2` reload into `r9`. `rdi` then comes off the clobber list — 14 GPRs to 13.

With commit 4, the other SIMD modes are also clean on GCC 13.3 under ASan: **avx2 0/101, avx 0/101,
sse2 0/101**. The SSE2 / AVX-sans-FMA / ARM copies of these macros are deliberately untouched — they
are not broken on any tested compiler, and editing hot asm without a demonstrated failure is not worth
the risk.

**GCC 11.4 is 0/101 both with and without commit 4**, so this is forward-compatibility for newer
compilers rather than a correction to the original work.

Residues are unchanged. Non-sanitizer AVX-512 builds before and after commit 4, under Intel SDE:

```
 48K {48,32,16}   p=1001467  Res64=6B1C76AB5431FDA4  AvgMaxErr=0.232575335
144K {144,16,32}  p=2934479  Res64=1B90A27301980A3A  AvgMaxErr=0.240325056
```

identical to each other and to the `MersVec[]` reference values, `AvgMaxErr` matching to every digit.
As a control that these cases really do execute the edited macro, injecting a single wrong
displacement into it turns the 48K run into `ERR_ROUNDOFF` with no residue emitted at all.

### Out of scope (pre-existing CI failures, not regressions from this PR)
These jobs fail identically on `main`:
- **old-Clang AVX-512** (`unknown register name 'xmm30'`) — a *different* #60 sub-item (old Clang rejecting AVX-512 extended registers xmm16-31), addressed by the `try_avx512_asm` probe in the build-portability PR #67, not here.
- **Windows** (`factor.c`/`factor.h` implicit-declaration / `USE_FMADD` `#error`) — pre-existing, unrelated files.

### DIFF
--- a/src/masterdefs.h
+++ b/src/masterdefs.h
@@ -45,6 +45,23 @@
 	#include <console.h>	/* Macintosh CW */
 #endif
 
+/* Task 23 / issue #60: AddressSanitizer's shadow-memory instrumentation elevates GPR pressure enough that
+the hand-tuned all-14-GPR-clobber SIMD DFT macros (e.g. SSE2_RADIX_11/13_DFT) can no longer be register-
+allocated ('operand has impossible constraints', GCC PR23200). Those macros have no non-asm fallback within
+a given SIMD build mode, so we suppress ASan *instrumentation* (not the asm) on the enclosing carry routines;
+the asm itself is unchanged, so run-time performance of normal builds is unaffected. */
+#if defined(__SANITIZE_ADDRESS__)
+	#define ATTR_NO_ASAN	__attribute__((no_sanitize_address))
+#elif defined(__has_feature)
+	#if __has_feature(address_sanitizer)	/* older Clang: no __SANITIZE_ADDRESS__ predefine */
+		#define ATTR_NO_ASAN	__attribute__((no_sanitize_address))
+	#else
+		#define ATTR_NO_ASAN
+	#endif
+#else
+	#define ATTR_NO_ASAN
+#endif
+
 #undef  EWM_DEBUG
 #define EWM_DEBUG		0	/* Set = 1 to turn on various debugging diagnostics, especially DBG_ASSERT, defined in util.c . */
 
--- a/src/masterdefs.h
+++ b/src/masterdefs.h
@@ -45,7 +45,7 @@
 	#include <console.h>	/* Macintosh CW */
 #endif
 
-/* Task 23 / issue #60: AddressSanitizer's shadow-memory instrumentation elevates GPR pressure enough that
+/* AddressSanitizer's shadow-memory instrumentation elevates GPR pressure enough that
 the hand-tuned all-14-GPR-clobber SIMD DFT macros (e.g. SSE2_RADIX_11/13_DFT) can no longer be register-
 allocated ('operand has impossible constraints', GCC PR23200). Those macros have no non-asm fallback within
 a given SIMD build mode, so we suppress ASan *instrumentation* (not the asm) on the enclosing carry routines;
--- a/src/radix176_ditN_cy_dif1.c
+++ b/src/radix176_ditN_cy_dif1.c
@@ -132,7 +132,7 @@
 
 /****************/
 
-int radix176_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[], double wt1[], int si[], double base[], double baseinv[], int iter, double *fracmax, uint64 p)
+ATTR_NO_ASAN int radix176_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[], double wt1[], int si[], double base[], double baseinv[], int iter, double *fracmax, uint64 p)
 {
 /*
 !...Acronym: DWT = Discrete Weighted Transform, DIT = Decimation In Time, DIF = Decimation In Frequency
@@ -2033,7 +2033,7 @@ void radix176_dit_pass1(double a[], int n)
 	#endif
 
 	void*
-	cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
+ATTR_NO_ASAN cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
 	{
 		struct cy_thread_data_t* thread_arg = targ;	// Move to top because scalar-mode carry pointers taken directly from it
 		double *addr;
--- a/src/radix176_ditN_cy_dif1.c
+++ b/src/radix176_ditN_cy_dif1.c
@@ -2032,8 +2032,8 @@ void radix176_dit_pass1(double a[], int n)
 		#error pthreaded carry code requires GCC build!
 	#endif
 
-	void*
-ATTR_NO_ASAN cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
+	ATTR_NO_ASAN void*
+	cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
 	{
 		struct cy_thread_data_t* thread_arg = targ;	// Move to top because scalar-mode carry pointers taken directly from it
 		double *addr;
--- a/src/radix176_ditN_cy_dif1.c
+++ b/src/radix176_ditN_cy_dif1.c
@@ -132,7 +132,7 @@
 
 /****************/
 
-int radix176_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[], double wt1[], int si[], double base[], double baseinv[], int iter, double *fracmax, uint64 p)
+ATTR_NO_ASAN int radix176_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[], double wt1[], int si[], double base[], double baseinv[], int iter, double *fracmax, uint64 p)
 {
 /*
 !...Acronym: DWT = Discrete Weighted Transform, DIT = Decimation In Time, DIF = Decimation In Frequency
@@ -2094,7 +2094,7 @@ void radix176_dit_pass1(double a[], int n)
 	#endif
 
 	void
-	cy176_process_chunk(void*targ, int thread_num)	// Thread-arg pointer *must* be cast to void and specialized inside the function
+ATTR_NO_ASAN cy176_process_chunk(void*targ, int thread_num)	// Thread-arg pointer *must* be cast to void and specialized inside the function
 	{
 		struct cy_thread_data_t* thread_arg = targ;	// Move to top because scalar-mode carry pointers taken directly from it
 		double *addr;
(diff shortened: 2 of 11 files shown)


src/masterdefs.h
```diff
@@ -45,6 +45,23 @@
 	#include <console.h>	/* Macintosh CW */
 #endif
 
+/* Task 23 / issue #60: AddressSanitizer's shadow-memory instrumentation elevates GPR pressure enough that
```
tdulcet: Task 23 seems like some internal Claude Code reference that should be omitted from the public comment here. This also occurs several times below.

tdulcet (commented):
Thanks for all of the additional PRs and for looking into this. The fixes seem strightforward, but I will let someone else double check the inline assembly changes.

> ASan's shadow-memory / fake-stack base register (modern GCC & Clang)

I wonder if we should instead update the CI to use NOSIMD builds with ASan, as testing the assembly code with ASan/UBSan may not provide much value in general. It would also simplify this PR.

src/masterdefs.h
MarkRose: Ah, yes it is. I'll get them removed. Claude has a bad habit of including internal references in comments. I had Fable do an analysis of the code base as well as all the issues here on GitHub into a combined plan. I've been going through the more most important and complex ones first, but i should have another dozen or so PRs for the remaining tasks/issues (some will naturally combine). Thank you for taking the time to review!

MarkRose:
> I wonder if we should instead update the CI to use NOSIMD builds with ASan, as testing the assembly code with ASan/UBSan may not provide much value in general. It would also simplify this PR.

I'd lean toward keeping ASan on the SIMD builds. The value isn't in checking the asm, it's that the C code surrounding it only executes in the SIMD build modes, and that's where an out-of-bounds bug would be most likely to hide and least likely to be caught. A NOSIMD-only ASan job wouldn't exercise those paths.

MarkRose:
One question on scoping: the AVX-512 `SSE2_RADIX16_DIF_0TWIDDLE` fix (the last commit here, detailed in the description) is not needed by the older compilers this PR targets — GCC 11.4 measures 0/101 failing TUs with and without it. It is GCC 13.3, which allocates differently, that needs it to reach 0/101.

Happy to split it into its own PR if you would rather keep this one scoped to the older-toolchain work.

src/masterdefs.h
```diff
@@ -45,6 +45,23 @@
 	#include <console.h>	/* Macintosh CW */
 #endif
 
+/* AddressSanitizer's shadow-memory instrumentation elevates GPR pressure enough that
+the hand-tuned all-14-GPR-clobber SIMD DFT macros (e.g. SSE2_RADIX_11/13_DFT) can no longer be register-
+allocated ('operand has impossible constraints', GCC PR23200). Those macros have no non-asm fallback within
+a given SIMD build mode, so we suppress ASan *instrumentation* (not the asm) on the enclosing carry routines;
+the asm itself is unchanged, so run-time performance of normal builds is unaffected. */
+#if defined(__SANITIZE_ADDRESS__)
```
tdulcet, line 53: This could be simplified to just `#ifdef __SANITIZE_ADDRESS__`. This also occurs below.

src/radix176_ditN_cy_dif1.c
```diff
@@ -2033,7 +2033,7 @@ void radix176_dit_pass1(double a[], int n)
 	#endif
 
 	void*
-	cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
+ATTR_NO_ASAN cy176_process_chunk(void*targ)	// Thread-arg pointer *must* be cast to void and specialized inside the function
```
tdulcet: Maybe put the attribute before the return type for consistency (e.g. before the `void*`). This also occurs below.

tdulcet (dismissed):
Thanks, this looks good to me, but it would be great if someone else could review the inline assembly changes.

src/masterdefs.h
MarkRose: That would break Clang 20 and below, which don't define it.

MarkRose:
Moved `ATTR_NO_ASAN` ahead of the return type on the five `cy*_process_chunk` definitions (radix 44, 52, 176, 208, 352) — you're right, it was inconsistent with the five `radix*_ditN_cy_dif1()` entry points in the same PR, which already had it first.

```c
	ATTR_NO_ASAN void*
	cy176_process_chunk(void*targ)
```

That also restores the leading tab on the name line, which my original reflow had dropped — the two-line "return type / name" layout is the existing style in these files.

Verified all five compile clean under gcc and clang, plain and with `-fsanitize=address` (the configuration the attribute exists for).

The **Task 23** references are already gone — no occurrences remain on the branch.

Two things worth flagging:

- **#79 touches these same lines.** It changes the `cy*_process_chunk` signatures (`void*` → `void` plus an `int thread_num`), so whichever of the two merges second will conflict here. Small and mechanical, but worth knowing about.
- My earlier question about splitting out the AVX-512 `SSE2_RADIX16_DIF_0TWIDDLE` commit — it's needed by GCC 13.3 rather than by the ancient toolchains this PR targets. I read your approval as "keep it here", so I've left it in; happy to split it out if you'd rather.

Sorry — the push will have dismissed your approval. Re-requesting.

tdulcet (dismissed):
Thanks.

MarkRose:
Rebased onto current `main` to clear the merge conflicts — sorry that this dismisses the approval, but the branch could not be merged as it stood. Two conflict sites, both from #79 changing the `cy*_process_chunk` signature to `void ...(void*targ, int thread_num)`: the resolution keeps that new signature and this branch's `ATTR_NO_ASAN`-before-the-return-type placement. Re-verified after: `avx2`, `avx512` and `nosimd` all build, and `avx2 -s tt` is rc=0 with 19 residues.

tdulcet (approved):

PR #72, opened 2026-07-10
Recover the p-1 stage-1 B1 from the .s1_prod savefile on restart (fixes #31)

### LINKED ISSUE #31: Assertion failed: For completed S1 expect ilo == maxiter == PM1_S1_PROD_BITS
I'm using the `mlucas.sh` script from Distributed-Computing-Scripts to run mlucas on aarch64 Linux (Fedora 40 Asahi). Initially, 8 processes were started. I stopped and started the processes several times using `jobs.sh` created in the `objs` directory (I didn't use crontab). At some point, I noticed that only 4 processes were running. The other 4 refused to start. They were the processes that reached stage 2. The log files for the failing processes show an assertion error:

```
 INFO: restart file p124130143 found...reading...
ERROR: at line 1776 of file ../src/Mlucas.c
Assertion failed: For completed S1 expect ilo == maxiter == PM1_S1_PROD_BITS!
```

I'm attaching the full log.
[Mlucas.out.txt](https://github.com/user-attachments/files/15574083/Mlucas.out.txt)

Also note that the log contains `Stage 2: No factor found.` but I'm not sure it's the final result.

I can reproduce the error by running `../Mlucas -core '4' -maxalloc 11.25` in the `Mlucas/obj/run4` directory.

MarkRose:
## Summary

Fixes #31. A p-1 run whose Stage 1 is interrupted and later resumed under a **different memory budget** (or a changed `-b1`) crashes on restart, because the auto-sized Stage 1 bound B1 is not recovered from the savefile that records it.

### Root cause
`pm1_set_bounds()` auto-sizes B1 from available RAM: when there is too little memory for Stage 2 it bumps B1 up ~25% to run a deeper Stage 1. On a restart with a different amount of free RAM it therefore picks a *different* B1 than the Stage 1 already in progress. The in-progress LR-binary powering of the Stage 1 prime-powers product — and the residue restart-file's iteration count — belong to the original B1, and a partial powering to one B1 is **not** interchangeable with any other B1 (the two use different prime sets). The old code silently recomputed the product at the current B1, after which the restart-file iteration count no longer matched the product length, tripping an assertion:
- `read_ppm1_savefiles`: `nsquares <= 0xFFFFFFFF && nsquares < 1.5*B1` (this is the one that actually fires first), or
- the `ilo == maxiter == PM1_S1_PROD_BITS` check in the main setup path.

The fix suggested in the issue thread — relax the assert and proceed to Stage 2 — is **unsafe**: a partial powering to the larger B1 is not a completed Stage 1 at the smaller B1, so jumping to Stage 2 silently loses factors.

### Fix
The `.s1_prod` product-savefile already records the B1 its product was built for, so nothing new needs to be written — the change is recovery-only: make that recorded B1 authoritative on restart.

`compute_pm1_s1_product()` now calls `read_pm1_s1_prod()` *before* it sizes anything from B1. `read_pm1_s1_prod()` compares the stored B1 against the current run's and, if they differ and the stored value is in range, adopts it (`B1 = b1`) and moves `B2_start` with it, so Stage 2's start-bound tracks the adopted Stage 1 bound and the run resumes *that* Stage 1 to completion. An out-of-range stored B1 is treated as a corrupt/foreign file and the product is recomputed at the current B1 as before. To instead start a fresh run at a different B1, delete the exponent's p-1 savefiles first (the code says so).

If the `.s1_prod` file is missing so the original B1 cannot be recovered, both the stage-1 restart-file bounds check (in `read_ppm1_savefiles()`) and the completed-S1 check (in `ernstMain()`) now fail with actionable guidance ("re-run with the original `-b1 <N>`, or delete this exponent's p-1 savefiles") instead of a bare assertion — never proceeding to Stage 2 on an incomplete powering.

### Verification
Built a baseline (pre-fix) and fixed binary from the same tree (via `git stash`), M65537 p-1 Stage 1:
- Interrupt mid-Stage-1 at **B1 = 30000** (restart file left at iteration ~42000), then restart at **B1 = 10000**:
  - **baseline** aborts (`read_ppm1_savefiles` nsquares bounds assert) — reproduces #31;
  - **fixed** logs the B1 adoption, resumes from the checkpoint, and completes Stage 1, reporting the result at the correct **B1 = 30000** (`{"status":"NF", ..., "B1":30000}`), matching a from-scratch full Stage 1 at B1 = 30000.
- **Same-B1 restart** (B1 = 30000) resumes normally, with no spurious adoption message — no regression.
- **`.s1_prod` deleted** restart prints the actionable error instead of a cryptic bounds assert.

### Scope of the diff
Three files: `src/pm1.c`, `src/Mlucas.c` and `src/Mlucas.h`.

- `pm1.c`: `read_pm1_s1_prod()`'s out-param changes from `uint64 arr[]` to `uint64 **arr` and the function now *owns* the `PM1_S1_PRODUCT` allocation (`*arr = ALLOC_UINT64(*arr, nlimbs);`), since only it knows the stored product's length; the B1-adoption logic lives there. `compute_pm1_s1_product()` is reordered so the read happens before the from-scratch sizing, and allocates only on the from-scratch path. Two new file-scope constants `PM1_B1_MIN`/`PM1_B1_MAX` replace the literals previously inline in `pm1_set_bounds()`, and bound the adopted value.
- `Mlucas.h`: the `read_pm1_s1_prod()` prototype.
- `Mlucas.c`: the two graceful-error conversions described above.

No behaviour change on a fresh run. On a matching-B1 restart the only difference is that `PM1_S1_PRODUCT` is now sized to the exact `nlimbs` read from the savefile rather than to the `((ebits+63)>>6)+1` estimate; the array is used only up to the stored product's length either way.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1785,8 +1785,18 @@ READ_RESTART_FILE:
 	int update_shift = (RES_SHIFT != 0ull);	// If shift = 0 at outset, don't update (only need for Fermat-mod, due to the random-bit aspect there)
 
 	if(TEST_TYPE == TEST_TYPE_PM1 && ilo >= maxiter) {
-		ASSERT(ilo == maxiter && ilo == PM1_S1_PROD_BITS,"For completed S1 expect ilo == maxiter == PM1_S1_PROD_BITS!");
-		snprintf(cbuf,sizeof(cbuf), "%s: p-1 stage 1 to b1 = %u already done -- proceeding to stage 2.\n",PSTRING,B1);
+		// Task 20/#31: For a genuinely-completed Stage 1 we expect ilo == maxiter == PM1_S1_PROD_BITS. If instead
+		// ilo *exceeds* the current-B1 product length, the savefile was produced with a larger B1 (e.g. a low-memory
+		// run bumps B1 up ~25%) and its '.s1_prod' product-savefile is missing, so compute_pm1_s1_product() could not
+		// recover/adopt the original B1. Do NOT proceed to Stage 2 - a partial powering to the larger B1 is not a
+		// completed Stage 1 at this B1, and treating it as such silently loses factors. Fail with actionable guidance.
+		if(ilo != maxiter || ilo != PM1_S1_PROD_BITS) {
+			snprintf(cbuf,STR_MAX_LEN*2, "ERROR: %s p-1 restart-file Stage 1 iteration count [%u] exceeds the Stage 1 prime-powers-product length [%u] for the current b1 = %u.\n"
+				"The savefile was written with a larger B1 and its '.s1_prod' product-savefile is missing, so the original B1 cannot be recovered automatically.\n"
+				"Re-run with the original (larger) '-b1 <N>', or delete this exponent's p-1 savefiles to start a fresh run.\n", PSTRING, ilo, PM1_S1_PROD_BITS, B1);
+			mlucas_fprint(cbuf,1);	ASSERT(0, "p-1 Stage 1 restart B1-mismatch: see preceding message.");
+		}
+		snprintf(cbuf,STR_MAX_LEN*2, "%s: p-1 stage 1 to b1 = %u already done -- proceeding to stage 2.\n",PSTRING,B1);
 		fprintf(stderr,"%s",cbuf);
 		ilo = ihi;		// Need this to differentiate between just-completed S1 and S1 residue read from restart file,
 		goto PM1_STAGE2;// in terms of whether we need to do a GCD before proceeding to S2
@@ -5233,7 +5243,18 @@ int read_ppm1_savefiles(const char *fname, uint64 p, uint32 *kblocks, FILE *fp, 
 			if(nsquares > 0xFFFFFFFFull)
 				ASSERT(B2_start <= nsquares, "P-1 stage 2 restart requires (B2_start in worktodo assignment) <= (savefile nsquares field)!");
 		} else {	// It's a stage 1 restart:
-			ASSERT(nsquares <= 0xFFFFFFFFull && nsquares < 1.5*(double)B1, "P-1 stage 1 restart: savefile nsquares value out of bounds!");
+			ASSERT(nsquares <= 0xFFFFFFFFull, "P-1 stage 1 restart: savefile nsquares value out of bounds!");
+			// Task 20/#31: nsquares >= ~1.5*B1 means the savefile's in-progress Stage 1 was run to a larger B1 than
+			// the current run's. compute_pm1_s1_product() normally recovers that B1 from the '.s1_prod' product-savefile
+			// and adopts it before we get here; if we still land here, that file is missing/unreadable, so the original
+			// B1 cannot be recovered. Fail with actionable guidance rather than the bare bounds-assert (a partial powering
+			// to the larger B1 is not a completed - nor a valid partial - Stage 1 at the smaller B1: proceeding loses factors).
+			if(nsquares >= 1.5*(double)B1) {
+				snprintf(cbuf,STR_MAX_LEN*2, "ERROR: %s: p-1 Stage 1 restart-file iteration count [%" PRIu64 "] is too large for the current b1 = %u.\n"
+					"The in-progress Stage 1 was run to a larger B1; its '.s1_prod' product-savefile (which records that B1) is missing or unreadable, so it cannot be recovered automatically.\n"
+					"Re-run with the original (larger) '-b1 <N>', or delete this exponent's p-1 savefiles to start a fresh run.\n", func, nsquares, B1);
+				mlucas_fprint(cbuf,1);	ASSERT(0, "p-1 Stage 1 restart B1-mismatch: see preceding message.");
+			}
 		}
 		// If S2 restart and (nsquares > B2_start), read the ensuing S2 interim residue; if (nsquares == B2_start)
 		// it means S2 started but was aborted for some reason before writing an interim S2 residue. That will set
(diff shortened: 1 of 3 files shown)


src/Mlucas.c
```diff
@@ -1775,7 +1775,17 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 	int update_shift = (RES_SHIFT != 0ull);	// If shift = 0 at outset, don't update (only need for Fermat-mod, due to the random-bit aspect there)
 
 	if(TEST_TYPE == TEST_TYPE_PM1 && ilo >= maxiter) {
-		ASSERT(ilo == maxiter && ilo == PM1_S1_PROD_BITS,"For completed S1 expect ilo == maxiter == PM1_S1_PROD_BITS!");
+		// Task 20/#31: For a genuinely-completed Stage 1 we expect ilo == maxiter == PM1_S1_PROD_BITS. If instead
```
tdulcet: Task 20 seems like some internal Claude Code reference that should be omitted from the public comment here. This also occurs several times below.

src/Mlucas.c
```diff
@@ -1775,7 +1775,17 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+		// ilo *exceeds* the current-B1 product length, the savefile was produced with a larger B1 (e.g. a low-memory
+		// run bumps B1 up ~25%) and its '.s1_prod' product-savefile is missing, so compute_pm1_s1_product() could not
+		// recover/adopt the original B1. Do NOT proceed to Stage 2 - a partial powering to the larger B1 is not a
+		// completed Stage 1 at this B1, and treating it as such silently loses factors. Fail with actionable guidance.
+		if(ilo != maxiter || ilo != PM1_S1_PROD_BITS) {
+			snprintf(cbuf,STR_MAX_LEN*2, "ERROR: %s p-1 restart-file Stage 1 iteration count [%u] exceeds the Stage 1 prime-powers-product length [%u] for the current b1 = %u.\n"
```
tdulcet: P-1 and B1 should be consistently capitalized in the error messages. This also occurs below.

src/pm1.c
```diff
@@ -333,12 +333,44 @@ global would be needed to store that - and remultiply by the appropriate one for
...
+			fclose(fp_s1p);
+			// Only adopt a sane, different B1 from a header whose type-tags match this run (guards against a truncated/foreign file):
+			if(!eof && test_types_compatible(c0,TEST_TYPE) && c1 == MODULUS_TYPE && fb1 >= 10000 && fb1 <= 2863311530u && fb1 != B1) {
+				sprintf(cbuf,"INFO: %s: current-run B1 [%u] differs from the in-progress Stage 1 savefile's B1 [%u]; adopting the savefile's B1 to safely resume that Stage 1 to completion.\n",func,B1,fb1);
+				mlucas_fprint(cbuf,pm1_standlone+1);
+				if(B2_start) B2_start = fb1;	// keep the Stage 2 start-bound tracking the adopted Stage 1 bound (Stage 2 begins where Stage 1 ends)
+				B1 = fb1;
+			}
+		}
+	}
```
tdulcet: Rather than duplicating this code and reading (part of) the `.s1_prod` savefile twice, could we just call the existing `read_pm1_s1_prod()` function sooner and updated it to restore the B1 value if needed.

tdulcet (changes requested):
Thanks for looking into this. I just have a couple of nits and a clarifying question.

src/pm1.c
MarkRose: It required a little refactoring, but yes. Done.

src/Mlucas.c
MarkRose: Cleaned up.

src/Mlucas.c
MarkRose: Also fixed capitalization on an old line.

src/pm1.c
```diff
@@ -483,29 +496,47 @@ int read_pm1_s1_prod(const char*fname, uint64 p, uint32*nbits, uint64 arr[], uin
...
 	}
 	if(B1 != b1) {
-		sprintf(cbuf, "INFO: %s: B1 of current run[%u] mismatches one[%u] of savefile data.\n",func,B1,b1);
-		goto PM1_S1P_READ_RETURN;
+		/* The savefile's Stage 1 was run to a different B1 than the current run's (pm1_set_bounds() auto-sizes B1 from
+		available RAM, so a restart under a different memory budget can pick a different B1). A partial powering to one
+		B1 is not interchangeable with any other, so adopt the savefile's B1 and resume that exact Stage 1 rather than
+		recomputing at a different B1. Only adopt a sane value; an out-of-range b1 (or the type-tag mismatches above)
+		means a truncated/foreign file, so bail and recompute at the current B1. */
+		if(b1 >= 10000 && b1 <= 2863311530u) {
```
tdulcet: Where does this `2863311530` value come from?

src/pm1.c
```diff
@@ -333,30 +333,43 @@ global would be needed to store that - and remultiply by the appropriate one for
 */
 uint32 compute_pm1_s1_product(const uint64 p) {
 	const double A = 1.1;
+	const char func[] = "compute_pm1_s1_product";
```
tdulcet: This variable does not seem to be used. (You could also just use `__func__` if needed.)

tdulcet (commented):
Thanks, this looks a lot better.

src/pm1.c
MarkRose: It's 2^33/3. The Stage 1 prime-powers product has ~1.5*B1 bits, and its bitlength is stored as a uint32, so B1 must satisfy 1.5*B1 <= 2^32, i.e. B1 <= 2^33/3 = 2863311530. I've pulled it out into a named `PM1_B1_MAX` constant (with `PM1_B1_MIN` for the paired 10000 lower bound) at the top of pm1.c with that derivation documented in one place, used by both `pm1_set_bounds()` and `read_pm1_s1_prod()` (6de7a56).

src/pm1.c
MarkRose: Good catch - removed. It was unused (`ASSERT` uses `__func__` directly).

MarkRose:

PR #76, opened 2026-07-10
Bound the preferred-FFT cfg scan to <= 2x default FFT length (fixes #36)

### LINKED ISSUE #36: Error: `Assertion failed: Call to get_preferred_fft_radix returns out-of-range FFT length`
The `get_preferred_fft_radix` function starts at the smallest supported FFT length for the exponent and then checks if any of the larger lengths listed in the `mlucas.cfg` file have a faster msec/iter timing. For the small FFT lengths or when using a large number of threads, sometimes a significantly larger length than the default will be the fastest, which will trigger this assertion: https://github.com/primesearch/Mlucas/blob/cfc33a8c2e68ef780df539ff92d692a0ce4afa31/src/Mlucas.c#L1319-L1324 The function need to be smarter to not return any FFT lengths that are too large for the exponent. However, this assertion is likely too conservative as well, as lengths greater than 2x the default would still work for many exponents.

This is a longstanding known issue, but was recently reported by Ken [on the forum](https://www.mersenneforum.org/node/22731?p=1052932#post1052932).

MarkRose:
Fixes #36.

## Problem (#36)

`get_preferred_fft_radix()` scans `mlucas.cfg` for any entry whose FFT length is `>= kblocks` (the target), with **no upper bound**. The caller in `ernstMain` (`Mlucas.c`) then asserts the returned length is `<= 2*kblocks`. So when the fastest cfg entry for an exponent is more than 2× the default FFT length, the scan returns it and the caller aborts:

```
Assertion `0' failed: Call to get_preferred_fft_radix returns out-of-range FFT length: asked for 16, returned 40
```

## Fix

- **Producer** (`get_preferred_fft_radix.c`): cap the scan at `i <= (kblocks<<1)`, so an out-of-range entry is never selected — and an in-range entry is not passed over in favour of an out-of-range one. The 2× cap itself is unchanged.
- **Caller** (`Mlucas.c`), defense-in-depth: instead of a bare `ASSERT`, warn and fall back to the existing "not found" path (request a fresh timing self-test for the length), so a stale or hand-edited cfg degrades gracefully rather than crashing.

## Verification (AVX2)

Crafted a cfg containing an entry `> 2×` the default length for M341749 (16K default):
- **unpatched**: aborts with the exact #36 assert (`out-of-range FFT length: asked for 16, returned 40`);
- **patched**: skips the oversized entry and runs normally.

`-s tiny` self-test passes.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1328,7 +1328,7 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 				a "not found" (dum == 0) result above, i.e. request a fresh timing self-test
 				for the requested length rather than trusting the out-of-range entry.
 				*/
-				snprintf(cbuf,STR_MAX_LEN*2,"WARN: get_preferred_fft_radix returned out-of-range FFT length: asked for %u, returned %u, packed value= %#8X -- ignoring and treating as 'not found' in '%s'; please rerun the self-test for this length.\n", kblocks, i, dum, CONFIGFILE);
+				snprintf(cbuf,sizeof(cbuf),"WARN: get_preferred_fft_radix returned out-of-range FFT length: asked for %u, returned %u, packed value= %#8X -- ignoring and treating as 'not found' in '%s'; please rerun the self-test for this length.\n", kblocks, i, dum, CONFIGFILE);
 				fprintf(stderr, "%s", cbuf);
 				if (!fft_length || MODULUS_TYPE == MODULUS_TYPE_MERSENNE) return ERR_RUN_SELFTEST_FORLENGTH + (kblocks << 8);
 			}
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1328,7 +1328,7 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 				a "not found" (dum == 0) result above, i.e. request a fresh timing self-test
 				for the requested length rather than trusting the out-of-range entry.
 				*/
-				sprintf(cbuf,"WARN: get_preferred_fft_radix returned out-of-range FFT length: asked for %u, returned %u, packed value= %#8X -- ignoring and treating as 'not found' in '%s'; please rerun the self-test for this length.\n", kblocks, i, dum, CONFIGFILE);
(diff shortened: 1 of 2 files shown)


tdulcet (dismissed):

src/Mlucas.c
```diff
@@ -1322,8 +1322,15 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
 			{
-				sprintf(cbuf,"Call to get_preferred_fft_radix returns out-of-range FFT length: asked for %u, returned %u, packed value= %#8X\n", kblocks, i, dum);
-				ASSERT(0, cbuf);
+				/* Should be unreachable now that get_preferred_fft_radix() itself bounds its
+				cfg-file scan to i <= 2*kblocks, but guard defensively rather than asserting/
+				crashing on a stale or hand-edited cfg file: warn and treat this the same as
+				a "not found" (dum == 0) result above, i.e. request a fresh timing self-test
+				for the requested length rather than trusting the out-of-range entry.
+				*/
+				snprintf(cbuf,STR_MAX_LEN*2,"WARN: get_preferred_fft_radix returned out-of-range FFT length: asked for %u, returned %u, packed value= %#8X -- ignoring and treating as 'not found' in '%s'; please rerun the self-test for this length.\n", kblocks, i, dum, CONFIGFILE);
```
tdulcet: We should use `sizeof(cbuf)` here, as we did in the other PRs.

MarkRose:
Done — `sizeof(cbuf)`. The hardcoded `STR_MAX_LEN*2` was wrong as well as brittle: `cbuf` is `STR_MAX_LEN*3`.

While in here I re-verified the underlying fix end to end, with a cfg that reproduces #36 — an exact-length 16K entry plus a *faster* 40K one, i.e. beyond the 2× bound the caller enforces:

| | result |
| --- | --- |
| unpatched `main` | `Assertion '0' failed` — `out-of-range FFT length: asked for 16, returned 40` |
| with this PR | completes the LL test of M341749: `{"exponent":341749, "worktype":"LL", "fft-length":18432}` |

18432 = 18K, within 2× of the 16K default — so the capped scan didn't just avoid the abort, it selected a usable in-range entry rather than passing over it in favour of the out-of-range one. That second part is what the PR claims and I hadn't previously shown directly.

One note for anyone reproducing it: the assert needs **both** an exact-length entry and a faster out-of-range one. With no 16K entry at all the scan returns "not found", the caller requests a timing self-test, that succeeds, and the bug never appears — which cost me a couple of false negatives before I worked it out.

tdulcet (approved):
Thanks.

PR #79, opened 2026-07-10
Align threadpool worker signatures with the run_func type (fixes #29)

### LINKED ISSUE #29: Undefined behavior on Windows: "call to function X through pointer to incorrect function type"

MarkRose:
Fixes #29.

## Problem (#29)

The threadpool declares its worker slot as:
```c
typedef void (*run_func)(void *data, int thread_num);
```
but the `mers`/`fermat`/`cyNN`_process_chunk workers were defined as `void *fn(void *targ)` and registered into `run_func` slots — a call through a function pointer of **incompatible type**, which is undefined behavior (and flagged by UBSan / `-Wcast-function-type`).

## Fix

Change the worker declarations, definitions, and their `task_control_t` registrations to match `run_func` exactly: `void fn(void *targ, int thread_num)`. No `threadpool.c/.h` change is needed — the workers now conform to the pool's existing contract. (39 files: `Mlucas.h`, the `mers/fermat/pm1/factor` dispatch sites, and each `radix*_ditN_cy_dif1.c`.)

## Verification (AVX2)

- Builds clean.
- Multithreaded (4-thread) self-test residues correct: radix-256 **byte-identical to baseline** (`95AFD7A5269F14F6`), and `-s tiny` passes all 32 lengths multithreaded.


tdulcet (dismissed):

src/Mlucas.h
```diff
@@ -369,60 +369,60 @@ void	radix32_dit_pass	(double a[], int n, struct complex rt0[], struct complex r
 
 #ifdef MULTITHREAD
 	/* Multithreaded version must be in form of 1-arg functor */
-	void *mers_process_chunk  (void *targ);
-	void *fermat_process_chunk(void *targ);
+	void mers_process_chunk(void*targ, int thread_num);
+	void fermat_process_chunk(void*targ, int thread_num);
 	// These are shared by both mers and fermat-mod, although the code contains switches to invoke the corr. carry macros:
-	void *cy12_process_chunk(void *targ);
```
tdulcet, line 375: The spaces after `void` in this file did improve readability.

tdulcet (dismissed):
Thanks.

MarkRose:
Good catch — restored. Scope is exactly the 53 declaration lines this PR reflowed, all in `Mlucas.h`; `git diff -w` against the previous tip is empty, so it's whitespace-only.

Left alone deliberately: the definitions in the `.c` files were already `void*targ` before this PR (that's the existing style throughout), as is `void*thread_arg` in `factor.c`/`factor.h`. Only the header declarations had the spacing that got collapsed.

Builds clean and `-s tt -cpu 0:3` matches plain main exactly — rc=0, 19 residues, including the same three pre-existing roundoff skips at 1K/2K under AVX2. Added as a separate commit so your approval stays intact.

tdulcet (approved):
Thanks.

PR #77, opened 2026-07-10
Relax the Gerbicz-check carry-out assert for p == 63 (mod 64) (fixes #33)

MarkRose:
## Problem (#33)

The periodic Gerbicz check computes `d[]*PRP_BASE` via `mi64_mul_scalar` and asserted the carry-out was exactly 0. For exponents `p == 63 (mod 64)` the top residue limb has only 1 spare bit, so a carry-out of up to `PRP_BASE-1` is a **normal, legal** outcome — the mod-reduction loop just below folds it back in. The strict assert crashed otherwise-healthy runs (~1/3 of Gerbicz checks for such exponents). All three reported exponents are `== 63 (mod 64)`.

## Fix

One line: `ASSERT(c_uint64_ptr[j] == 0ull, …)` → `ASSERT(c_uint64_ptr[j] < itmp64, …)` (carry must be < base).

## Verification (AVX2)

PRP test of p=1000639 (prime, `== 63 (mod 64)`), Gerbicz interval temporarily lowered to force many checks: **7 of 24 checks showed a real carry-out of 1** (each of which the old assert would have crashed); all passed with the relaxed assert. A full-cadence run reached 100% with no assert firing.

_Note: verifying this fix surfaced a separate, pre-existing bug — a corrupted carryout in `mi64_mul_scalar_add_vec2`'s asm that crashes the final PRP-determination (`Mlucas.c` `rmodb == 0` assert after short-div). Root-caused and fixed in #80; unrelated to this change._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fixes #33.

Issue #33, opened 2024-06-13
Assertion failed: d[]*PRP_BASE result has unexpected carryout!


efvb:
I am using code from main, May 5th, compiled with GCC 11 with tuned Makefile with -O3 -flto=auto -DUSE_ARM_V8_SIMD -mcpu=neoverse-n1
This is an Ampere Arm VM.

I also have the file to restart, if needed.

    Mlucas 21.0.1

    https://www.mersenneforum.org/mayer/README.html

INFO: testing qfloat routines...
System total RAM = 23236, free RAM = 11262
INFO: 11262 MB of free system RAM detected.
CPU Family = ARM Embedded ABI, OS = Linux, 64-bit Version, compiled with Gnu C [or other compatible], Version 11.4.1 20230605 (Red Hat 11.4.1-2.1.0.1).
INFO: Build uses ARMv8 advanced-SIMD instruction set.
INFO: Using prefetch.
INFO: Using inline-macro form of MUL_LOHI64.
INFO: Using FMADD-based 100-bit modmul routines for factoring.
INFO: MLUCAS_PATH is set to ""
INFO: using 53-bit-significand form of floating-double rounding constant for scalar-mode DNINT emulation.
INFO: testing IMUL routines...
INFO: Testing 64-bit 2^p (mod q) functions with 100000 random (p, q odd) pairs...
INFO: System has 4 available processor cores.
INFO: testing FFT radix tables...
Set affinity for the following 4 cores: 0.1.2.3.
User did not set LowMem in mlucas.ini ... allowing all test types.
User did not set CheckInterval in mlucas.ini ... using default.
NTHREADS = 4
Setting ITERS_BETWEEN_CHECKPOINTS = 10000.
 looking for worktodo.txt file...
 worktodo.txt file found...reading next assignment...
 worktodo.txt entry: PRP=16A20FCE71DF730C49E6B299E558CC11,1,2,11396863,-1,99,0,3,5,"61547191186328636909209"

INFO: Maximum recommended exponent for FFT length (576 Kdbl) = 11450805; p[ = 11396863]/pmax_rec = 0.9952892395.
Initial DWT-multipliers chain length = [hiacc] in carry step.
 INFO: restart file p11396863 found...reading...
Computing 2990000-squaring residue R (mod known prime q = 61547191186328636909209)
        A: R == 44566622423702911301179 (mod q)
        B: R == 44566622423702911301179 (mod q)
mers_mod_square: Init threadpool of 4 threads
[...]

PR #80, opened 2026-07-10
Fix corrupted carryout in mi64_mul_scalar_add_vec2 asm (crashes final PRP-determination)

MarkRose:
## Problem

Every completed Mersenne PRP test on a recent compiler aborts at the very end:
```
Assertion `rmodb == 0ull' failed: After short-div, R != 0 (mod B)
```
in the Suyama final-residue extraction (`Mlucas.c`), which divides `residue + k·m` by `PRP_BASE^2` and asserts the division is exact.

## Root cause

The x86-64 `YES_ASM` path of `mi64_mul_scalar_add_vec2` (`mi64.c`) ends with `adcq %%r9,%[__cy]`, which **reads** the `cy` memory operand — but `cy` was declared **`"=m"` (write-only)**. The compiler is not required to materialize the callers `cy = 0` init into that stack slot, so the `adcq` adds `r9 + CF` to leftover **stack garbage**, yielding a spurious nonzero carryout. The low result limbs are always correct; only the returned carryout is wrong.

That bogus carryout breaks the `residue + k·m` divisibility-by-`b²` invariant, tripping the assert. Instrumentation confirmed: the real limbs of `r + k·m` are divisible by `b²`, but the carryout limb was `4` (should be `0`), and `4·(2^64)^j ≡ 1 (mod 9)` exactly produced the observed remainder.

Notes:
- **Not `p mod 64`-specific** — reproduced on `p=250007` (≡23) and `p=250687`/`p=1000639` (≡63). It escaped testing because the `-s` self-tests use `-iters` (partial) and never reach this block; only a full PRP-to-completion does.
- Latent UB, exposed by newer compilers (hit on **GCC 16**); older toolchains happened to zero the slot.
- The protective `rmodb==0` assert makes the PRP path fail **loud** (crash), not silently report a wrong prime/composite verdict. The other caller — `mi64_mul_vector` (`mi64_div_mont`, used in cofactor/factoring paths) — is a genuinely-silent vector this fix also closes.

## Fix

Declare `cy` as **`"+m"`** (read-modify-write) so its `0`-init is honored — matching the already-correct sibling asm at `mi64.c:2374`. Only corrects the buggy case; where it already worked, behavior is unchanged. Two lines.

## Verification (AVX2, GCC 16)

- asm path + fix → `p=250007` and `p=250687` complete with `rmodb==0` and `Res64` **byte-identical to a C-fallback (`#ifndef YES_ASM`) reference build** (`E68EB51E76CFFDF2` / `BB61BB1229A1469A`), ROE 0.
- `-s tiny` regression: 93 tests, 0 failures.
- P-1 of M2000003 finds the known factor 160000241 (GCD path).
- PRP-cofactor runs cleanly through the `mi64_mul_vector` known-factor setup.

Found while verifying the #33 fix (#77).

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -2835,8 +2835,8 @@ uint64	mi64_mul_scalar_add_vec2(const uint64 x[], uint64 a, const uint64 y[], ui
 		"leaq	0x8(%%rsi), %%rsi	\n\t"\
 	"decq	%%rcx \n\t"\
 	"jnz 0b 	\n\t"/* loop1 end; continue is via jump-back if rcx != 0 */\
-		"adcq	%%r9 ,%[__cy]	\n\t"/* Carryout. */\
-		: [__cy] "=m" (cy) /* outputs: cy */\
+		"adcq	%%r9 ,%[__cy]	\n\t"/* Carryout: cy (pre-set to 0 by caller) += r9 + final CF. MUST be "+m", not "=m": this adcq READS cy, so a write-only "=m" lets the compiler leave that memory uninitialized -> spurious nonzero carryout. */\
+		: [__cy] "+m" (cy) /* outputs: cy (read-modify-write) */\
 		: [__x0] "g" (x)	/* All inputs from memory/register here */\
 		 ,[__y0] "g" (y)	\
 		 ,[__z0] "g" (z)	\


tdulcet (approved):
Thanks for the fix. We fixed a similar issue in #42. Maybe we should audit the codebase to make sure that no more of these cases exist.

I will leave the PR open for a few days in case anyone else wants to review.

MarkRose:
> Maybe we should audit the codebase to make sure that no more of these cases exist.

I had Claude do an audit for carryout bugs of this category and no more were found.

It did, however, find more of the #84 class in qfloat.c.

PR #81, opened 2026-07-10
Fix Mfactor build break: disable the never-implemented USE_FMADD 100-bit modpow path

MarkRose:
## Problem

The standalone **Mfactor** build fails on GCC 14+/C23:
```
factor.c: error: implicit declaration of function `twopmodq100_2WORD_DOUBLE_q4`
```

## Root cause

The `#ifdef USE_FMADD` sites in `factor.c` (for `TRYQ` 1/2/4) call `twopmodq100_2WORD_DOUBLE` / `_q2` / `_q4`, but **that FMADD 100-bit modpow family was never implemented**: only an AVX-512-only `_q32` is defined (plus a `_q64` *declaration*), and neither is ever called. So those calls always referenced nonexistent functions — a **link** error historically, and (because `makemake.sh`’s `mfac` target hardcodes `-DTRYQ=4`) a hard **compile** error once modern GCC upgraded implicit declarations from warning to error. (`factor_test.h`’s "FMADD" self-test calls are already inert — it `#undef`s `USE_FMADD` at its top.)

Mfactor-only: `factor.c` is in `OBJS_MFAC`, not the main Mlucas binary — which is why it went unnoticed, and CI masks Mfactor build failures via `makemake.sh mfac … || true`.

## Fix

Replace the three `#ifdef USE_FMADD` with `#if 0` so each `TRYQ` case falls through to the already-correct **integer `twopmodq*` path** — the exact routines the non-FMADD build and the `TRYQ==4` self-test already use. **Provably lossless**: no FMADD 100-bit path ever worked, so nothing functional is removed (there was simply never any FMA speed-up available for these batch sizes). `factor.c` only, well-commented.

## Verification (GCC 16)

- Unfixed tree fails at the `_q4` call; fixed tree **compiles clean**.
- `Mfactor` `test_fac` → "Factoring self-tests completed successfully", including the `TRYQ==4` integer-`_q4` block the fallback routes to.

## Not covered by this PR (separate, pre-existing)

- A pre-existing **LTO asm-label collision** in `twopmodq96.c` (`LoopBeg`/`LoopEnd`/`twopmodq96_q4_pshiftjmp` collide under parallel `-flto`) still blocks the Mfactor `-flto` **link** stage. This PR clears the compile error; that link bug is tracked separately.
- The Mfactor `test_fac` PRP self-test (M607) also needs the `mi64_mul_scalar_add_vec2` carryout fix in #80.

### DIFF
--- a/src/factor.c
+++ b/src/factor.c
@@ -3482,7 +3482,7 @@ MFACTOR_HELP:
 
 						  #else
 
-							#if 0	/* USE_FMADD has no single-candidate twopmodq100_2WORD_DOUBLE
+							#if defined(USE_FMADD) && 0	/* USE_FMADD has no single-candidate twopmodq100_2WORD_DOUBLE
 									variant - the only FMADD-based 100-bit modpow ever implemented is the
 									AVX-512-only, 32-way twopmodq100_2WORD_DOUBLE_q32 in twopmodq100.c (unused
 									by this TRYQ==1 code path, and in any case unreachable here since factor.h
@@ -3540,7 +3540,7 @@ MFACTOR_HELP:
 
 						  #elif(defined(P1WORD))
 
-							#if 0	/* USE_FMADD has no 2-way twopmodq100_2WORD_DOUBLE_q2
+							#if defined(USE_FMADD) && 0	/* USE_FMADD has no 2-way twopmodq100_2WORD_DOUBLE_q2
 									variant - see the longer note at the TRYQ==1 case above. There is no
 									integer-based q2 batch routine either, so unlike TRYQ==1/4 there is no
 									fallback to fall through to; if this is ever hit for real (it currently
@@ -3589,7 +3589,7 @@ MFACTOR_HELP:
 
 						  #else	// Default single-word-p mode:
 
-							#if 0	/* USE_FMADD has no 4-way twopmodq100_2WORD_DOUBLE_q4
+							#if defined(USE_FMADD) && 0	/* USE_FMADD has no 4-way twopmodq100_2WORD_DOUBLE_q4
 									variant - see the longer note at the TRYQ==1 case above (only a
 									never-called, AVX-512-only 32-way _q32 routine was ever completed for
 									this 100-bit FMADD family). This #if 0 previously read '#ifdef USE_FMADD'
--- a/src/factor.c
+++ b/src/factor.c
@@ -3482,8 +3482,12 @@ MFACTOR_HELP:
 
 						  #else
 
-							#ifdef USE_FMADD
-								/* Use 50x50-bit FMADD-based modmul routines, if def'd: */
+							#if 0	/* USE_FMADD has no single-candidate twopmodq100_2WORD_DOUBLE
+									variant - the only FMADD-based 100-bit modpow ever implemented is the
+									AVX-512-only, 32-way twopmodq100_2WORD_DOUBLE_q32 in twopmodq100.c (unused
+									by this TRYQ==1 code path, and in any case unreachable here since factor.h
+									restricts USE_FMADD builds to TRYQ = 1, 2 or 4). Fall through to the
+									already-correct paths below instead of calling a nonexistent function. */
 								res = twopmodq100_2WORD_DOUBLE(p[0],k);
 							#elif(defined(USE_FLOAT))
 								/* Otherwise use 78-bit floating-double-based modmul: */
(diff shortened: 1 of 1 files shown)


src/factor.c
```diff
@@ -3482,8 +3482,12 @@ candidate factors that survive sieving.	*/
 
 						  #else
 
-							#ifdef USE_FMADD
-								/* Use 50x50-bit FMADD-based modmul routines, if def'd: */
+							#if 0	/* USE_FMADD has no single-candidate twopmodq100_2WORD_DOUBLE
```
tdulcet: Maybe make it more clear, in case these functions are ever implemented:
```suggestion
							#if defined(USE_FMADD) && 0	/* USE_FMADD has no single-candidate twopmodq100_2WORD_DOUBLE
```

tdulcet (commented):
Thanks for looking into this and the PR.

> CI masks Mfactor build failures via `makemake.sh mfac … || true`.

We should remove these masks once all of the Mfactor build issues are resolved.

src/factor.c
MarkRose: Done (d97df09) - applied it to all three parallel blocks (the single-candidate, 2-way and 4-way USE_FMADD 100-bit paths) so they're consistent. Behavior is unchanged since the guard still evaluates to 0, but it now reads as USE_FMADD-specific-and-disabled rather than an anonymous `#if 0`.

MarkRose:
Re-measured this against #193, which targets the same defect by a different mechanism, and **the
mechanism in this PR does not actually fix the AVX2 build.** Recommending this PR be closed in favour of
#193.

## The two mechanisms

| PR | mechanism |
|---|---|
| **#81** (this one) | disable the three call sites: `#ifdef USE_FMADD` → `#if defined(USE_FMADD) && 0`; `USE_FMADD` itself stays defined |
| **#193** | `#undef USE_FMADD` for the whole translation unit |

They look interchangeable and are not. `USE_FMADD` does more than select a modpow routine — the same
`!defined(USE_FMADD)` term gates the **declarations** of three variables that the integer fallback path
then uses, in `PerPass_tfSieve`:

* `factor.c:2631` → `double fbits_in_2p = targ->fbits_in_2p;` (2636)
* `factor.c:2696` → `double fbits_in_k = 0, fbits_in_q = 0;` (2701)
* `factor.c:3095` → `fbits_in_q = fbits_in_2p + fbits_in_k;` (3104)

Disabling only the call sites leaves `USE_FMADD` defined, so those guards stay false and `fbits_in_q` is
never declared — while the `#else` integer path that now compiles in its place uses it at `factor.c:3636`.
The missing-function error is simply replaced by a different compile error.

## Measurement

Both mechanisms synthesised onto **`upstream/main` @ `418f365`**, so nothing else differs, and compiled
with exactly the `factor.o` rule from `makemake.sh:553`
(`-DFACTOR_STANDALONE -DTRYQ=4`, plus the mode flags):

| tree | gcc 16.1.1 avx2 | gcc avx512 | clang 22.1.8 avx2 | clang avx512 |
|---|---|---|---|---|
| unfixed `main` | `error: implicit declaration of 'twopmodq100_2WORD_DOUBLE_q4'` | same | `error: call to undeclared function` | same |
| **#81's mechanism** | **`factor.c:3636:68: error: 'fbits_in_q' undeclared`** | OK | **`error: use of undeclared identifier 'fbits_in_q'`** | OK |
| **#193's mechanism** | OK | OK | OK | OK |

AVX-512 is clean under *both* because `factor.c:604-609` self-defines `USE_FLOAT` for `USE_AVX512`, so the
dispatch takes the `#elif defined(USE_FLOAT)` 78-bit branch, which never touches `fbits_in_q` and leaves
the guards consistently false.

**So AVX2 is the only mode where the two mechanisms diverge — and it is the mode this PR is meant to fix.**
That also explains the "Verified (GCC 16) … compiles clean" note in the original description: on a host
whose `makemake.sh` auto-detect lands on AVX-512, this PR does work. On AVX2 it does not.

Worth flagging that this was not caught by a build matrix that records only pass/fail per configuration if
that matrix happened to be run on an AVX-512 host — the failing cell is specifically
`{gcc,clang} × avx2 × Mfactor`.

## What #193 carries over from here

#193 has been updated to adopt the one thing this PR does better: the `TRYQ = 2 / P1WORD` `#error` no
longer offers `USE_FMADD` as an alternative (it is undef'd there), and instead names the missing
`twopmodq100_2WORD_DOUBLE_q2` and says to build with `TRYQ = 1` or `4`. It also now records inline *why*
the undef has to be file-scope rather than per-call-site, so this distinction does not get "simplified"
back out later.

## On the CI masks

> We should remove these masks once all of the Mfactor build issues are resolved.

Agreed. For the record on what is still in the way after #193: on clang, #193 alone unblocks the Mfactor
build end-to-end at every SIMD level. On gcc with the default (LTO-on) `CFLAGS` it clears the compile error
and then hits a separate pre-existing bug — `twopmodq96.c:706: Error: symbol 'LoopBeg' is already defined`,
the non-local inline-asm labels colliding when LTO merges translation units. The same pattern is in
`twopmodq.c`, `twopmodq80.c` and `twopmodq64_test.c`; that is #82's territory. So the mask can come off
once #193 and #82 are both in.

MarkRose:
Closing in favour of #193, per the measurement above: this PR's mechanism leaves `USE_FMADD` defined, so `fbits_in_q` stays undeclared and the AVX2 Mfactor build still fails — the very configuration this was meant to fix. #193 undef's it file-wide and is clean across {gcc,clang} × {avx2,avx512}, and has taken over this PR's improved `TRYQ = 2` `#error` wording.

The branch is kept, so this is trivially reopenable if the call turns out to be wrong.

PR #82, opened 2026-07-10
Make inline-asm labels asm-local to fix LTO symbol collisions in the Mfactor build

MarkRose:
> [!IMPORTANT]
> **Depends on #193 — must be merged after it.** This PR fixes an LTO *link*-stage failure that is only reachable once #193 clears the earlier `factor.c` *compile* error. On its own (without #193), the Mfactor build still fails earlier at `factor.c`, so this branch is intentionally not standalone-buildable — it is scoped to only the label fix. The two PRs are independent diffs (no shared commits) purely so each stays single-purpose.

## Problem

With #193 applied (Mfactor now compiles), `bash makemake.sh mfac` fails at the **LTO link** stage:
```
twopmodq96.c: Error: symbol `LoopBeg` is already defined
twopmodq96.c: Error: symbol `twopmodq96_q4_pshiftjmp` is already defined
twopmodq96.c: Error: symbol `LoopEnd` is already defined
```

## Root cause

Several `twopmodq*.c` inline-asm blocks use **hardcoded non-local labels** (`LoopBeg`, `LoopEnd`, `twopmodqNN_qM_pshiftjmp`, …). Under `-flto` parallel-partition codegen the containing functions are duplicated across partitions, so those labels become duplicate **global** symbols → link failure. (Same class as the parallel-LTO work in #56.) Mfactor-specific in practice; the main Mlucas binary already builds clean with the same files + `-flto`. It was doubly masked: the compile error (#193) aborted before link, and `ci.yml` runs `makemake.sh mfac … || true`.

## Fix

Append GCC’s `%=` (unique-per-instantiation token) to every label definition and jump reference within each affected asm block — **label renames only**, no instruction/operand/constraint changes, control flow identical. Files:
- `twopmodq96.c` — `twopmodq96_q4` block
- `twopmodq64_test.c` — `twopmodq64_q4` and `_q8` blocks
- `twopmodq80.c` — the two live (non-`USE_IMCI512`) q4/q8 blocks (one independently reused the name `twopmodq96_q4_pshiftjmp`)
- `twopmodq.c` — `twopmmodq64_q4` block

Sibling files (`twopmodq100/128/128_96/160/192/256.c`) were checked — no real non-local asm labels. Dead-code (`#if 0`) label duplicates left as-is.

## Verification (GCC 16, `-flto`)

Confirmed via A/B on this exact build: **reverting `twopmodq96.c`** to the unfixed version reproduces the exact `already defined` link errors; **with the fix**, `makemake.sh mfac` links clean and produces `obj_mfac/Mfactor`. Mfactor `test_fac` self-test passes (incl. the 96-bit factor test exercising `twopmodq96_q4`). _(The self-test’s M607 PRP step also needs the mi64 carryout fix in #80.)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,11 +65,9 @@ jobs:
                         else
                             # Exercise Mfactor's built-in self-tests (test_fac), which run during
                             # factoring-init and validate the modpow/sieve machinery. M(127) has no
-                            # factors <= 2^20 so the trial-factoring pass is quick. We tolerate the
-                            # run's own exit status and instead assert the self-test success marker,
-                            # so a test_fac regression fails the job.
-                            ./Mfactor${word:+_$word} -m 127 -bmax 20 2>&1 | tee "mfac_selftest_${word:-none}.log" || true
-                            grep -q 'self-tests completed successfully' "mfac_selftest_${word:-none}.log"
+                            # factors <= 2^20 so the trial-factoring pass is quick. Mfactor already
+                            # exits nonzero on a test_fac failure, so no extra check is needed here.
+                            ./Mfactor${word:+_$word} -m 127 -bmax 20
                         fi
                     fi
                 )
@@ -137,11 +135,9 @@ jobs:
                         else
                             # Exercise Mfactor's built-in self-tests (test_fac), which run during
                             # factoring-init and validate the modpow/sieve machinery. M(127) has no
-                            # factors <= 2^20 so the trial-factoring pass is quick. We tolerate the
-                            # run's own exit status and instead assert the self-test success marker,
-                            # so a test_fac regression fails the job.
-                            ./Mfactor${word:+_$word} -m 127 -bmax 20 2>&1 | tee "mfac_selftest_${word:-none}.log" || true
-                            grep -q 'self-tests completed successfully' "mfac_selftest_${word:-none}.log"
+                            # factors <= 2^20 so the trial-factoring pass is quick. Mfactor already
+                            # exits nonzero on a test_fac failure, so no extra check is needed here.
+                            ./Mfactor${word:+_$word} -m 127 -bmax 20
                         fi
                     fi
                 )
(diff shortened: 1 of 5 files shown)


tdulcet (dismissed):

> Mfactor `test_fac` self-test passes

I did not know that Mfactor had a self-test, but we should probably enable it in the CI to prevent regressions.

MarkRose:
Added a commit wiring Mfactor's built-in `test_fac()` self-test into CI (your suggestion above). Every Mfactor CI build previously ran only `-h`, which exits before the factoring-init path where `test_fac()` runs; the new step runs a quick `-m 127 -bmax 20` job (M127 has no factors ≤ 2^20, so the pass is near-instant) and asserts the "self-tests completed successfully" marker, tolerating the run's own exit status so a `test_fac` regression fails the job.

Doing this surfaced that an actual Mfactor factoring run *aborts* — two pre-existing bugs in the never-exercised checkpoint code (savefile corruption from read/write interleaving on an `"r+"` stream, plus a trivial `k=0`→`q=1` candidate that trips a PRP-test assert on tiny exponents). Those are fixed separately in #114; the self-test marker-check here is deliberately robust to that, so the two PRs are independent.

.github/workflows/ci.yml
```diff
@@ -57,7 +57,15 @@ jobs:
...
                     mv -v build.log build_${word}.log
-                    [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
+                    if [[ -x Mfactor${word:+_$word} ]]; then
+                        # Exercise Mfactor's built-in self-tests (test_fac), which run during
+                        # factoring-init and validate the modpow/sieve machinery. M(127) has no
+                        # factors <= 2^20 so the trial-factoring pass is quick. We tolerate the
+                        # run's own exit status and instead assert the self-test success marker,
+                        # so a test_fac regression fails the job.
+                        ./Mfactor${word:+_$word} -m 127 -bmax 20 2>&1 | tee "mfac_selftest_${word:-none}.log" || true
+                        grep -q 'self-tests completed successfully' "mfac_selftest_${word:-none}.log"
```
tdulcet: The `grep` is unnecessary, as Mfactor should already exit with a nonzero exit status if it fails:
```suggestion
                        ./Mfactor${word:+_$word} -m 127 -bmax 20
```
This also occurs below, in #83 and #103.

tdulcet (changes requested):
Thanks for adding the Mfactor self-test to the CI. That will be very helpful to prevent regressions.

Maybe we should combine this with the change requested in https://github.com/primesearch/Mlucas/pull/83#discussion_r3597177681, to avoid merge conflicts.

.github/workflows/ci.yml
tdulcet: See https://github.com/primesearch/Mlucas/pull/103#discussion_r3696024150.

tdulcet (approved):

PR #83, opened 2026-07-11
Windows build portability: probe-gate AVX-512, fix ARM detection, and Mfactor multiword+FMADD

MarkRose:
## Problem

The three **Windows** CI jobs fail on every PR. They are `continue-on-error: true` (non-blocking), but they show a persistent, confusing red X. Diagnosis of the *fatal* (non-`|| true`) step of each:

- **`windows-latest, gcc`** (exit 1): the main AVX-512 build aborts — MinGW gcc's assembler lacks the extended register names (`zmm16-31/xmm16-31/k0-7`): *"gcc's assembler does not support the AVX-512 extended register names … aborting"*. (The FMADD Mfactor `#error`s above it are `|| true`-masked noise, not the killer.)
- **`windows-11-arm, gcc`** (exit 1): the *same* AVX-512 abort — because the MSYS2 bash there is an x86_64 build under emulation, so `$HOSTTYPE` reports `x86_64`; the job takes the x86 path and tries to build AVX-512 with an ARM gcc.
- **`windows-latest, clang`** (exit 139): a SIGSEGV in the AVX-512 self-test — clang *can* assemble AVX-512, so it builds+runs it and hits the `#65`/`#16` register-clobber bug. **That is fixed by #68, not this PR.**

## Fixes (three independent root causes, one commit each)

1. **Probe-gate AVX-512 in the Windows job** (`ci.yml`) — gate the `AVX512` MODES entry on the same extended-register + `vrcp28pd` assembler probe used for the Linux-old job (mirrors the merged approach in #73), with a `::warning::` when skipped. MinGW gcc stops at AVX2; Windows clang still builds AVX-512.
2. **Detect ARM from the matrix OS label** (`ci.yml`) — replace `[[ $HOSTTYPE == aarch64 ]]` with `endsWith(matrix.os, '-arm')` (already used in the job for no_gmp/pkg-prefix) in the Before-script probe and the MODES selection, so `windows-11-arm` takes the ASIMD path instead of attempting x86 AVX-512. Linux/macOS keep `$HOSTTYPE` (accurate on native runners).
3. **Mfactor multiword-p + `USE_FMADD`** (`factor.h`) — FMADD 100-bit modmul is one-word-p only, so `factor.h` `#error`s on P2/P3/P4WORD + `USE_FMADD`. But `USE_FMADD` is auto-defined per-platform on any FMA target, so the standalone Mfactor multiword variants fail to compile on every FMA x86 machine (masked by `makemake.sh mfac … || true`). Since the combination arises automatically, `#undef USE_FMADD` for multiword and fall through to the standard modmul instead of erroring. This is the multiword counterpart to the one-word fix in **#193**.

## Masking note
On `windows-latest gcc` / `windows-11-arm gcc` the FMADD Mfactor errors are `|| true`-masked and never fatal; the AVX-512 abort in the (unmasked) main build is what kills the job. So commits 1 & 2 are what turn those two jobs green; commit 3 is build hygiene (and fixes the same Mfactor break on FMA Linux too). This complements #193 (one-word FMADD) and #82 (Mfactor LTO labels), and addresses the Windows-code-path follow-up raised in the #67 review.

## Verification
- **Commit 3 verified on Linux** (GCC 16, FMA): A/B — unfixed `makemake.sh mfac 2word` fails at `factor.h:105 #error P2WORD`; fixed builds `obj_mfac/Mfactor_2word` clean.
- **Commits 1 & 2 are CI-only and Windows-CI-gated** — there is no Windows toolchain to build on locally, so they are verified by reasoning + parity with the already-verified Linux-old probe (#73). Not claimed Windows-verified; the Windows CI on this PR is the gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -532,7 +532,11 @@ jobs:
       shell: bash
       run: |
         $CC --version
-        if [[ $HOSTTYPE == aarch64 ]]; then
+        # The windows-11-arm runner ships no native aarch64 toolchain (no MSYS2/CLANGARM64), so its
+        # "gcc" is really an x86_64-w64-mingw32 compiler running under emulation. Both $HOSTTYPE
+        # (x86_64) and the matrix OS label (arm) mislead here, so key the arch off what the compiler
+        # itself targets:
+        if [[ "$($CC -dM -E - </dev/null)" == *__aarch64__* ]]; then
             $CC -dM -E -mcpu=native - </dev/null
         else
             $CC -dM -E -march=native - </dev/null
@@ -542,7 +546,10 @@ jobs:
       run: |
         set -x
         MODES=( '' NOSIMD )
-        if [[ $HOSTTYPE == aarch64 ]]; then
+        # The windows-11-arm runner ships no native aarch64 toolchain, so its "gcc" actually targets
+        # x86_64 (run under emulation); detect the arch the compiler targets rather than the matrix OS
+        # label or $HOSTTYPE, both of which say the wrong thing on that runner:
+        if [[ "$($CC -dM -E - </dev/null)" == *__aarch64__* ]]; then
             MODES+=( ASIMD )
         else
             MODES+=( SSE2 AVX AVX2 AVX512 )
@@ -553,11 +560,12 @@ jobs:
             (cd obj${mode:+_$mode}; make clean)
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
-                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
+                # The windows-11-arm runner has no GMP installed (the Install step is x86-only), so the
+                # Mfactor build must also request no_gmp there, exactly as the main build above does:
+                bash -e -o pipefail -- makemake.sh mfac ${{ endsWith(matrix.os, '-arm') && 'no_gmp' || '' }} $word $mode || true
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
         done
@@ -567,6 +575,13 @@ jobs:
             cd ../obj_NOSIMD
         fi
         bash -- ../config-fermat.sh -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))"
+        # Mfactor self-test on the native build only. The explicit-ISA builds above are for compile
+        # coverage: a runner may not be able to execute what they target, which is exactly what the
+        # AVX-512 SIGILL here was - GitHub's Windows runners are predominantly AVX2-only. Same
+        # treatment #103 applies to the Linux and macOS jobs.
+        for word in '' 1word 2word 3word 4word nword; do
+            ( cd "../obj_mfac${word:+_$word}" && time "./Mfactor${word:+_$word}" -m 127 -bmax 20 )
+        done
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -736,7 +736,11 @@ jobs:
       shell: bash
       run: |
         $CC --version
-        if [[ $HOSTTYPE == aarch64 ]]; then
+        # The windows-11-arm runner ships no native aarch64 toolchain (no MSYS2/CLANGARM64), so its
+        # "gcc" is really an x86_64-w64-mingw32 compiler running under emulation. Both $HOSTTYPE
+        # (x86_64) and the matrix OS label (arm) mislead here, so key the arch off what the compiler
+        # itself targets:
+        if [[ "$($CC -dM -E - </dev/null)" == *__aarch64__* ]]; then
             $CC -dM -E -mcpu=native - </dev/null
         else
             $CC -dM -E -march=native - </dev/null
@@ -746,7 +750,10 @@ jobs:
       run: |
         set -x
         MODES=( '' NOSIMD )
-        if [[ $HOSTTYPE == aarch64 ]]; then
+        # The windows-11-arm runner ships no native aarch64 toolchain, so its "gcc" actually targets
+        # x86_64 (run under emulation); detect the arch the compiler targets rather than the matrix OS
+        # label or $HOSTTYPE, both of which say the wrong thing on that runner:
+        if [[ "$($CC -dM -E - </dev/null)" == *__aarch64__* ]]; then
             MODES+=( ASIMD )
         else
             MODES+=( SSE2 AVX AVX2 AVX512 )
@@ -757,7 +764,9 @@ jobs:
             (cd obj${mode:+_$mode}; make clean)
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
-                bash -e -o pipefail -- makemake.sh mfac $word $mode
+                # The windows-11-arm runner has no GMP installed (the Install step is x86-only), so the
+                # Mfactor build must also request no_gmp there, exactly as the main build above does:
+                bash -e -o pipefail -- makemake.sh mfac ${{ endsWith(matrix.os, '-arm') && 'no_gmp' || '' }} $word $mode
                 (cd obj_mfac${word:+_$word}${mode:+_$mode}; make clean)
             done
         done
--- a/makemake.sh
+++ b/makemake.sh
@@ -155,6 +155,29 @@ if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
 fi
 MAKE_ARGS+=(-j "$CPU_THREADS")
 
+# Windows/MinGW gcc (observed on both 15.2 and 16.1) has a longstanding x86-64 codegen bug: when
+# autovectorization creates 256/512-bit temporaries, gcc spills them with ALIGNED stores (vmovdqa,
+# needing 32/64-byte alignment) into stack frames it never dynamically realigns - but the Win64 ABI
+# only guarantees 16-byte alignment at function entry, and Windows randomizes the initial thread
+# stack phase per run. Result: plain C code (no inline asm involved) faults on ~half of all runs,
+# at a stable instruction but "moving" whenever the code is recompiled - observed as Mfactor's
+# test_fac() SIGSEGVing intermittently in CI (vmovdqa %ymm0,0x170(%rsp) with rsp = 0 mod 32, where
+# gcc assumed 16 mod 32). Linux and Wine always start with the compatible phase, which is why this
+# never reproduced off real Windows. -mstackrealign does NOT help (it realigns only to the 16-byte
+# preferred boundary). Instead, keep compiler-GENERATED vector code to 128 bits so no 32/64-byte-
+# aligned spill slots exist at all; Mlucas's hand-written SIMD asm is unaffected (the assembler
+# needs no -m flags), so AVX2/AVX-512 build modes lose nothing but gcc's autovectorization width:
+if [[ $OSTYPE == msys || $OSTYPE == cygwin ]] && ! "${CC:-gcc}" --version 2>/dev/null | grep -qi clang; then
+	pvw_tmp=$(mktemp -d)
+	printf 'int main(void){return 0;}\n' >"$pvw_tmp/t.c"
+	# Compile a real temp file to a real output: native MinGW gcc misparses '-o /dev/null' as C:\dev\null.
+	if "${CC:-gcc}" -mprefer-vector-width=128 "$pvw_tmp/t.c" -o "$pvw_tmp/t.out" >/dev/null 2>&1; then
+		echo -e "MinGW gcc detected: adding -mprefer-vector-width=128 to work around gcc's unaligned-AVX-spill bug on Windows.\n"
+		ARGS+=(-mprefer-vector-width=128)
+	fi
+	rm -rf "$pvw_tmp"
+fi
+
 # $0 contains script-name, but $@ starts with first ensuing cmd-line arg, if it exists:
 echo "Total number of input parameters = $#"
 
--- a/makemake.sh
+++ b/makemake.sh
@@ -165,6 +165,29 @@ if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
 fi
 MAKE_ARGS+=(-j "$CPU_THREADS")
 
+# Windows/MinGW gcc (observed on both 15.2 and 16.1) has a longstanding x86-64 codegen bug: when
+# autovectorization creates 256/512-bit temporaries, gcc spills them with ALIGNED stores (vmovdqa,
+# needing 32/64-byte alignment) into stack frames it never dynamically realigns - but the Win64 ABI
+# only guarantees 16-byte alignment at function entry, and Windows randomizes the initial thread
+# stack phase per run. Result: plain C code (no inline asm involved) faults on ~half of all runs,
+# at a stable instruction but "moving" whenever the code is recompiled - observed as Mfactor's
+# test_fac() SIGSEGVing intermittently in CI (vmovdqa %ymm0,0x170(%rsp) with rsp = 0 mod 32, where
+# gcc assumed 16 mod 32). Linux and Wine always start with the compatible phase, which is why this
+# never reproduced off real Windows. -mstackrealign does NOT help (it realigns only to the 16-byte
+# preferred boundary). Instead, keep compiler-GENERATED vector code to 128 bits so no 32/64-byte-
+# aligned spill slots exist at all; Mlucas's hand-written SIMD asm is unaffected (the assembler
+# needs no -m flags), so AVX2/AVX-512 build modes lose nothing but gcc's autovectorization width:
+if [[ $OSTYPE == msys || $OSTYPE == cygwin ]] && ! "${CC:-gcc}" --version 2>/dev/null | grep -qi clang; then
+	pvw_tmp=$(mktemp -d)
+	printf 'int main(void){return 0;}\n' >"$pvw_tmp/t.c"
+	# Compile a real temp file to a real output: native MinGW gcc misparses '-o /dev/null' as C:\dev\null.
+	if "${CC:-gcc}" -mprefer-vector-width=128 "$pvw_tmp/t.c" -o "$pvw_tmp/t.out" >/dev/null 2>&1; then
+		echo -e "MinGW gcc detected: adding -mprefer-vector-width=128 to work around gcc's unaligned-AVX-spill bug on Windows.\n"
+		ARGS+=(-mprefer-vector-width=128)
+	fi
+	rm -rf "$pvw_tmp"
+fi
+
 # $0 contains script-name, but $@ starts with first ensuing cmd-line arg, if it exists:
 echo "Total number of input parameters = $#"
 
--- a/makemake.sh
+++ b/makemake.sh
@@ -165,18 +165,32 @@ if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
 fi
 MAKE_ARGS+=(-j "$CPU_THREADS")
 
-# Windows/MinGW gcc (observed on both 15.2 and 16.1) has a longstanding x86-64 codegen bug: when
-# autovectorization creates 256/512-bit temporaries, gcc spills them with ALIGNED stores (vmovdqa,
-# needing 32/64-byte alignment) into stack frames it never dynamically realigns - but the Win64 ABI
-# only guarantees 16-byte alignment at function entry, and Windows randomizes the initial thread
-# stack phase per run. Result: plain C code (no inline asm involved) faults on ~half of all runs,
-# at a stable instruction but "moving" whenever the code is recompiled - observed as Mfactor's
+# Windows/MinGW gcc (observed on both 15.2 and 16.1) has a longstanding x86-64 codegen bug - GCC PR
+# 54412, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412, open since 2012 and still unfixed as of
+# gcc 16: when autovectorization creates 256/512-bit temporaries, gcc spills them with ALIGNED stores
+# (vmovdqa, needing 32/64-byte alignment) into stack frames it never dynamically realigns - but the
+# Win64 ABI only guarantees 16-byte alignment at function entry, and Windows randomizes the initial
+# thread stack phase per run. Result: plain C code (no inline asm involved) faults on ~half of all
+# runs, at a stable instruction but "moving" whenever the code is recompiled - observed as Mfactor's
 # test_fac() SIGSEGVing intermittently in CI (vmovdqa %ymm0,0x170(%rsp) with rsp = 0 mod 32, where
 # gcc assumed 16 mod 32). Linux and Wine always start with the compatible phase, which is why this
 # never reproduced off real Windows. -mstackrealign does NOT help (it realigns only to the 16-byte
 # preferred boundary). Instead, keep compiler-GENERATED vector code to 128 bits so no 32/64-byte-
 # aligned spill slots exist at all; Mlucas's hand-written SIMD asm is unaffected (the assembler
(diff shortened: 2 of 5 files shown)


MarkRose:
_Analysis below performed by Claude Code (Anthropic's AI CLI) on my behalf; I've reviewed it._

Heads-up on the Windows CI for this PR: the `windows-latest, gcc` job will not go green from this PR alone — **it now depends on #45 merging.**

The reason is that this PR fixes the *first* Windows-gcc failure (the AVX-512 build abort) by probe-gating the AVX-512 mode. That's confirmed working — the job now emits the `::warning::` and skips AVX-512. But skipping it lets the build get *further* than it ever did before: it reaches the `-s t` self-test, which then trips the pre-existing thread-local memcheck assert from #8:

```
ERROR: Function radix16_ditN_cy_dif1, at line 1444 of file ../src/radix16_ditN_cy_dif1.c
Assertion 'tdat[ithread].wt0 == wt0' failed: thread-local memcheck fail!
```

(16K FFT, radices 16/16/32; the step then exits 127.) That assert is exactly what **#45** fixes, and I've verified #45 covers every affected radix. So the merge order is: this PR + #45 together green the Windows x86-64 gcc job.

For completeness, the other two Windows jobs fail for separate reasons and are **not** resolved by #45:
- `windows-11-arm, gcc` — the runner's `gcc` is `x86_64-w64-mingw32` (x86 under emulation), so forcing the ARM/ASIMD path makes it try to assemble ARM registers (`dft_macro.c: unknown register name 'x15'`). This needs a toolchain/CI decision (native `clangarm64` vs. detecting the compiler's target arch) rather than a source fix.
- `windows-latest, clang` — a separate segfault (looks like the teensy-self-test / #7 class discussed on #45), not #8.

MarkRose:
_Analysis by Claude Code (Anthropic's AI CLI) on my behalf; I've reviewed it._

Pushed e1945de to fix the build/execution failures on the other two Windows jobs. CI run confirms both are fixed — all three Windows jobs now build and advance to the self-test:

- **windows-11-arm (gcc)** — was failing to compile (exit 1): the runner's `gcc` is `x86_64-w64-mingw32` (x86 under emulation), and keying the SIMD path off the OS label forced the ARM/ASIMD build, which the x86 compiler can't assemble (`unknown register name 'x15'`). Now detects the compiler's actual target arch, so it builds the x86 modes it can compile. → now reaches the self-test (exit 127).
- **windows-latest (clang)** — was SIGILL'ing (exit 132): clang *builds* AVX-512, but the runner CPU is AVX2-only, so running the AVX-512 `Mfactor -h` faulted. Now builds AVX-512 for compile-time coverage but skips *executing* its binaries, with a CI warning. → now reaches the self-test (exit 139).

The remaining failures are pre-existing runtime bugs, each with its own fix PR:

- **both gcc jobs** fail at `-s t` with the thread-local memcheck assert (`radix16_ditN_cy_dif1.c:1444`) → **#8**, fixed by **#45**.
- **clang** segfaults earlier, at the teensy self-test `-s tt` (1K FFT, radices 32×16) → **#7**, fixed by **#69**.

So the Windows matrix goes fully green once **#45** and **#69** merge alongside this PR. (These jobs are `continue-on-error`, so they don't block the run in the meantime.)

.github/workflows/ci.yml
```diff
@@ -540,10 +544,33 @@ jobs:
...
+                if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm0,%%zmm0,%%zmm0":::"zmm0");return 0;}' | "${CC:-gcc}" -mavx512f -x c -o conftest_avx512run - >/dev/null 2>&1 && ./conftest_avx512run >/dev/null 2>&1; then
+                    run_avx512=1
+                else
+                    run_avx512=0
+                    echo "::warning title=AVX-512 execution skipped::$("${CC:-gcc}" --version | head -n1): this runner's CPU cannot execute AVX-512; building the AVX-512 mode for compile coverage but not running its test binaries."
+                fi
+                rm -f conftest_avx512run conftest_avx512run.exe
+            else
+                echo "::warning title=AVX-512 build skipped::$("${CC:-gcc}" --version | head -n1): its assembler cannot handle the AVX-512 constructs Mlucas's kernels need; skipping the AVX-512 build mode."
+            fi
```
tdulcet: The analysis here is incorrect. If you look at the CI for other PRs, the AVX512 builds compile fine on Windows. The issue is that the changes to the `makemake.sh` script in #67 are not working on Windows. I would suggest fixing the problem there.

src/factor.h
```diff
@@ -100,15 +100,13 @@ PLEASE REFER TO FACTOR.C FOR A DESCRIPTION OF THE APPLICABLE #DEFINES
...
 
-	/* FMADD-based modmul currently only supported for one-word p's: */
-	#ifdef P2WORD
-		#error P2WORD may not be used together with USE_FMADD!
-	#endif
-	#ifdef P3WORD
-		#error P3WORD may not be used together with USE_FMADD!
-	#endif
-	#ifdef P4WORD
-		#error P4WORD may not be used together with USE_FMADD!
```
tdulcet, line 111: Maybe downgrade the `#error`'s to `#warn`, so it is clear that the `USE_FMA` define is not being used and there will be no performance improvement.

tdulcet (changes requested):
Thanks for the PR. It looks like we should be able to simplify it.

.github/workflows/ci.yml
MarkRose: You're right - thanks for the correction. I've dropped the assembler-capability probe and now always build the AVX512 mode on x86 (MinGW's assembler handles the extended registers fine). The underlying makemake.sh issue is fixed in #67 (7cf3f13), which now runs `try_avx512_asm` on the Windows/fallback auto-detect path too. Kept only the runtime execution-skip here, since the AVX2-only runner CPUs still SIGILL on the AVX-512 test binaries.

src/factor.h
MarkRose: Done - it now emits a `#warning` making clear that USE_FMADD is disabled for multiword `p` and the build gets no FMADD speedup, right before the `#undef`.

.github/workflows/ci.yml
```diff
@@ -540,23 +544,46 @@ jobs:
...
+            # GitHub's Windows runner CPUs are AVX2-only, so the built AVX-512 binaries hit SIGILL when
+            # run. Probe whether baseline AVX-512 actually executes here; if not, still build the mode
+            # (compile-time coverage) but skip running its test binaries below.
+            if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm0,%%zmm0,%%zmm0":::"zmm0");return 0;}' | "${CC:-gcc}" -mavx512f -x c -o conftest_avx512run - >/dev/null 2>&1 && ./conftest_avx512run >/dev/null 2>&1; then
+                run_avx512=1
+            else
+                run_avx512=0
+                echo "::warning title=AVX-512 execution skipped::$("${CC:-gcc}" --version | head -n1): this runner's CPU cannot execute AVX-512; building the AVX-512 mode for compile coverage but not running its test binaries."
+            fi
+            rm -f conftest_avx512run conftest_avx512run.exe
```
tdulcet: It should only be executing the native build, so rather than all of this complexity, just add something like this near the end of the step, after the self-test:
```bash
cd ../obj_mfac
for word in '' {{1..4},n}word; do ./Mfactor${word:+_$word} -h; done
```
This change would also be needed in #103 on Linux and macOS, and for the singlethreaded builds. Long term, instead of just running the -h option, we should update it to run the actual Mfactor self-test that you mentioned, to prevent regressions.

tdulcet (commented):
Thanks for the fixes. It looks like we should be able to simplify this even further.

MarkRose:
Converged the Windows Mfactor step on the real `test_fac()` self-test (your "long term" note): it now runs `-m 127 -bmax 20` and asserts the "self-tests completed successfully" marker instead of `-h`. I kept the AVX-512 run-guard, though — the mode loop force-builds the AVX512 Mfactor variant even on runners whose CPU cannot execute AVX-512, so running it unconditionally would SIGILL. (Same self-test form is now used across the general Linux/macOS blocks in #82 and the run-guarded blocks in #103, so they stay consistent.)

tdulcet:
> I kept the AVX-512 run-guard, though — the mode loop force-builds the AVX512 Mfactor variant even on runners whose CPU cannot execute AVX-512, so running it unconditionally would SIGILL.

I meant that we should not be running the cross compiled Mfactor binaries at all. This is an existing bug on the `main` branch. Instead, I was suggesting that we only test the native Mfactor binary (compiled with `-march=native` or similar), as the CI already does for Mlucas when running the self-test and `config-fermat.sh` script. Near the end of this CI step, there is a `cd obj` line, which changes into the `obj` directory where the native Mlucas binaries are built. This section is where we should run the Mfactor self-test from the `obj_mfac` directory. Also see https://github.com/primesearch/Mlucas/pull/82#pullrequestreview-4724101318.

**Edit**: The https://github.com/primesearch/Mlucas/pull/93/changes/fd7beac2b34bc93b8764f8deed46573cbe85ab9b commit to #93 shows what I am suggesting (among other changes).

.github/workflows/ci.yml
MarkRose: Thanks for the patience on this one. On the "native build" point: I pulled the actual CI history to check how often GitHub's runners land on AVX-512-capable hardware rather than assuming either way.

Sampling real runs across #83, #103, #93 and #163 (~230 runner-jobs total): **Windows gets an AVX-512-capable runner about 11% of the time (5/45)**, and **Linux about 21% of the time (30/144)**. So this isn't a "never happens, just hardcode the skip" situation — it's a genuine per-run lottery, consistent with the runner pool being a mix of hardware generations.

Given that, I'd like to keep the execution-probe-and-warn approach here rather than dropping to a bare `-h`: it's the only way to know, for a given run, whether AVX-512 can actually execute, and on the ~1-in-9 (Windows) or ~1-in-5 (Linux) runs where it can, the full self-test is exactly what would catch a SIMD-specific regression — which matters here since Mfactor's factoring code doesn't get exercised in any SIMD mode otherwise (only the default/native build gets a real self-test elsewhere in this job).

This also now matches the identical pattern already in place on Linux/macOS (build always, probe execution, warn-not-fail if it can't run, full self-test otherwise) — so it's not extra complexity relative to the rest of the CI, just applying the same approach here too.

Also fixed the redundant `grep`-for-success-marker you flagged on #82 (applies here too) — dropped in favor of Mfactor's own exit code.

tdulcet: From https://github.com/primesearch/Mlucas/pull/103#discussion_r3696024150:
> It seems I am not explaining this very clearly. I fully agree that should run the full Mfactor self-test instead of just the `-h` output, which is why I originally suggested that in #83. What I meant above is that we should not be running the cross compiled builds in these jobs, which is error prone as you noted and is an existing bug on the main branch. Instead, I was suggesting that we test the native Mfactor binary (compiled with `-march=native` or similar), as these jobs already do below when running the Mlucas self-tests and `config-fermat.sh` script. (In the future we may want to test the cross compiled builds, but that should likely be in a separate job to avoid a runtime explosion, and it should test both Mlucas and Mfactor.)
> 
> This commit shows exactly what I am proposing (among other unrelated improvements): https://github.com/primesearch/Mlucas/pull/230/changes/6ae9f8fccc370a43421f581778e1c2788282bd18. If it is easier, I could push the suggested changes directly to this branch.

MarkRose:
Done — dropped the AVX-512 execution probe and the per-mode Mfactor run entirely. The job now runs the native Mfactor self-test once per word size, right after `config-fermat.sh`, which is the same treatment #103 now applies to Linux and macOS. Sorry for the detour; I'd been arguing runner-availability statistics against a point you weren't making.

Also rebased onto current `main` and squashed. The branch was 200 commits behind and conflicted with #222's per-word Mfactor object directories.

## One part of this is more load-bearing than its title suggests

The `factor.h` change — falling back to standard modmul for multiword `p` under `USE_FMADD` instead of erroring — is not Windows-specific, and it turns out to gate #103.

`USE_FMADD` is auto-defined on every FMA-capable target, so the hard `#error P2WORD may not be used together with USE_FMADD!` means 2word/3word/4word Mfactor does not build on avx2 or avx512. That includes the **native** build on any modern x86 runner. Measured on current `main`:

```
native mfac:  <base> OK  1word OK  2word FAIL  3word FAIL  4word FAIL  nword OK
```

With this PR all six build — each emitting the new `#warning` — and run. The `#error` dates to the 2009 import, so those builds have presumably never worked; `|| true` in CI just meant nobody saw it. Since #103 removes that mask, it needs this PR first, and I've recorded **#83 + #114 → #103** in #224.

## Verification

- `mfac {2,3,4}word` × `{avx2, avx512}`: all six build, all six emit the warning, and the binaries run
- native `makemake.sh avx2` build rc=0, `-s tt` rc=0 with 19 residues
- the `llrint` change is a no-op on LP64, where it is the same libm entry point as `lrint`

The MinGW and LLP64 halves can only really be confirmed by the Windows CI, since neither reproduces off real Windows — Linux and Wine always start with the compatible stack phase.

One thing to expect: the Mfactor run needs **#114**. Without it the multiword binaries here abort with `There were errors writing the savefile`; the same binaries return rc=0 on a tree carrying both. That is independent of this PR.

makemake.sh
```diff
@@ -160,6 +155,29 @@ if "$MAKE" --help 2>/dev/null | grep -wq -- '-O'; then
...
+# needing 32/64-byte alignment) into stack frames it never dynamically realigns - but the Win64 ABI
+# only guarantees 16-byte alignment at function entry, and Windows randomizes the initial thread
+# stack phase per run. Result: plain C code (no inline asm involved) faults on ~half of all runs,
+# at a stable instruction but "moving" whenever the code is recompiled - observed as Mfactor's
+# test_fac() SIGSEGVing intermittently in CI (vmovdqa %ymm0,0x170(%rsp) with rsp = 0 mod 32, where
+# gcc assumed 16 mod 32). Linux and Wine always start with the compatible phase, which is why this
+# never reproduced off real Windows. -mstackrealign does NOT help (it realigns only to the 16-byte
+# preferred boundary). Instead, keep compiler-GENERATED vector code to 128 bits so no 32/64-byte-
+# aligned spill slots exist at all; Mlucas's hand-written SIMD asm is unaffected (the assembler
+# needs no -m flags), so AVX2/AVX-512 build modes lose nothing but gcc's autovectorization width:
```
tdulcet: Maybe link to the bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412 and note that this should be revisited after the bug is resolved, as this would likely reduce resulting performance on Windows.

Did you try added something like:
```c
#if defined(__MINGW32__) && defined(__GNUC__) && !defined(__clang__)
__attribute__((target("prefer-vector-width=128")))
#endif
```
to the affected functions (i.e. `test_fac()`)? This could reduce the performance degradation compared to compiling all of Mlucas and Mfactor with that `-mprefer-vector-width=128` option on Windows.

mlucas.cfg
tdulcet, line 1: I do not think you meant to commit this file.

tdulcet (changes requested):
Thanks for the fixes. Just a couple of minor changes are needed.

MarkRose:
Both fixed — and thanks for the bug reference, that's the right one and it turns out to be worth citing for more than provenance.

**`mlucas.cfg`** — no, that was accidental. Removed.

**GCC PR 54412** — now linked in the comment, with a TODO to gate the block on a gcc version once it's fixed.

## The per-function `target` attribute

I tried it. GCC does accept `__attribute__((target("prefer-vector-width=128")))` and it does what it says — clang ignores it with `-Wignored-attributes`, but this block excludes clang anyway:

```
$ gcc -O3 -mavx2 -S pvw.c     # ymm uses, per function
f_plain  3
f_attr   0
```

So the mechanism works. What made me keep it build-wide is the size of the exposure, which I hadn't actually measured before and can now: `x86_64-w64-mingw32-gcc` 16.2 cross-compiles the whole tree on Linux, so this is checkable without a Windows box.

Disassembling every object for aligned 256/512-bit spill stores into `%rsp`/`%rbp` slots, and checking whether the containing function got a realigning prologue:

| build | functions | objects | spill stores | of those, realigned |
|---|---|---|---|---|
| MinGW-w64, avx2 | 10 | 8 | 86 | **0** |
| MinGW-w64, avx512 | 8 | 6 | 24 | **0** |
| MinGW-w64, either, with `-mprefer-vector-width=128` | 0 | 0 | 0 | — |
| **same TUs, same gcc, `--target` Linux** | 20 | — | — | **20** |

That last row is the control, and it is PR 54412 in one line: same compiler binary, same sources, same flags, only the target differs — Linux realigns every frame that needs it, MinGW realigns none of them. It also shows the check is capable of reporting "realigned", which is what makes the zeroes above mean something.

The avx2 list:

```
   61 spills  radix32_ditN_cy_dif1.o    radix32_ditN_cy_dif1
    7 spills  radix4032_ditN_cy_dif1.o  radix4032_ditN_cy_dif1
    4 spills  factor.o                  test_fac
    4 spills  twopmodq256.o             twopmmodq256
    2 spills  Mlucas.o                  read_ppm1_savefiles
    2 spills  Mlucas.o                  extract_known_factors
    2 spills  twopmodq256.o             twopmodq256
    2 spills  util.o                    xmody256
    1 spills  radix240_ditN_cy_dif1.o   radix240_ditN_cy_dif1
    1 spills  radix60_ditN_cy_dif1.o    radix60_ditN_cy_dif1
```

`test_fac()` is 4 of those 86 stores. It is where CI happened to catch the crash, not where the bug lives — six of the ten functions are in Mlucas-only translation units, and `radix32_ditN_cy_dif1` alone accounts for 61. Annotating a list means re-deriving it for every gcc release and every source change, with each miss coming back as a crash on roughly half of runs somewhere new, which is the failure mode that took longest to pin down in the first place.

## The performance cost

Also measurable, and smaller than I expected. `-fft L -iters 1000 -radset 0 -cpu 0:3`, five interleaved repeats per cell on an idle machine, medians:

| FFT | without flag | with flag | ratio |
|---|---|---|---|
| 1024K | 2.884 s | 2.914 s | 1.010 |
| 2048K | 5.724 s | 5.715 s | 0.999 |
| 4096K | 11.619 s | 11.803 s | 1.016 |

Run-to-run spread within a cell is about ±5%, so ≤1.6% is not distinguishable from zero here. That is roughly what the structure predicts: the hot paths are hand-written asm, and the flag only narrows what gcc autovectorizes in the surrounding C.

One caveat on where that was measured — this is Linux, where the no-flag build additionally pays the realignment prologues that the Windows build (buggily) omits, so it very slightly flatters the 128-bit build. At ~1% it does not change the conclusion, but I did not want to present it as a Windows measurement when it is not one.

If you would still rather have the narrow form, I am happy to switch: the attribute works, and the function list is now enumerable rather than guesswork. I would want a CI check that regenerates the list, though, or it rots silently.

One question while we're here: should `mlucas.cfg` go in `.gitignore`? A self-test run from the repo root drops it there, which is how it got committed.

## Rebase and verification

Also rebased onto current `main` (`5ff344bf`) — the branch was 14 commits behind, across #251/#155/#254. Clean rebase, and re-verified afterwards rather than trusting that, since #254 rewrote much of `makemake.sh`:

```
control, unmodified main:   mfac base OK  1word OK  2word FAIL  3word FAIL  4word FAIL  nword OK
this branch, avx2:          all six BUILT
this branch, avx512:        all six BUILT
native build:               BUILT, -s tt rc=0, 19 residues
```

The control still fails post-#254, so the `factor.h` change is still doing work rather than fixing something already fixed.

I also fixed a stale comment in the same file while there — `makemake.sh` still claimed the generated Makefile emits `CFLAGS ?=` and therefore defers to an exported `CFLAGS`. #254 changed it to `CFLAGS =`, so that route silently does nothing. It is not cosmetic: it is exactly what let the sanitizer CI jobs go on building uninstrumented binaries, which is the subject of the other half of #103.

MarkRose:
Small addendum to the above: I force-pushed once more to take `mlucas.cfg` out of the branch history rather than just deleting it in a follow-up commit, so the add never happens at all. The two commits are otherwise unchanged and the resulting tree is byte-identical to the one I described — only the stray file's add/delete pair is gone.

.github/workflows/ci.yml
```diff
@@ -683,11 +690,12 @@ jobs:
             (cd obj${mode:+_$mode}; make clean)
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
-                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
+                # The windows-11-arm runner has no GMP installed (the Install step is x86-only), so the
+                # Mfactor build must also request no_gmp there, exactly as the main build above does:
+                bash -e -o pipefail -- makemake.sh mfac ${{ endsWith(matrix.os, '-arm') && 'no_gmp' || '' }} $word $mode || true
```
tdulcet: We should remove the `|| true` masking while we are here, as you did in #103.

tdulcet (changes requested):
Thanks, just a minor change is needed.

> One question while we're here: should `mlucas.cfg` go in `.gitignore`? A self-test run from the repo root drops it there, which is how it got committed.

Mlucas is typically run from the `obj*` directories, which are already in `.gitignore`, but feel free to include any additional files as well.

tdulcet (approved):
Thanks.

makemake.sh
tdulcet: Also see https://github.com/niXman/mingw-builds-binaries/issues/120. Did you try the suggested `-Wa,-muse-unaligned-vector-move` and does it improve performance compared to `-mprefer-vector-width=128`?

MarkRose: Yes, tried it. It does fix the fault class, but I don't think it's the better trade, and it isn't faster.

**Does it work.** Cross-building the tree with `x86_64-w64-mingw32-gcc` 16.2 / binutils 2.47 and disassembling the linked `Mlucas.exe` (counts are on the linked LTO binary, so not directly comparable to the per-object figures upthread):

| build | 32/64-byte-aligned stores into stack slots | aligned vector moves, total | of those, 256/512-bit |
|---|---|---|---|
| baseline | 187 | 78,246 | 59,179 |
| `-mprefer-vector-width=128` | 0 | 91,295 | 51,649 |
| `-Wa,-muse-unaligned-vector-move` | 0 | **0** | **0** |

Both drive the spill count to zero. The difference is scope. `-mprefer-vector-width=128` changes only what the compiler generates, which is exactly where the bug is. The assembler flag rewrites every aligned vector move in the binary, and most of those are Mlucas's hand-written SIMD asm rather than compiler output: 51,649 of the 59,179 wide aligned moves survive `-mprefer-vector-width=128`, which means the compiler cannot have emitted them.

**Is it faster.** Not measurably. gcc 16.2, avx2, `-fft L -iters 1000 -radset 0 -cpu 0:3`, five interleaved repeats on an idle machine, medians in seconds:

| FFT | baseline | `-mprefer-vector-width=128` | `-Wa,-muse-unaligned-vector-move` |
|---|---|---|---|
| 1024K | 2.587 | 2.608 | 2.564 |
| 2048K | 5.150 | 5.191 | 5.224 |
| 4096K | 11.036 | 10.984 | 10.842 |

The largest gap between configs is 1.8% and its sign flips between lengths, against 1.8-29% run-to-run spread within a single cell, so nothing here is distinguishable from noise. Res64 was identical across all three configs at every length, so the rewrite at least isn't changing results.

Caveat on where that was measured, same as the earlier table: this is Linux, not Windows. Same compiler and the same assembler flag, but I have no Windows box to time on, and the no-flag Linux build additionally pays the realignment prologues the Windows build omits.

**Why I'd keep the current flag** unless you'd rather not: the aligned-move forms are a free alignment check on Mlucas's own asm, and `-Wa,-muse-unaligned-vector-move` removes it from all ~52k of them in order to work around a compiler bug in the other ~7.5k. That's a diagnostic loss rather than a correctness one - unaligned forms accept aligned addresses - but it's a real one, and it buys no measured speed. It also makes the workaround invisible in the generated code: nothing in the disassembly says why those moves are unaligned.

One thing I checked in case we do switch: with `-flto=auto`, a `-Wa,` option present only on the compile line does survive to the LTO link step, since gcc replays the recorded per-TU options. Compile-only and compile-plus-link builds scanned identically, so it works as a plain `ARGS+=` addition to the same block.

Happy to switch if you prefer it - it's a one-line change, and I'd add the niXman link alongside the PR 54412 one either way.

Issue #32, opened 2024-06-13
Assertion failed: Full-residue == 3^nsquares (mod q) check fails!


efvb:
I am using code from main, May 5th, compiled with GCC 11 with tuned Makefile with -O3 -flto=auto -DUSE_ARM_V8_SIMD -mcpu=neoverse-n1
This is an Ampere Arm VM.

I also have the file to restart, if needed.

    Mlucas 21.0.1

    https://www.mersenneforum.org/mayer/README.html

INFO: testing qfloat routines...
System total RAM = 23236, free RAM = 10691
INFO: 10691 MB of free system RAM detected.
CPU Family = ARM Embedded ABI, OS = Linux, 64-bit Version, compiled with Gnu C [or other compatible], Version 11.4.1 20230605 (Red Hat 11.4.1-2.1.0.1).
INFO: Build uses ARMv8 advanced-SIMD instruction set.
INFO: Using prefetch.
INFO: Using inline-macro form of MUL_LOHI64.
INFO: Using FMADD-based 100-bit modmul routines for factoring.
INFO: MLUCAS_PATH is set to ""
INFO: using 53-bit-significand form of floating-double rounding constant for scalar-mode DNINT emulation.
INFO: testing IMUL routines...
INFO: Testing 64-bit 2^p (mod q) functions with 100000 random (p, q odd) pairs...
INFO: System has 4 available processor cores.
INFO: testing FFT radix tables...
Set affinity for the following 4 cores: 0.1.2.3.
User did not set LowMem in mlucas.ini ... allowing all test types.
User did not set CheckInterval in mlucas.ini ... using default.
NTHREADS = 4
Setting ITERS_BETWEEN_CHECKPOINTS = 10000.
 looking for worktodo.txt file...
 worktodo.txt file found...reading next assignment...
 worktodo.txt entry: PRP=164C2E2861CC217CF6033B299322924C,1,2,8546743,-1,99,0,3,5,"209104614239,2707424768671102639169"

INFO: Maximum recommended exponent for FFT length (448 Kdbl) = 8954601; p[ = 8546743]/pmax_rec = 0.9544526886.
Initial DWT-multipliers chain length = [long] in carry step.
 INFO: restart file p8546743 found...reading...
Computing 8540000-squaring residue R (mod known prime q = 209104614239)
        A: R == 9997626983 (mod q)
        B: R == 135567088145 (mod q)
ERROR: at line 5271 of file ../src/Mlucas.c
[...]

PR #84, opened 2026-07-11
Fix operator typo in twopmmodq64 fast path (1ull < p should be 1ull << p)

MarkRose:
### Problem

The `p < 64` fast path in `twopmmodq64()` (`src/twopmodq.c:1201`) is meant to compute `2^p (mod q)` directly, but a typo used the comparison operator `<` instead of the left-shift `<<`:

```c
if(p < 64)
    return (1ull < p) % q;   // (0 or 1) % q  — a boolean, not 2^p
```

`1ull < p` evaluates to `0` or `1`, so the function returned `0 % q` or `1 % q` instead of `2^p mod q` — a wrong result for every `p` in `[1, 63]`.

### Fix

```c
if(p < 64)
    return (1ull << p) % q;
```

The guarding `if(p < 64)` keeps the shift width well-defined (no shift-by-≥64 UB), and the adjacent `p == 64` branch already uses the intended `(1ull << 63) % q` pattern.

### Relationship to #32

Found while investigating #32. **This typo is not the root cause of that report.** Ground-truth computation shows the reporter's savefile residue was genuinely corrupt — produced by the then-broken LTO AArch64 build (since fixed) — so the assert there correctly fired. The reporter should retest on current `main`; this PR just removes a real latent bug uncovered along the way.

### Verification

Compiles clean (`makemake.sh avx2`, plus an independent single-file compile); no other `1ull < <var>` occurrences exist in the file.

### DIFF
--- a/src/twopmodq.c
+++ b/src/twopmodq.c
@@ -1198,7 +1198,7 @@ uint64 twopmmodq64(uint64 p, uint64 q)
 	if(debug) printf("twopmmodq64: computing 2^%" PRIu64 " (mod %" PRIu64 ")\n",p,q);
 	// If p <= 64, directly compute 2^p (mod q):
 	if(p < 64)
-		return (1ull < p) % q;
+		return (1ull << p) % q;
 	else if(p == 64) {
 		x = (1ull << 63) % q;
 		MOD_ADD64(x,x,q,x);


tdulcet (approved):
Thanks for catching this!

PR #85, opened 2026-07-11
Fix P-1 error-code plumbing so multi-error batches can't crash a healthy run (fixes #49)

### LINKED ISSUE #49: Error during F33 P-1 stage 2: `Assertion 'ierr < ERR_MAX' failed: Error code out of range!`
This error was reported by [Ken on the forum](https://www.mersenneforum.org/node/22731?p=1084412#post1084412), when attempting to run P-1 stage 2 on F33 with two different AVX512 systems (both Intel and AMD) using GCC 13.3:
```
worktodo.txt entry: Pminus1=1,2,8589934592,+1,10000000,900000000,96,800000000

INFO: precomputed p-1 stage 1 primes-product file f33.s1_prod found...reading...
INFO: Successfully read precomputed/bit-reversed Stage 1 prime-powers product savefile for this modulus and B1 = 10000000.
Product of Stage 1 prime powers with b1 = 10000000 is 14427494 bits (225430 limbs), vs estimated 14751202. Setting PRP_BASE = 3.
INFO: Maximum recommended exponent for FFT length (524288 Kdbl) = 8937021925; p[ = 8589934592]/pmax_rec = 0.9611629762.
INFO: restart file f33.s1 found...reading...
F33: p-1 stage 1 to b1 = 10000000 already done -- proceeding to stage 2.
Available memory allows up to 40 Stage 2 residue-sized memory buffers.
Using 40 Stage 2 memory buffers, D = 330, M = 1, psmall = 0.
Using B2_start = 800000000, B2 = 900000000, Bigstep = 330, M = 1
Using Bigstep 330, pairing-window multiplicity M = 1: Init M*40 = 40 [base^(A^(b^2)) % n] buffers for Stage 2...
Using 16 threads in carry step
Buffer-init done; clocks = 00:08:37.254, MaxErr = 0.066406250.
Stage 2 q0 = 799999860, k0 = 2424242
Computing Stage 2 loop-multipliers...
Stage 2 loop-multipliers: clocks = 00:04:26.607, MaxErr = 0.066406250.
Error code: 20, Max Error: 16
ERROR: Function returnMlucasErrCode, at line 4903 of file ../src/Mlucas.c
Assertion 'ierr < ERR_MAX' failed: Error code out of range!
```
The error code 20 is possibly two `ERR_CARRY = 10` errors: https://github.com/primesearch/Mlucas/blob/b6dda5253c9edfc34d5fbcdad12704bb61941370/src/Mdata.h#L170-L187

MarkRose:
Fixes #49.

### Problem (#49)

P-1 stage 2 accumulates the per-modmul error code from each `func_mod_square` call into `nerr`. It did so by **adding**:

```c
ierr = func_mod_square(...); nerr += ierr;
```

The `ERR_*` codes are small distinct enum values (`Mdata.h`, `ERR_MAX = ERR_GERBICZ_CHECK = 16`). When two or more errors — of the same or different type — occur within a single bigstep-loop batch, their codes **sum into a meaningless composite** (e.g. two `ERR_CARRY` → 20). That value is then passed to `returnMlucasErrCode()`, which did:

```c
ASSERT(ierr < ERR_MAX, "Error code out of range!");
```

so an otherwise-recoverable run was killed by the diagnostic path. Ernst's own comment at `pm1.c` already predicted this and prescribed the fix.

### Fix

Adopt the prescribed **bitmask** scheme:

- **`pm1.c`** — every `nerr += ierr` becomes `if(ierr) nerr |= 1<<ierr`. Bit *k* is set iff error type *k* occurred at least once in the batch; the `if(ierr)` guard skips the no-error case (bit 0 would otherwise be set on every success).
- **`Mlucas.c` stage-2 return handling** — decode by **bit tests** instead of the old sum/modulo heuristic. The raw `ERR_INTERRUPT` scalar is checked **first** (it is returned unshifted, short-circuiting before any bitmask accumulation), then `1<<ERR_ROUNDOFF`, then a **new** `1<<ERR_CARRY` branch that retries from the most-recent savefile — mirroring the main PRP/LL loop's existing handling of that same residue-corruption error (see `Mlucas.c` main loop) — rather than aborting.
- **`returnMlucasErrCode()`** — return `"unknown/composite error code"` for any out-of-range value instead of asserting. This also corrects a **pre-existing off-by-one**: the old `ierr < ERR_MAX` test rejected the valid top code `ERR_GERBICZ_CHECK` (16); the array has `ERR_MAX` entries indexed `ierr-1`.

### Also fixed (latent, in scope)

The stage-2 "remaining q-singles" loop used `ierr += func_mod_square(...)` where the structurally identical sibling block uses `ierr =`. `ierr` is not reset between passes, so a stale nonzero code from an earlier pass kept accumulating — which under the new mask would set a bit for an error that never actually occurred that pass. Aligned it to `ierr =`.

### Encode/decode consistency

`pm1_stage2()` returns one of: `0` (success); a raw `ERR_INTERRUPT` (returned unshifted, before accumulation); a raw setup-failure sentinel `1` (cannot collide with any bitmask, since bit 0 is never set); or the accumulated bitmask. The decode side matches each shape — exact-match for the scalars, bit-tests for the mask. `modpow()` also uses the mask internally but returns only `(nerr != 0)`, and every caller ignores its return value, so that conversion is behavior-neutral.

### Underlying corruption

The *underlying* `ERR_CARRY` on F33 stage 2 (AVX-512) is a separate residue-corruption issue tracked with the AVX-512 clobber work (#52 / the AVX-512 clobber PR). This change fixes the **error-plumbing crash** that masked it and makes carry errors recoverable via savefile retry.

### Verification

- Diff-reviewed the full encode→decode path; confirmed `err_code[]` has exactly `ERR_MAX` entries (bounds-safe), the `ERR_*` values don't collide across the reordered branches, and `READ_RESTART_FILE` is a valid backward-goto target already used for `ERR_CARRY` in the main loop.
- Builds clean via `makemake.sh avx2`; the `USE_PP1_MULTS` branch (not built by default) also compiles cleanly with `-DUSE_PP1_MULTS`.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -2679,7 +2679,7 @@ PM1_STAGE2:	// Stage 2 invocation is several hundred lines below, but this needs
 					for(i = 0; i < NRADICES; i++) { RADIX_VEC[i] = 0; }		NRADICES = 0;
 					goto SETUP_FFT;
 				} else if(ierr & (1<<ERR_CARRY)) {	// One or more modmuls hit a nonzero-exit-carry (residue-corruption) error
-					sprintf(cbuf,"p-1 stage 2 hit a carry error, retrying from most-recent savefile in case the issue was due to residue-data corruption.\n");
+					snprintf(cbuf,STR_MAX_LEN*2,"p-1 stage 2 hit a carry error, retrying from most-recent savefile in case the issue was due to residue-data corruption.\n");
 					mlucas_fprint(cbuf,1);
 					goto READ_RESTART_FILE;
 				} else if(ierr) {
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -2664,15 +2664,24 @@ PM1_STAGE2:	// Stage 2 invocation is several hundred lines below, but this needs
 				ierr = pm1_stage2(p, pm1_bigstep, pm1_stage2_mem_multiple,
 					a,mult,arrtmp,	// Pointers stage 1 residue in a[], double** scratch storage 4-vector mult[], arrtmp
 					func_mod_square, n, scrnFlag, &tdiff, gcd_str);
-				if(ierr && (ierr % ERR_ROUNDOFF) == 0) {	// Workaround for pthreaded-run issue where a single ROE sometimes shows up as an integer multiple of ERR_ROUNDOFF
+				// pm1_stage2() returns either a single raw ERR_INTERRUPT code (check that first, via exact match) or
+				// a bitmask in which bit ierr_code is set if modmul-error type ierr_code occurred 1-or-more times
+				// among the batch of stage 2 modmuls done since the last such check - so test for specific error
+				// types via bit tests, not equality or (as used pre-bitmask, to work around distinct hits of the
+				// same error type within one batch summing to an integer multiple of it) modulo tests:
+				if(ierr == ERR_INTERRUPT) {
+					// First print the signal-handler-generated message:
+					mlucas_fprint(cbuf,1);
+					exit(1);
+				} else if(ierr & (1<<ERR_ROUNDOFF)) {	// One or more modmuls hit a roundoff error - bump FFT length and restart
 					n = get_nextlarger_fft_length(n);	kblocks = (n >> 10);
 					// Clear out current FFT-radix data, since get_preferred_fft_radix() expects that:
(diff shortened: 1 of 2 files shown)


tdulcet (approved):
Thanks for fixing this. I will leave the PR open for a few days in case anyone else wants to review.

PR #87, opened 2026-07-11
Accept PrimeNet's abbreviated PRP/PRPDC worktodo format (fixes #30)

### LINKED ISSUE #30: Mlucas 21.0.1 - Assertion failed: Expected ',' not found after TF_BITS field in assignment-specifying line!

MarkRose:
Fixes #30.

### Problem (#30)

The worktodo PRP-line parser assumed any comma after `k,b,n,c` introduced a mandatory `how_far_factored,tests_saved` pair. PrimeNet's abbreviated assignment format makes those optional and allows a trailing quoted known-factors list, so:

```
PRP=<aid>,1,2,16582879,-1,"1628121653153521"
```

tripped `Assertion 'Expected ',' not found after TF_BITS field...'` — the parser mistook the known-factors structure for the missing numeric fields. Worse, the factor-extraction path was only reachable when the optional `base,residue_type` fields were present, so abbreviated PRP-CF lines were fundamentally unhandled.

### Grammar now supported (per Prime95 `commonc.c`)

```
{PRP|PRPDC}=[AID,]k,b,n,c[,how_far_factored,tests_saved[,base,residue_type]][,known_factors]
```

All of the below is in the worktodo-assignment parsing block of `ernstMain()`, `src/Mlucas.c`.

- Locate any trailing quoted known-factors list **first**. It is always introduced by a double-quote and the numeric fields never contain one, so `strchr(cptr,'"')` finds it directly; a `num_end` pointer — the comma that introduces the list, itself excluded, or end-of-line when there is no list — then bounds every numeric-field scan, so a comma inside `"factor1,factor2,…"` can never be read as a field separator.
- Require exactly **0, 2, or 4** optional numeric fields (never 1 or 3), with a clear error otherwise. Since that count is validated up front against `num_end`, the field parsing itself is plain `strchr()`/`strtoX`. Defaults: `TF_BITS=0xffffffff`, `tests_saved=0`, `base=3`, `residue_type = 5 if known-factors present else 1`.
- **Extract known factors independent of whether `base,residue_type` were given** — this was the core omission.
- All prior validation retained: `tests_saved ∈ [0,2]`; `residue_type ∈ {1,5}`; known-factors ⇒ type 5; Fermat PRP-CF base must be 3. p-1 splitting is unchanged and preserves the trailing `base,residue_type,known_factors` verbatim while zeroing `tests_saved`.
- Same bounded parsing applied to the **`Pminus1`** branch (which had a narrower version of the same bug: an unbounded scan could walk into a comma inside a multi-entry known-factors list when `B2_start` was omitted). **`Pfactor`** makes its numeric fields mandatory with no factors clause, so it is left unchanged.
- Updated the stale worktodo-format doc comment.

A second commit hardens the same parser: `(unsigned char)` casts on the `isspace`/`isalpha` arguments (passing a plain `char` is UB once the byte has its high bit set, e.g. UTF-8 or Latin-1 in a worktodo line), and the new "0, 2, or 4 fields" diagnostic uses a bounded `snprintf` instead of `sprintf`, matching the convention already used in that function.

> Note for anyone who read an earlier revision of this description: it described two new helper functions, `known_factors_bound()` and `memchr_bounded()`, plus a `Mlucas.h` export. Those were dropped on review in `d85d2d4c` in favour of the `strchr`/`num_end` form above — no new functions, `Mlucas.h` unchanged, `Mlucas.c` delta down from +119/−43 to +85/−43. Parse behaviour and the test results below are unchanged.

### Verification

Built clean (`makemake.sh avx2`) and ran a table-driven parse test (running `Mlucas` against a crafted `worktodo.txt`, reading the assignment):

| Line | Result |
|---|---|
| `…,-1,76,1` (long, 2 fld) | parses; correctly splits off `Pminus1=…` |
| `…,-1` (abbreviated, 0 fld) | parses |
| `…,-1,"…"` (0 fld + factor — the #30 repro) | **parses** (was the assert) |
| `…,-1,76,0,3,5,"…"` (4 fld + factor) | parses |
| `PRPDC=…,76,0,3,1` | parses |
| `PRP=n/a,…,-1,"…"` (`n/a` AID + factor) | parses |
| `…,-1,76,1,3,5,"…"` (4 fld + factor, ts=1) | splits p-1; rewritten PRP keeps `76,0,3,5,"…"` |
| `…,3,5,"f1,f2"` (multi-factor, type 5) | parses; both factors validated |
| `…,-1,76` (1 field) | **rejected**: "must have 0, 2, or 4 … found 1!" |
| `…,76,0,3` (3 fields) | **rejected**: "… found 3!" |
| `…,3,1,"f1,f2"` (type 1 + factors) | **rejected**: "Only PRP-CF tests of type 5 supported!" |

### Prior art

The grammar follows contributor @Artoria2e5's reference (see the untested draft in #55); this is a focused, tested reimplementation confined to the parser (it does not pull in #55's unrelated cross-file changes).

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -393,7 +393,7 @@ uint32	ernstMain
 /*...What a bunch of characters...	*/
 	char *cptr = 0x0, *endp, gcd_str[STR_MAX_LEN], aid[33] = "\0";	// 32-hexit Primenet assignment id needs 33rd char for \0
 	// Used to bound optional-numeric-field parsing in PRP|Pminus1 assignment lines to precede any trailing known-factors list:
-	char *bound = 0x0, *kf_start = 0x0, *tsv_ptr = 0x0;
+	char *num_end = 0x0, *kf_start = 0x0, *tsv_ptr = 0x0;
 /*...initialize logicals and factoring parameters...	*/
 	int restart = FALSE, use_lowmem = 0, check_interval = 0;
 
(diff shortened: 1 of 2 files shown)


tdulcet (commented):
Thanks for working on this and continuing the previous effort from #55.

Do you think it would be possible to simply the required changes. That could make it easier for us to review. Prime95 is able to correctly parse PRP/PRPDC assignment with [just 32 lines of C code](https://github.com/primesearch/Prime95/blob/dd3c130a04481a793ce608803825842abe80f67e/commonc.c#L3075-L3107), but this PR includes over 150 lines of changes, including two new custom functions.

MarkRose:
Thanks - simplified it a fair bit (87937dd). Dropped both custom helper functions (`memchr_bounded` and the backward-scanning `known_factors_bound`); Mlucas.h no longer changes and the Mlucas.c delta went from +119/-43 to +85/-43.

The key simplification: the optional known-factors list is always introduced by a double-quote and the numeric fields never contain one, so a single `strchr(cptr,'"')` locates it directly. A `num_end` pointer (the comma that introduces the list, or end-of-line) then bounds the numeric region. Since the abbreviated PRP format allows exactly 0, 2, or 4 numeric fields, I count them up front and parse with plain `strchr`/`strtoX` gated by that count — no bounded-scan helper needed. The Pminus1 branch uses the same `num_end` idea so a factors-list comma directly after B1,B2 isn't misread as a TF_BITS/B2_start separator.

I verified behavior is unchanged across the format variants — k,b,n,c-only, 0/2/4 numeric fields, abbreviated PRP-CF (`PRP=…,-1,"factor"`), and multi-factor lists whose internal commas must not be counted as field separators.

MarkRose:

PR #88, opened 2026-07-11
Add checked heap-allocation wrappers; convert unchecked alloc sites (fixes #51)

### LINKED ISSUE #51: Check the return value of the `malloc` family of functions
All of the `malloc`, `calloc` and `realloc` function calls should have either an `if` or `assert` statement to check that the return value is not NULL. Otherwise, when an OOM error occurs, this can result in a seg fault when Mlucas/Mfactor attempts to use the pointer. For example, see [this report from Ken on the forum](https://www.mersenneforum.org/node/1082613?p=1084488#post1084488).

MarkRose:
Fixes #51.

### Problem (#51)

Many `malloc`/`calloc`/`realloc` sites don't check the returned pointer, so an allocation failure (e.g. OOM on a memory-constrained host) dereferences NULL and crashes with **SIGSEGV** instead of a clean diagnostic.

### Change

- **`util.c` / `util.h`** — add `malloc_check` / `calloc_check` / `realloc_check`, which on failure print the caller's `file:line` and the requested size through the existing `ABORT()` path (clean SIGABRT), used via `MALLOC` / `CALLOC` / `REALLOC` macros that capture `__FILE__`/`__LINE__`. The checks fire only on a genuine failure of a **nonzero-size** request, so a legitimate `malloc(0)` / `realloc(p,0)` returning NULL is not treated as an error.
- **`align.h`** — route the `ALLOC_*` macros through `realloc_check()` instead of raw `realloc()`. Allocation sizes, the `+256`/`+512` alignment padding, and the pointer casts are **unchanged** — only the called function differs — so this checks **~104** call sites at once. A forward declaration covers files that include `align.h` without `util.h`.
- **~65 remaining genuinely-unchecked bare sites** converted to the wrappers: the thread-array allocations in `mers_mod_square.c` / `fermat_mod_square.c` / `pm1.c`, the per-file `tdat` calloc in the 34 `radix*_ditN_cy_dif1.c` carry routines, and scratch buffers in `factor.c` / `qfloat.c` / `br.c` / `util.c`.

Sites that **already** null-check their result, and allocations inside disabled (`#if 0` / CUDA-only / debug) blocks, are left as-is to keep the change focused and reviewable. **No allocation sizes, casts, or alignment padding were altered.**

### Verification

- Builds clean (`makemake.sh avx2`), no new warnings.
- OOM smoke test — under a tight `ulimit -v`, an allocation now aborts cleanly instead of segfaulting:
  ```
  ERROR: Function realloc_check, at line 1419 of file ../src/Mlucas.c
  Assertion 'ptr != 0x0' failed: realloc to 189481472 bytes failed - out of memory
  ```
  Exit code **134 (SIGABRT)**, not 139 (SIGSEGV). The `file:line` points to the real call site (an `ALLOC_DOUBLE` via the routed macro).

### Notes

- This issue was assigned to @f4ni (Aug 2025); there's been no PR since, so I've picked it up — happy to coordinate if that work is still in progress.
- **Out of scope, but flagged:** at a slightly larger memory cap the process can still SIGSEGV via an unchecked **`pthread_create()`** returning ENOMEM (a separate defect, not a `malloc`/`calloc`/`realloc` site). Worth a follow-up issue.

### DIFF
--- a/src/align.h
+++ b/src/align.h
@@ -35,42 +35,48 @@ pointer is null we exit immediately, thus the resulting memory leak is never an 
 In the Align macros we cast pointers to longto accommodate architectures which use 64-bit address arithmetic.
 Note that rather than simply assuming sizeof(void *) <= sizeof(long), we check this at program invocation, in
 util.c::check_nbits_in_types()>
-*/
 
-#define ALLOC_INT(_p,_n)	(int           *)realloc(_p,(_n)*sizeof(int           )+256)
+The ALLOC_* macros route through realloc_check() (util.c) rather than calling realloc() directly, so a
+failed allocation prints the file:line of the call site and aborts cleanly instead of handing back NULL
+for the caller to blindly dereference. realloc_check() is declared in util.h, which every file using
+these macros already includes (directly or via Mdata.h/Mlucas.h).
+*/
+void *realloc_check(void*ptr, size_t nbytes, const char*file, int line);
+
+#define ALLOC_INT(_p,_n)	(int           *)realloc_check(_p,(_n)*sizeof(int           )+256, __FILE__, __LINE__)
 #define ALIGN_INT(_p)		(int           *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_UINT(_p,_n)	(uint32        *)realloc(_p,(_n)*sizeof(uint32        )+256)
+#define ALLOC_UINT(_p,_n)	(uint32        *)realloc_check(_p,(_n)*sizeof(uint32        )+256, __FILE__, __LINE__)
 #define ALIGN_UINT(_p)		(uint32        *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_INT64(_p,_n)	(int64         *)realloc(_p,(_n)*sizeof(int64         )+256)
+#define ALLOC_INT64(_p,_n)	(int64         *)realloc_check(_p,(_n)*sizeof(int64         )+256, __FILE__, __LINE__)
 #define ALIGN_INT64(_p)		(int64         *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_UINT64(_p,_n)	(uint64        *)realloc(_p,(_n)*sizeof(uint64        )+256)
+#define ALLOC_UINT64(_p,_n)	(uint64        *)realloc_check(_p,(_n)*sizeof(uint64        )+256, __FILE__, __LINE__)
 #define ALIGN_UINT64(_p)	(uint64        *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_UINT128(_p,_n)(uint128       *)realloc(_p,(_n+_n)*sizeof(uint64     )+256)
+#define ALLOC_UINT128(_p,_n)(uint128       *)realloc_check(_p,(_n+_n)*sizeof(uint64     )+256, __FILE__, __LINE__)
 #define ALIGN_UINT128(_p)	(uint128       *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_FLOAT(_p,_n)	(float         *)realloc(_p,(_n)*sizeof(float         )+256)
+#define ALLOC_FLOAT(_p,_n)	(float         *)realloc_check(_p,(_n)*sizeof(float         )+256, __FILE__, __LINE__)
 #define ALIGN_FLOAT(_p)		(float         *)(((intptr_t)(_p) | 63)+1)
 
-#define ALLOC_DOUBLE(_p,_n)	(double        *)realloc(_p,(_n)*sizeof(double        )+512)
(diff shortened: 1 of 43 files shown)


MarkRose:
The unchecked-`pthread_create` follow-up flagged above is now fixed in #89 (the crash was actually in `threadpool_init`'s partial-failure cleanup joining never-created thread handles). Verified A/B: same OOM scenario went from SIGSEGV to a clean abort.

src/util.h
```diff
@@ -235,6 +235,16 @@ void	WARN	(long line, char*file, char*warn_string, char*warn_file, int copy2stde
...
 
+/* Checked heap-allocation wrappers (defined in util.c): on failure they print the file:line
+of the call site plus the requested size, then abort, rather than returning NULL for the
+caller to dereference. Use via the MALLOC/CALLOC/REALLOC macros, which capture the call site: */
+void	*malloc_check (size_t nbytes, const char*file, int line);
+void	*calloc_check (size_t nmemb, size_t size, const char*file, int line);
+void	*realloc_check(void*ptr, size_t nbytes, const char*file, int line);
+#define MALLOC(n)		malloc_check ((n), __FILE__, __LINE__)
+#define CALLOC(nm,sz)	calloc_check ((nm),(sz), __FILE__, __LINE__)
+#define REALLOC(p,n)	realloc_check((p),(n), __FILE__, __LINE__)
```
tdulcet: Maybe combine these functions with the macros, to eliminate the extra function call overhead. We could also add `__func__` to each macro, as `ASSERT` does above, so the error message would show the actual function where the allocation failed in.

src/align.h
```diff
@@ -35,42 +35,48 @@ pointer is null we exit immediately, thus the resulting memory leak is never an
...
+
+The ALLOC_* macros route through realloc_check() (util.c) rather than calling realloc() directly, so a
+failed allocation prints the file:line of the call site and aborts cleanly instead of handing back NULL
+for the caller to blindly dereference. realloc_check() is declared in util.h, which every file using
+these macros already includes (directly or via Mdata.h/Mlucas.h).
 */
+void *realloc_check(void*ptr, size_t nbytes, const char*file, int line);
 
-#define ALLOC_INT(_p,_n)	(int           *)realloc(_p,(_n)*sizeof(int           )+256)
+#define ALLOC_INT(_p,_n)	(int           *)realloc_check(_p,(_n)*sizeof(int           )+256, __FILE__, __LINE__)
```
tdulcet: Maybe use the macros here as well, to avoid repeating `__FILE__, __LINE__` in every definition.

tdulcet (changes requested):
Thanks for your work on this. I just have a couple of suggestions for improvement.

src/util.h
MarkRose: Done (fc84567). MALLOC/CALLOC/REALLOC are now statement-expression macros that do the allocation inline on the success path — no wrapper-function call — and only call a single noreturn `alloc_fail()` on failure. They now also pass `__func__`, so an OOM abort names the function the allocation was requested in, like `ASSERT`. The three `*_check` wrappers collapsed into that one `alloc_fail`.

src/align.h
MarkRose: Done. The ALLOC_* macros now expand to `REALLOC(...)` instead of open-coding `realloc_check(..., __FILE__, __LINE__)` in each definition, so the call-site capture lives in one place and `__FILE__`/`__LINE__`/`__func__` still resolve to the actual ALLOC_* use site.

tdulcet (dismissed):
Thanks, this looks great. I will leave the PR open for a few days in case anyone else wants to review.

MarkRose:
Pushed a fix: this branch did not compile.

```
pm1.c:1174:28: error: 'attr' undeclared (first use in this function)
```

The first commit added

```c
	// Initialize and set thread detached attribute:
	pthread_attr_init(&attr);
	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_JOINABLE);
```

to `pm1_stage2()`, but nothing in `pm1.c` declares `attr`. It fails on this branch's head, on its merge-base, and on `main` — so it was not a merge interaction, just broken.

I removed the block rather than adding a declaration. `pm1.c` has **no `pthread_create` call at all**, on this branch or on `main` — Stage 2 dispatches through the threadpool (`tpool`), so a `pthread_attr_t` here would be initialised and then never read. The block is also out of scope for a checked-heap-allocation change; it looks like it came in from another branch during a rebase. The CALLOC conversions that are this PR's actual subject are untouched.

Verified after the fix: `makemake.sh nosimd` rc=0 on the branch head, and merged onto current `main` both `nosimd` and `avx2` build rc=0 with no conflict.

Kept as a separate commit on top rather than rewriting history, since this PR already has review activity.

Found by a pairwise merge-and-build checker over the open PR set — it compiles each PR against `main` and against every other PR sharing a file with it, which is how a PR that simply does not build got noticed.

tdulcet (approved):
Thanks.

PR #90, opened 2026-07-11
util.c hardening: ctype-on-signed-char UB, unbounded path construction

MarkRose:
> **Revised.** This PR originally also converted several `float`↔`uint64` pointer-cast reinterpretations to `memcpy`. That work landed independently via **#124** (merged 2026-07-24), which introduced the `f64_to_u64()`/`u64_to_f64()` union helpers, and a later rebase dropped this branch's copy. The section describing it, and its verification bullet, have been removed — the diff no longer contains any such change.
>
> A follow-up commit also removes the two declarations that work left stranded. One of them, `u_dhi`/`u_dlo` in `cmp_fma_lohi_vs_exact()`, was a live `-Wunused-variable` in every avx2/avx512 build; `makemake.sh avx2` now matches a clean `main` build exactly.

A focused hardening pass on `src/util.c`, addressing three classes of latent defect found in an adversarial review. No behavioral change — each fix is semantics-preserving.

### 1. ctype on possibly-signed `char` (UB)

`isdigit(c)` / `isspace(c)` on a plain `char` in the five `convert_base10_char_{double,uint64,uint128,uint192,uint256}` parsers is undefined when `char` is signed and the byte's high bit is set. Added `(unsigned char)` casts at all ten call sites.

### 2. Unbounded path construction

`mlucas_fopen()` built its path with `strcpy()`+`strcat()`, and `mkdir_p()` with `strcpy()`, into fixed-size buffers with no length guard. Switched both to a single bounded `snprintf(dst, sizeof(dst), …)` producing the identical string.

### Verification

- Builds clean (`makemake.sh avx2`).
- `-s tt` self-test passes with **unchanged residues**, A/B-compared against an unpatched build (only non-deterministic timing/RAM lines differed; all `Res64`/radix-set lines identical) — confirming the hot float-estimate paths (`finvest`/`fisqrtest`) are unaffected.

_Split by file from a larger review backlog; a companion PR hardens `Mlucas.c`. The pervasive `sprintf(cbuf)`→`snprintf` sweep is deliberately deferred to avoid conflicting with in-flight PRs._

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -3100,6 +3100,7 @@ I = 981 Needed extra sub: a = 916753724; p = 11581569; pinv = 370 [a/p = 79.1562
 			  #if 0
 				#error to-do!
 				double r1,r2, lo,hi;
+				uint64 i64r1,i64r2;
 				r1 = rng_isaac_rand_double_norm_pm1() * pow2_dmult;	// in [-2^50, +2^50]
 				r2 = rng_isaac_rand_double_norm_pm1() * pow2_dmult;	// in [-2^50, +2^50]
 				mul50x50_debug(r1,r2, &lo,&hi);
@@ -3227,7 +3228,7 @@ void	mul50x50_debug(double a, double b, double *lo, double *hi)
 	int cmp_fma_lohi_vs_exact(double dx, double dy, double dhi, double dlo, uint64 ix, uint64 iy, uint64 ihi, uint64 ilo)
 	{
 		int retval;
-		uint64 i64, e1,e0, m1,m0;
+		uint64 i64, e1,e0, m1,m0, u_dhi,u_dlo;
 		uint32 s1,s0;
 		const uint64 two52 = 0x0010000000000000ull, mmask = 0x000FFFFFFFFFFFFFull;
 		uint128 exact;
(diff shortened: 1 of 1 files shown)


tdulcet (commented):
Thanks for working on this.

I believe the strict-aliasing warnings were already fixed in #34. If these specific instances were fixed, you should add @ldesnogu as a coauthor of your commit.

MarkRose:
I dug into the overlap with #34. This PR changes only `src/util.c`, and #34 does not touch `util.c` at all (it's still open, and `util.c` isn't in its file list) — so these strict-aliasing fixes don't duplicate any of ldesnogu's work, and there's no coauthor attribution to add here. I confirmed the specific type-puns this PR fixes still exist on `main` (e.g. `itmp = *(uint64 *)&x;` at util.c:5714, and the `*(uint64*)&dhi`/`*(double*)&itmp` casts in `finvest`/`fisqrtest`/`mul50x50_debug`), i.e. #34 did not address them.

Since the two PRs modify disjoint files there won't be a merge conflict either. If you'd still prefer these util.c aliasing fixes to live in #34 for consistency, I'm happy to close this and let ldesnogu fold them in — just let me know.

MarkRose:
Good call — confirmed the overlap: #34 fixes the same `util.c` strict-aliasing instances this PR does (`mul50x50_debug`, `finvest`, `fisqrtest`; ldesnogu used an `f64_to_u64()` helper where this PR uses `memcpy`). Added `Co-Authored-By: Laurent Desnogues` to the commit crediting @ldesnogu, and force-pushed.

tdulcet (approved):
Thanks.

PR #91, opened 2026-07-11
Mlucas.c: NUL-terminate command-line argument copies

MarkRose:
A focused hardening pass on `src/Mlucas.c`, companion to the `util.c` hardening PR. No behavioral change.

### Unterminated command-line argument copies (OOB read)

Command-line parsing used, in 20 places (across `ernstMain` and `parse_cmd_args_get_shift_value`):

```c
strncpy(stFlag, argv[nargs++], sizeof(stFlag)-1);
```

`stFlag` is an uninitialized `char[STR_MAX_LEN]` stack buffer, and `strncpy` does **not** NUL-terminate when the source is `>= STR_MAX_LEN-1` chars — so a long argv token left `stFlag` unterminated and the subsequent `STREQ`/`strcmp` read out of bounds. Switched each to:

```c
snprintf(stFlag, sizeof(stFlag), "%s", argv[nargs++]);
```

which copies the same bytes for normal-length arguments and always terminates. (`snprintf` evaluates `argv[nargs++]` exactly once, preserving the side effect.)

### Removed: the Suyama residue-variable initialisers

An earlier revision of this PR also initialised `itmp64`/`Res35m1`/`Res36m1` at their declaration in `Suyama_CF_PRP()`. Per review those double-initialised variables the existing code already covers, and checking showed they were not fixing anything, so they have been dropped rather than removing the existing `else` arm:

- all three arms of the `MODULUS_TYPE` chain already assign `Res35m1`/`Res36m1` before the diagnostic print — Fermat via `convert_res_FP_bytewise()`, Mersenne via `res_SH()`, `else` via the `UINT64_MAX` sentinel added in `cecc6260`;
- `itmp64` is written by `convert_res_FP_bytewise()` before its first read, with only aborting conditionals in between;
- `gcc -O2 -Wall -Wextra` on unmodified `main` reports no uninitialised warnings anywhere in `Mlucas.c`.

Keeping the `else` also preserves the `UINT64_MAX` sentinel, which is visibly invalid where `0` would read as a plausible residue.

### Verification

- Builds clean (`makemake.sh avx2`); `gcc -O2 -Wall -Wextra` adds no new warnings.
- Command-line parsing exercised via `-s tt -cpu 0:3` (radix-sets pass) and `-fft 192 -radset 0 -iters 100` (100 iterations of M3888509, correct `Res64`) — flags parse identically.

_Split by file from a larger review backlog (companion: the `util.c` hardening PR). The pervasive `sprintf(cbuf)`→`snprintf` sweep is deferred to avoid conflicting with in-flight Mlucas.c PRs._

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -3321,8 +3321,7 @@ uint32 Suyama_CF_PRP(uint64 p, uint64 *Res64, uint32 nfac, double a[], double b[
 	uint64 *ai = (uint64 *)a, *bi = (uint64 *)b;	// Handy 'precast' pointers
 	uint32 i,j,k, isprime, ierr = 0, ihi, fbits,lenf;
 	uint32 kblocks = n>>10, npad = n + ( (n >> DAT_BITS) << PAD_BITS );	// npad = length of padded data array
-	uint64 itmp64 = 0, Res35m1 = 0, Res36m1 = 0;	// Res64 from original PRP passed in via pointer; these are locally-def'd
-	// (init to 0 so the diagnostic prints below are well-defined even if the modulus type matches neither branch)
+	uint64 itmp64, Res35m1, Res36m1;	// Res64 from original PRP passed in via pointer; these are locally-def'd
 	cbuf[0] = '\0';
 	snprintf(cbuf,sizeof(cbuf),"Suyama-PRP on cofactors of %s: using FFT length %uK = %u 8-byte floats.\n",PSTRING,kblocks,n);//	strcat(cbuf,g_cstr);
 	// sprintf(g_cstr, " this gives an average %20.15f bits per digit\n",1.0*p/n);	strcat(cbuf,g_cstr);
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -3321,7 +3321,8 @@ uint32 Suyama_CF_PRP(uint64 p, uint64 *Res64, uint32 nfac, double a[], double b[
 	uint64 *ai = (uint64 *)a, *bi = (uint64 *)b;	// Handy 'precast' pointers
 	uint32 i,j,k, isprime, ierr = 0, ihi, fbits,lenf;
 	uint32 kblocks = n>>10, npad = n + ( (n >> DAT_BITS) << PAD_BITS );	// npad = length of padded data array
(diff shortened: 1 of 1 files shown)


tdulcet (dismissed):
Thanks for these improvements. I will leave the PR open for a few days in case anyone else wants to review.

src/Mlucas.c
tdulcet, line 3346: Maybe remove this now, as there is no need to double initialize these variables.

MarkRose:
Reply for PR #91 (hardening-mlucas-c). Posted as a PR comment.

---

Good catch — and checking it turned out to settle the question in the other direction, so I've removed the initialisers rather than the `else` arm.

They weren't fixing anything. All three arms of the `MODULUS_TYPE` chain already assign `Res35m1` and `Res36m1` before the diagnostic print at the end of that block — Fermat via `convert_res_FP_bytewise()`, Mersenne via `res_SH()`, and the `else` via the `UINT64_MAX` sentinel. `itmp64` is likewise written by `convert_res_FP_bytewise()` before its first read, with only aborting conditionals in between. And `gcc -O2 -Wall -Wextra` on unmodified `main` reports no uninitialised warnings anywhere in `Mlucas.c`.

The comment I'd attached to them claimed they made the prints "well-defined even if the modulus type matches neither branch". That was simply wrong: the `else` arm *is* that case, and it was already handled. Apologies for the noise.

One reason to prefer keeping the `else` over keeping the `= 0`: it preserves the `UINT64_MAX` sentinel. Zero reads as a plausible residue where `UINT64_MAX` is visibly invalid — and radix1008 returning an all-zero residue with a *perfect* roundoff error (#204) is a decent argument for not letting bogus values look reasonable.

So this PR is now just the command-line hardening. That half is a real fix: `strncpy(stFlag, argv[n], sizeof(stFlag)-1)` leaves `stFlag` unterminated for any argument of 1024 bytes or more, and `stFlag` is an uninitialised stack buffer, so a long argument reads past the copy.

Also — sorry, your approval on this got dismissed because I rebased the branch on 29 July, a fortnight after you'd approved it. Entirely my fault and I've stopped doing that. Re-requesting now.

tdulcet (approved):
Thanks.

PR #86, opened 2026-07-11
Accept the valid MSW==0 carryout boundary in convert_res_bytewise_FP (fixes #19)

MarkRose:
Fixes #19.

### Problem (#19)

`convert_res_bytewise_FP()` rejected a **legal** balanced-representation carry-propagation state with an assert, killing an otherwise-healthy run.

After the per-digit normalize step:

```c
itmp += cy;                       // add carry-in from previous digit
if(itmp > 0) {
    cy   = itmp >> bits[ii];
    itmp -= (cy << bits[ii]);
    if(itmp > (base[ii]>>1)) { itmp -= base[ii]; cy += 1; }
} else cy = 0;
a[j1] = (double)itmp;
```

an all-ones top digit (`base[ii]-1`) plus an incoming carry of 1 gives `itmp = base[ii]` → `cy = 1`, `itmp` reduced to exactly **0**, no rebalance. That is an ordinary carry ripple — not corruption — with probability ≈ `2^-bits_per_word` per conversion, which is why only oversized-FFT runs ever tripped it. But the guard

```c
if(cy && (a[j1] >= 0 || cy != +1)) ASSERT(0, ...);
```

treated `MSW == 0 && cy == 1` as illegal and aborted.

Compounding it, the diagnostic printed `a[j]` — where `j` has run past the last data word (an unused pad element) rather than indexing the top digit that was actually tested. That produced the spurious "most sig. word = -21.0000" and "= 16448.0000" readings reported in #19, which misdirected the entire diagnosis thread.

### Fix

- Relax the comparison `a[j1] >= 0` → `a[j1] > 0`, admitting exactly the legal `cy == +1 && a[j1] == 0` state while keeping `cy == +1 && a[j1] > 0` (top digit failed to reduce → `cy` genuinely bogus) illegal. `cy == +1 && a[j1] < 0` was already legal.
- Print `a[j1]` (the padded index of the top digit actually tested) instead of `a[j]`.
- Make that diagnostic bounded: `sprintf(cbuf, ...)` → `snprintf(cbuf, STR_MAX_LEN*2, ...)`. `cbuf` is `char cbuf[STR_MAX_LEN*3]`, so the bound is conservative but safe.

The fold-in that follows (`a[0] += cy` Mersenne / `a[0] -= cy` Fermat) uses **only** `cy`, so it handles `a[j1] == 0` identically to the already-legal `a[j1] < 0` case — the relaxation is provably safe and no wider than the one new state.

### Sibling-path audit

[...]

PR #78, opened 2026-07-10
Fix executed shift-exponent and null-pointer-offset UB (fixes #28, #9)

MarkRose:
## Problem (#28, #9)

Two categories of executed undefined behavior:
- **Shift-exponent UB (#28)**, three sites:
  - `shift_word()` (`Mlucas.c`) evaluated `itmp64 << (64-mod64)` when `mod64 == 0` (i.e. `<< 64`) before masking the result to 0.
  - `RADIX_256_DIT` and `SSE2_RADIX256_DIT` (`dft_macro.c`) each compute `i_idx >> nshift` with `nshift = i+i+2` as setup for the *next* loop pass, so on the last pass (`i == 15`) they shift a `uint32` by 32 — the result is unused, but the shift is still evaluated.
- **Null-pointer-offset UB (#9):** in `SSE2_RADIX_64_DIT` (`dft_macro.c`), the seven `rhi`-pointers `*r10 = r00+0x10, … *r70 = r00+0x70` were formed in the declaration list, but `r00` is NULL on the init-mode (`thr_id == -1`) calls — forming a pointer offset from a null base is UB. This is the file's only live instance of the construct; the sibling in `SSE2_RADIX_64_DIF` sits in a `#else` block whose third line is an `#error`, so it never compiles.

## Fix

- Compute the `mod64==0` result directly via a conditional (no out-of-range shift).
- Guard the two last-pass setups with `i < 15`.
- Move `SSE2_RADIX_64_DIT`'s `r10..r70` assignments past the init-mode early-return. They are used later in the function, so they are moved rather than deleted.

## Verification (AVX2)

- Radix-256 residue **byte-identical to baseline** (`95AFD7A5269F14F6`).
- UBSan (`-fsanitize=undefined`): **0 runtime errors** on the radix-256 path and full `-s tiny`, no residue mismatches.

## Verification (ARM hardware, ASan + UBSan, full `-s medium`)

Before/after A/B on a Raspberry Pi 4B (aarch64), built with the CI's `-fsanitize=address,undefined -Og` flags and run over the full medium self-test (FFT lengths 2M–15M, `-cpu 0:2`). "before" = base `main` without this PR; "after" = this PR applied. Both: EXIT 0, 24/24 FFT lengths, **no ASan memory errors**.

| UBSan errors | before | after |
|---|---|---|
| shift-exponent (#28) — the three sites above: `shift_word()`, `RADIX_256_DIT`, `SSE2_RADIX256_DIT` | 3 | **0** |
[...]

PR #93, opened 2026-07-12
[DRAFT] Integration branch: all open recent fix PRs combined

MarkRose:
> **DRAFT — integration branch, not intended for merge as-is.** This aggregates every currently-open fix PR into one branch so the full set can be built and CI-tested together. Please review and merge the individual PRs listed below; this branch just demonstrates they compose cleanly and gives one combined CI run across the whole matrix.

## What this is

`all-fixes-integration` is `primesearch/Mlucas@1f60681` (current `main`) with all 26 of my open fix PRs (#67–#92) plus community PR #45 merged in, in that order. Each was merged as a normal merge commit, so provenance is preserved and any individual PR can still be reviewed on its own.

It builds and self-tests clean:
- **x86-64 / AVX2** (glibc 2.43, gcc 16): `makemake.sh` → 0 errors, `-s tt` passes.
- **aarch64 / ASIMD** (Raspberry Pi 4B, Debian 12, gcc 12.2): builds clean. *(Notably, current `main` does **not** build here — it fails to link with `undefined reference to CPU_ZERO/CPU_SET` — until #67 adds `-D_GNU_SOURCE`. So this branch also serves as a real-world aarch64 build fix demonstration.)*

## Conflict resolutions

Only two pairs of PRs touched overlapping lines; both resolved as unions with no behavior lost:

1. **#45 × #79** — 5 radix carry files (`radix{44,52,176,208,352}_ditN_cy_dif1.c`): #45 adds the `ATTR_NO_ASAN` attribute to `cyNN_process_chunk`, #79 changes its signature to the `run_func` type `void(void*, int)`. Resolved by keeping both: `void ATTR_NO_ASAN cyNN_process_chunk(void*targ, int thread_num)`.
2. **#70 × #88** — `pm1.c` Stage-2 thread setup: #70 adds the free-before-realloc block and removes the dead `pthread_attr_t` init; #88 converts the `calloc`s to the checked `CALLOC` wrapper. Resolved by keeping #70's free block + dropped attr and adopting #88's `CALLOC`.

## Included PRs

**Build / portability / CI**
- #67 — probe toolchain capabilities instead of hardcoding version cutoffs (#60)
- #71 — fix `'asm operand has impossible constraints'` on ancient compilers (part of #60)
- #73 — CI: skip AVX-512 build mode where the assembler can't do it
- #75 — CI: drop the Linux-old `continue-on-error` mask
- #81 — fix Mfactor build break: disable never-implemented USE_FMADD 100-bit modpow
- #82 — make inline-asm labels asm-local to fix LTO symbol collisions in Mfactor
- #83 — Windows build portability: probe-gate AVX-512, fix ARM detection, Mfactor multiword+FMADD

**Correctness / crashes**
- #45 — fix occasional thread-local memcheck fail (#8) *(community, @KarlLudwig3485)*
- #68 — declare AVX-512 opmask/high-vector clobbers; harden carry-loop index arithmetic (#65/#52/#16)
- #69 — skip too-small FFT lengths on the AVX/AVX-512 `wrapper_square` path (#7)
- #77 — relax the Gerbicz-check carry-out assert for p ≡ 63 (mod 64) (#33)
- #80 — fix corrupted carryout in `mi64_mul_scalar_add_vec2` asm (crashes final PRP determination)
- #84 — fix operator typo in `twopmodq64` fast path (`1ull < p` → `1ull << p`)
- #86 — accept the valid MSW==0 carryout boundary in `convert_res_bytewise_FP` (#19)

**Robustness / error handling / leaks**
- #70 — free threadpools on re-init instead of leaking them (#44)
- #79 — align threadpool worker signatures with the `run_func` type (#29)
- #85 — fix P-1 error-code plumbing so multi-error batches can't crash a healthy run (#49)
- #88 — add checked heap-allocation wrappers; convert unchecked alloc sites (#51)
- #89 — fix `threadpool_init` partial-failure cleanup joining never-created threads (OOM SIGSEGV)

**Feature / parsing / resume**
- #72 — persist/recover P-1 stage-1 B1 across restarts (#31)
- #74 — gate the per-checkpoint FFT-length reversion (#21)
- #76 — bound the preferred-FFT cfg scan to ≤ 2× default FFT length (#36)
- #87 — accept PrimeNet's abbreviated PRP/PRPDC worktodo format (#30)

**Undefined-behavior / hardening**
- #78 — fix executed shift-exponent and null-pointer-offset UB (#28, #9)
- #90 — `util.c`: strict-aliasing UB, ctype-on-signed-char UB, unbounded-path build
- #91 — `Mlucas.c`: NUL-terminate arg copies; initialize Suyama residue vars
- #92 — bound `cbuf` writes that format variable-length input with `snprintf`

*(Issue numbers are listed for reference only — this branch intentionally uses no `Fixes #N` keywords so nothing auto-closes; the individual PRs carry those.)*


MarkRose:
Closing as superseded by #163, which says so on its own side but was never mirrored here — so both drafts have been sitting in the list looking equally current.

#163 is the same idea rebuilt fresh against current `main`. This branch had grown stale: its "Included PRs" list still names PRs that have since merged or been closed (#81 among them, now superseded by #193), so anyone using it as an index of open work would be misled.

Nothing is lost — the integration branch is a build-and-test vehicle, not a source of unique changes, and every fix it aggregated is tracked on its own PR. #224 now carries the merge ordering and the per-capability index that this PR's description was informally serving.

PR #103, opened 2026-07-16
CI: guard AVX-512 execution on runners that can't run it; build ASan NOSIMD

MarkRose:
Unmasks the CI failures the `continue-on-error` blanket was hiding, and makes the jobs run enough to be worth unmasking. Grew out of the all-fixes integration branch (#93), where removing the masking made these visible.

## What it does

**Removes the masking.** Three `continue-on-error` expressions go away (Linux `ubuntu-24.04`+clang / `ubuntu-26.04`, Container non-ARM, ASan-Linux), along with six `|| true` suffixes on `makemake.sh mfac` calls. A failing build or self-test is now a failing job.

**Runs more of the self-test.** Where a job ran only `-s m`, it now runs `-s tt`, `-s t`, `-s s`, `-s m`, plus the PRP self-tests `-prp -s {tt,t,s}`. `config-fermat.sh` gains `-e -o pipefail`, which it needed to report its own failures.

**Self-tests only the native Mfactor build.** Per @tdulcet on this PR and #83: the explicit-ISA builds (`sse2`, `avx`, `avx2`, `avx512`, …) exist for *compile* coverage, and a runner may not be able to execute what they target — that is what produced the AVX-512 SIGILL (exit 132) on `ubuntu-26.04` and `macos-15-intel`. Rather than probing for AVX-512 execution support at runtime, the four Linux/macOS jobs now drop the per-mode `Mfactor -h` loop entirely and run `Mfactor -m 127 -bmax 20` once per word size against the native (`-march=native`) build. All the explicit-ISA binaries are still built.

On ARM the run is replaced by `-h`, because trial factoring is unimplemented there — `twopmodq80.c:3325` asserts `"No TF support on ARMv8!"` — so `test_fac` cannot run at all.

**ASan/UBSan builds switch to NOSIMD.** The macOS sanitizer build fails to assemble `pm1.o` under `-Og -fsanitize=address` with `invalid CFI advance_loc expression` (ASan plus hand-tuned SIMD inline asm). Per the discussion on #71, testing the assembly under the sanitizers adds little and NOSIMD exercises the same C-level logic they care about. Also re-enables the TSan-macOS job on the tiny self-test.

**Container matrix:** adds `ubuntu:14.04` back, excluding only the one combination that genuinely cannot work — `ubuntu-24.04-arm` + clang, whose 3.4-era clang predates AArch64 NEON `v`-register asm syntax and cannot assemble `dft_macro.c` (`unknown register name 'v0'`). Its gcc builds fine on ARM, and both compilers build fine on x86-64.

**Static-analysis jobs** get an explicit `-std=gnu99` / `--std=c99` so gcc-analyzer, clang-tidy and cppcheck all parse the tree the same way the real build does.

## Merge order — three prerequisites

The native `Mfactor -m 127 -bmax 20` run added here does not work on `main` today. Each of these is an independent, already-open fix; all three must land first or this PR turns CI red:

| PR | without it |
|---|---|
| **#193** | `makemake.sh mfac` does not build on the AVX2 runners at all |
| **#82** | the LTO link fails on duplicate asm labels in `twopmodq96.c` (three collisions) |
| **#114** | the run itself aborts: `Assertion '!i' failed: There were errors writing the savefile` |

#114 is the least obvious of the three. `write_savefile` interleaves `fgets` reads and `fprintf` writes on an `"r+"` stream with no intervening `fseek`/`fflush` (C11 §7.21.5.3p7 undefined behaviour), which garbles the checkpoint file — `kmax` overwritten by `know`, `passmin` by `min`, a duplicated `passnow` — and the next pass then fails to parse it. It already fires today: run `30322834573` on #82 hit it, on unmodified `main`.

Verified locally on `main` + #193 + #82 + #114: build rc=0, run rc=0, well-formed `t127`, and the run completes with `M(127) has 0 factors in range k = [0, 16336320], passes 0-15`.

Recorded in #224.

## Scope note

Leaves the Windows job's AVX-512 run-line to #83, which applies the same reasoning there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -760,15 +760,7 @@ jobs:
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode
-                (
-                    cd obj_mfac${word:+_$word}${mode:+_$mode}
-                    make clean
-                    if [[ -x Mfactor${word:+_$word} ]]; then
-                        # Run Mfactor's test_fac() self-test (see #82) instead of -h; M(127) has no
-                        # factors <= 2^20 so the pass is quick. Mfactor already exits nonzero on failure.
-                        ./Mfactor${word:+_$word} -m 127 -bmax 20
-                    fi
-                )
+                (cd obj_mfac${word:+_$word}${mode:+_$mode}; make clean)
             done
         done
         cd obj
@@ -778,6 +770,15 @@ jobs:
             cd ../obj_NOSIMD
         fi
         bash -e -o pipefail -- ../config-fermat.sh -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))"
+        # Mfactor self-test on one build only. The explicit-ISA builds above are for compile
+        # coverage; running them means executing a binary this runner may not support, which is
+        # how the FreeBSD jobs came to SIGILL. Test the build matched to this host instead.
+        # Same substitution config-fermat.sh already makes above, and the loop above has already
+        # built these, so it costs no extra build time.
+        if [[ $HOSTTYPE == aarch64 || $HOSTTYPE == arm64 ]]; then MSFX=_NOSIMD; else MSFX=; fi
+        for word in '' 1word 2word 3word 4word nword; do
+            ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 127 -bmax 20 )
+        done
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,6 @@ jobs:
     name: Mlucas Linux
 
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ (matrix.os == 'ubuntu-24.04' && matrix.cc == 'clang') || matrix.os == 'ubuntu-26.04' }}
     strategy:
       matrix:
         os: [ubuntu-22.04, ubuntu-24.04, ubuntu-26.04, ubuntu-22.04-arm, ubuntu-24.04-arm, ubuntu-26.04-arm]
@@ -57,11 +56,10 @@ jobs:
             (cd obj${mode:+_$mode}; make clean)
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
-                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
+                bash -e -o pipefail -- makemake.sh mfac $word $mode
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
         done
@@ -70,10 +68,22 @@ jobs:
         grep 'warning:' build.log | sed 's/\x1B\[\([0-9]\+\(;[0-9]\+\)*\)\?[mK]//g' | awk '{ print $NF }' | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
         echo '```' >> $GITHUB_STEP_SUMMARY
         for s in tt t s m l; do time ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" |& tee -a test.log; done
+        for s in tt t s; do time ./Mlucas -prp -s $s -cpu "0:$(( $(nproc --all) - 1 ))" |& tee -a test.log; done
         if [[ $HOSTTYPE == aarch64 ]]; then
             cd ../obj_NOSIMD
         fi
-        bash -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
+        bash -e -o pipefail -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
+        # Mfactor self-test on one build only. The explicit-ISA builds above are for compile
+        # coverage: a runner may not be able to execute what they target.
+        # On ARM the native build is ASIMD, whose trial-factoring routines are stubs -
+        # twopmodq80.c's "No TF support on ARMv8!" lives under #ifdef USE_ARM_V8_SIMD - so run
+        # the NOSIMD build there instead, which takes the scalar TF path and self-tests normally.
+        # Same substitution config-fermat.sh already makes above, and the loop above has already
+        # built these, so it costs no extra build time.
+        if [[ $HOSTTYPE == aarch64 || $HOSTTYPE == arm64 ]]; then MSFX=_NOSIMD; else MSFX=; fi
+        for word in '' 1word 2word 3word 4word nword; do
+            ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 127 -bmax 20 )
+        done
     - uses: actions/upload-artifact@v7
       if: always()
       with:
@@ -84,11 +94,10 @@ jobs:
     name: Mlucas Singlethreaded Linux
 
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ !endsWith(matrix.os, '-arm') }}
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm]
-        cc: [gcc]
+        cc: [gcc, clang]
       fail-fast: false
     env:
       CC: ${{ matrix.cc }}
@@ -97,7 +106,6 @@ jobs:
     - name: Before script
       run: |
         sed -i 's/-DUSE_THREADS//' makemake.sh
-        sed -i 's/^MAX=29/MAX=28/' config-fermat.sh
         $CC --version
     - name: Script
       run: |
@@ -119,20 +127,31 @@ jobs:
             (cd obj${mode:+_$mode}; make clean)
             for word in '' 1word 2word 3word 4word nword; do
                 echo -e "\nMfactor $word\n"
-                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
+                bash -e -o pipefail -- makemake.sh mfac $word $mode
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
         done
         cd obj
-        time ./Mlucas -s m
+        for s in tt t s m; do time ./Mlucas -s $s; done
+        for s in tt t s; do time ./Mlucas -prp -s $s; done
         if [[ $HOSTTYPE == aarch64 ]]; then
             cd ../obj_NOSIMD
         fi
-        bash -- ../config-fermat.sh
+        bash -e -o pipefail -- ../config-fermat.sh
+        # Mfactor self-test on one build only. The explicit-ISA builds above are for compile
+        # coverage: a runner may not be able to execute what they target.
+        # On ARM the native build is ASIMD, whose trial-factoring routines are stubs -
+        # twopmodq80.c's "No TF support on ARMv8!" lives under #ifdef USE_ARM_V8_SIMD - so run
+        # the NOSIMD build there instead, which takes the scalar TF path and self-tests normally.
+        # Same substitution config-fermat.sh already makes above, and the loop above has already
+        # built these, so it costs no extra build time.
+        if [[ $HOSTTYPE == aarch64 || $HOSTTYPE == arm64 ]]; then MSFX=_NOSIMD; else MSFX=; fi
+        for word in '' 1word 2word 3word 4word nword; do
+            ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 127 -bmax 20 )
+        done
     - uses: actions/upload-artifact@v7
       if: always()
       with:
@@ -148,8 +167,15 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm]
-        container: ["ubuntu:16.04", "ubuntu:18.04", "ubuntu:20.04", "alpine:3.8", "alpine:3.12", "alpine:3.16", "alpine:3.20", "alpine:3.24"]
+        container: ["ubuntu:14.04", "ubuntu:16.04", "ubuntu:18.04", "ubuntu:20.04", "alpine:3.8", "alpine:3.12", "alpine:3.16", "alpine:3.20", "alpine:3.24"]
         cc: [gcc, clang]
+        exclude:
+          # Ubuntu 14.04's clang (3.4-era) predates AArch64 NEON 'v'-register asm syntax, so it
+          # cannot assemble the inline asm in dft_macro.c on ARM ("unknown register name 'v0'").
+          # Its gcc builds fine there, as do both compilers on x86-64, so drop only this one combo.
+          - os: ubuntu-24.04-arm
+            container: "ubuntu:14.04"
+            cc: clang
       fail-fast: false
     env:
       CC: ${{ matrix.cc }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -41,10 +41,21 @@ jobs:
...
+            # GitHub runners can build AVX-512 but often cannot execute it (SIGILL at runtime). Probe
+            # whether AVX-512 actually runs here; if not, skip running the AVX-512-mode test binaries
+            # below (they are still built above, for compile coverage):
+            if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm0,%%zmm0,%%zmm0":::"zmm0");return 0;}' | ${CC:-cc} -mavx512f -x c -o conftest_avx512run - >/dev/null 2>&1 && ./conftest_avx512run >/dev/null 2>&1; then
+                run_avx512=1
+            else
+                run_avx512=0
+                echo "::warning title=AVX-512 execution skipped::this runner's CPU cannot execute AVX-512; AVX-512-mode test binaries were built but will not be run."
+            fi
+            rm -f conftest_avx512run
```
tdulcet: See https://github.com/primesearch/Mlucas/pull/83#discussion_r3597177681. Same for the occurrences below.

tdulcet (changes requested):
Thanks for fixing these breakages.

MarkRose:
Per your note (and the #83 discussion): the run-guarded Mfactor steps now run the real `test_fac()` self-test (`-m 127 -bmax 20`, asserting the "self-tests completed successfully" marker) instead of `-h`, in all four AVX-512 run-guarded jobs plus the remaining plain occurrence. This matches the self-test form used in #82 (general blocks) and #83 (Windows), so the Mfactor CI invocation is now uniform across all jobs.

.github/workflows/ci.yml
MarkRose: I would like to keep the per-mode Mfactor self-test (`-m 127 -bmax 20`) rather than fall back to `-h` here, for the coverage it adds. Right now the only compute the CI actually runs is the native build's `./Mlucas -s ...` self-test — which on the GitHub x86 runners is AVX2. The explicit SSE2/AVX/AVX2/AVX512 binaries only get `-h`, so **Mfactor's factoring code is never exercised in any SIMD mode**. Per-mode self-tests close that gap: they run on SSE2/AVX/AVX2 today (all executable on the runners) and are exactly what would catch a SIMD-specific factoring regression — e.g. the kind of AVX-512 asm miscompile that has bitten this codebase before, which `-h` can never surface.

On the AVX-512 complexity specifically: the run-guard is deliberately forward-looking. It skips *running* AVX-512 today (the runners SIGILL), while still building it for compile coverage — but it's written so that the moment GitHub's runners gain AVX-512 execution (which seems likely before long), the AVX-512 self-test starts running automatically with no further CI changes. And this is the "run the actual Mfactor self-test" long-term goal you noted in the #83 discussion.

Happy to adjust the mechanics (e.g. factor the probe into a helper to reduce the repetition across jobs) if the duplication is the main concern — but I think the self-test coverage is worth keeping over `-h`. What do you think?

MarkRose: Following up with data: I pulled the actual CI history and posted the numbers on the linked #83 thread — GitHub's runners land on an AVX-512-capable box about 11% of the time on Windows and 21% on Linux, across ~230 sampled runner-jobs. So the execution-probe-and-warn approach here isn't just forward-looking, it's actively catching real runs where the self-test coverage matters. See [this comment](https://github.com/primesearch/Mlucas/pull/83#discussion_r3663250857) for the full breakdown.

tdulcet: It seems I am not explaining this very clearly. I fully agree that should run the full Mfactor self-test instead of just the `-h` output, which is why I originally suggested that in #83. What I meant above is that we should not be running the cross compiled builds in these jobs, which is error prone as you noted and is an existing bug on the main branch. Instead, I was suggesting that we test the native Mfactor binary (compiled with `-march=native` or similar), as these jobs already do below when running the Mlucas self-tests and `config-fermat.sh` script. (In the future we may want to test the cross compiled builds, but that should likely be in a separate job to avoid a runtime explosion, and it should test both Mlucas and Mfactor.)

This commit shows exactly what I am proposing (among other unrelated improvements): https://github.com/primesearch/Mlucas/pull/230/changes/6ae9f8fccc370a43421f581778e1c2788282bd18. If it is easier, I could push the suggested changes directly to this branch.

MarkRose: Ah, you mean the different instruction set extensions. I got you now. I was a bit confused by what you meant by cross-compile, since that typically means compiling for an entirely different architecture, like making an ARM build on an x86 host.

.github/workflows/ci.yml
```diff
@@ -66,10 +64,21 @@ jobs:
...
         fi
-        bash -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
+        bash -e -o pipefail -- ../config-fermat.sh -cpu "0:$(( $(nproc --all) - 1 ))"
+        # Mfactor self-test on the native build only. The explicit-ISA builds above are for
+        # compile coverage: a runner may not be able to execute what they target.
+        cd ../obj_mfac
+        if [[ $HOSTTYPE == aarch64 || $HOSTTYPE == arm64 ]]; then
+            # Trial factoring is unimplemented on ARMv8 - twopmodq80.c asserts
+            # "No TF support on ARMv8!" - so test_fac cannot run. Smoke-test the binaries instead.
+            for word in '' 1word 2word 3word 4word nword; do ./Mfactor${word:+_$word} -h; done
```
tdulcet: Maybe run the NOSIMD Mfactor build on ARM instead, similar to what we do above with the `config-fermat.sh` script, so we could get some additional testing of Mfactor. This also occurs below.

MarkRose:
You're right, and thanks — the ARM skip was too broad. Now runs the NOSIMD build there instead.

The `"No TF support on ARMv8!"` stubs in `twopmodq80.c` are all under `#ifdef USE_ARM_V8_SIMD`, so they apply to the **ASIMD** build only. A NOSIMD build takes the scalar TF path and self-tests normally — I'd guarded on the machine when the actual constraint is the build mode. It also costs nothing: the build loop already produces those binaries, and it's the same substitution `config-fermat.sh` makes a few lines above.

Measured on a Raspberry Pi 4 (aarch64, gcc 12.2.0), NOSIMD, all six word sizes — no ARMv8 assert, `Factoring self-tests completed successfully`, and every run completing with `M(127) has 0 factors in range k = [0, 16336320], passes 0-15`:

| word | time | trial divides |
| --- | --- | --- |
| base | 5.7s | 1,658,492 |
| 1word | 1.5s | 1,658,492 |
| 2word | 3.2s | 779,513 |
| 3word | 5.8s | 778,312 |
| 4word | 10.2s | 776,362 |
| nword | 19.5s | 772,386 |

46 s total on a Pi, so it will be well under that on a runner.

Also rebased onto current `main` and squashed to one commit — the branch was 200 commits behind and the `ci.yml` hunks had drifted badly against #222's per-word object directories.

## This now needs #83

Worth flagging, because removing the `|| true` uncovers something it was hiding. `factor.h` errors with `#error P2WORD may not be used together with USE_FMADD!` for 2word/3word/4word wherever `USE_FMADD` is auto-defined — which is every FMA-capable target. That includes the **native** build on any modern x86 runner, so this PR's own self-test loop cannot build three of its six word sizes:

```
native mfac on current main:   <base> OK  1word OK  2word FAIL  3word FAIL  4word FAIL  nword OK
same tree carrying #83:        all six OK, and the loop then runs 6/6 with rc=0
```

The `#error` dates to the 2009 import, so those builds have presumably never worked — the mask just meant nobody saw it. #83 fixes it by `#undef`-ing `USE_FMADD` for multiword p so it falls back to standard modmul, which makes that part of #83 rather more than the portability cleanup it looks like.

So the order is **#83 + #114 → #103**; both are now recorded in #224.

MarkRose:
Rebased onto current `main` and finished the outstanding review item.

> What I meant above is that we should not be running the cross compiled builds in these jobs [...] Instead, I was suggesting that we test the native Mfactor binary

The Linux and macOS jobs had already moved to the native `obj_mfac${word:+_$word}${MSFX}` build, but **Windows had not** — it still ran every mode's binary inside the build loop. That's now fixed, so all eight Mfactor self-test sites across every job family run the native build only:

```
  Linux               native only
  Singlethread-Linux  native only
  ASan-Linux          native only
  macOS               native only
  Singlethread-macOS  native only
  ASan-macOS          native only
  Windows             native only     <- was inside the per-mode loop
  ASan-Windows        native only
```

The explicit-ISA builds stay for compile coverage; what changed is what gets executed. And your reasoning has since been demonstrated the hard way: the FreeBSD jobs added in #251 run per-mode binaries, and three of the four are red on `main` right now because the AVX-512 build traps with `SIGILL` before reaching the self-tests below the loop.

> Maybe run the NOSIMD Mfactor build on ARM instead

That's in as `MSFX=_NOSIMD`, since ASIMD's trial-factoring routines are stubs.

## Rebase notes

Fourteen commits of `main` had landed, five conflicting hunks. Two things worth knowing about how I resolved them:

- **Sanitizer flags** now use `main`'s `sed -i 's/-O3 -flto/-Og -fsanitize=.../' makemake.sh` rather than this branch's `CFLAGS=... >> $GITHUB_ENV`. #254 removed `CFLAGS ?=` from the generated Makefile, so the environment route no longer takes effect at all.
- **ARM detection on macOS** now uses `main`'s `[[ $HOSTTYPE == arm64 ]]` in place of this branch's `sysctl -n hw.optional.neon` probe. The added PRP self-test coverage from this branch is kept.

## Two masks left, on purpose

`Linux-K1OM` and `FreeBSD` both still carry `|| true` on their `makemake.sh mfac` calls. Both jobs were added to `main` after this branch was written, and neither mask can come off yet:

- **K1OM** needs #257 first. `platform.h` makes `USE_IMCI512` imply `USE_AVX512`, so `factor.c` compiles the AVX-512 vector sieve for Knights Corner and all six word variants fail to build. Without #257 that job goes red immediately.
- **FreeBSD** absorbs the same `P2WORD/P3WORD/P4WORD may not be used together with USE_FMADD` failures that #83 removes.

Which is also the honest status of this PR: **it stays red until #83 lands**, for exactly the reason it exists — those failures are real and this stops hiding them. Measured on current `main`, the `2word`/`3word`/`4word` variants fail in every FMA-capable mode.

.github/workflows/ci.yml
```diff
@@ -97,7 +106,6 @@ jobs:
     - name: Before script
       run: |
         sed -i 's/-DUSE_THREADS//' makemake.sh
-        sed -i 's/^MAX=29/MAX=28/' config-fermat.sh
```
tdulcet, line 100: We likely want to keep this line, to keep the CI runtime manageable.

.github/workflows/ci.yml
```diff
@@ -84,11 +94,10 @@ jobs:
     name: Mlucas Singlethreaded Linux
 
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ !endsWith(matrix.os, '-arm') }}
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm]
-        cc: [gcc]
+        cc: [gcc, clang]
```
tdulcet: I added clang just as an extra test for the integration branch, but it is likely redundant otherwise and not worth a permanent CI job.

.github/workflows/ci.yml
```diff
@@ -291,16 +317,31 @@ jobs:
     - name: Before script
       run: |
         sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
```
tdulcet: This was my fault in #251, but since the `makemake.sh` script was updated, the `-flto` needs to be removed on this line (and similarly for the other ASan/TSan jobs), as otherwise this would not actually run the sanitizers as expected.

.github/workflows/ci.yml
```diff
@@ -291,16 +317,31 @@ jobs:
...
+        # ("invalid CFI advance_loc expression"). NOSIMD exercises the same C-level logic.
+        bash -e -o pipefail -- makemake.sh use_hwloc nosimd
+        (cd obj_nosimd; make clean)
+        for word in '' 1word 2word 3word 4word nword; do
+            echo -e "\nMfactor $word\n"
+            bash -e -o pipefail -- makemake.sh mfac $word nosimd
+            (
+                cd obj_mfac_nosimd
+                make clean
+                mv -v build.log build_${word}.log
```
tdulcet: This section needs to be updated. Specifically, the directory needs to be fixed and the no longer needed `mv` command should be removed. This also occurs below.

tdulcet (changes requested):
Thanks for the rebases and updates.

This PR includes several unrelated CI changes from the #238 integration branch (a failed rebase?), so maybe it would be worth starting from a clean slate. Of course, many of those changes are improvements that would be needed anyway, so feel free to keep them if you prefer.

> `Linux-K1OM` and `FreeBSD` both still carry `|| true` on their `makemake.sh mfac` calls. Both jobs were added to `main` after this branch was written

Please do fix the K1OM and FreeBSD jobs as well.

MarkRose:
All four inline comments addressed, plus K1OM and FreeBSD.

**`MAX=28` in the singlethreaded jobs** — restored, Linux and macOS both. It stays *added* in the sanitizer jobs, for the same runtime reason.

**clang in the singlethreaded Linux matrix** — dropped, back to `cc: [gcc]`.

## The sanitizer `sed`

You're right, and it's worse than the `-flto` alone: since #254 the string `-O3 -flto` doesn't occur in `makemake.sh` at all, so the substitution matches nothing and the sanitizer jobs have been building ordinary optimized binaries. Applying each form to current `main` and reading back the generated `obj_nosimd/Makefile`:

```
old sed ->  CFLAGS = -fdiagnostics-color -std=gnu99 -Wall -g -O3 -flto=auto
new sed ->  CFLAGS = -std=gnu99 -Wall -g -Og -fsanitize=address,undefined
```

No `-fsanitize` anywhere in the first one. So ASan-Linux, ASan-macOS, ASan-Windows, TSan-Linux and TSan-macOS have all been reporting green on uninstrumented builds.

With the fix, the instrumented build works end to end locally — `makemake.sh use_hwloc nosimd` returns 0, the binary links 50 `__asan`/`__ubsan` symbols, and `-s tt` completes rc=0 with 29 residues and no sanitizer diagnostics.

I fixed it by rewriting the generated Makefile's `CFLAGS` line outright rather than patching `-O3`:

```
sed -i 's/^CFLAGS = .*/CFLAGS = -std=gnu99 -Wall -g -Og -fsanitize=address,undefined/' makemake.sh
```

Two reasons for that form. `-O3` also appears on the SIMD-probe compile at `makemake.sh:380`, which shouldn't be instrumented and would be caught by a bare `s/-O3/`. And rewriting the whole line drops `-flto=auto`, which is now appended separately by the `try_lto` probe and so would survive any edit to the `C_ARGS` line. Applied at all five sites.

While in there: `makemake.sh:453-455` still says the generated Makefile uses `CFLAGS ?=` and defers to an exported `CFLAGS`. #254 changed that to `CFLAGS =`, so the comment now documents a route that no longer works — it is what cost me a rebase on this branch, and it will mislead the next person to touch these jobs. Fixed in #83, which already touches that file, so there's no conflict with this one.

## The Mfactor object directory

Fixed to `obj_mfac${word:+_$word}_nosimd`, and the `mv` is gone — #222 gave each word variant its own directory, so `build.log` no longer collides. Three sites: ASan Linux, macOS and Windows.

## K1OM and FreeBSD

Both done, and FreeBSD needed more than the mask:

- **FreeBSD** was still running the cross-ISA binaries inside the mode loop (`[ ! -x Mfactor… ] || ./Mfactor… -h`) — exactly what SIGILLs three of the four releases on `main` today. It now gets the same treatment as every other job: build every mode, self-test only the native one. Its `config-fermat.sh` call also gains `-e -o pipefail`, which it was missing. Kept POSIX `sh` throughout, since that job runs `usesh: true`.
- **K1OM**'s `|| true` is off. That one needs #257 first: `platform.h` makes `USE_IMCI512` imply `USE_AVX512`, so the AVX-512 vector sieve gets compiled for Knights Corner and all six word variants fail to assemble. Without #257 the job goes red the moment the mask comes off — which is the point of the PR, but it is another prerequisite.

So the merge order is now **#83 + #114 + #257 → this**. Updated in #224.

## The unrelated changes

Fair — those did arrive with the integration branch rather than being written for this PR. What's in that category:

- `ubuntu:14.04` added back to the container matrix, with the single `ubuntu-24.04-arm` + clang exclusion
- TSan-macOS re-enabled, on `-s tt` rather than `-s m`
- the PRP self-tests, `-prp -s {tt,t,s}`
- widening `-s m` to `-s {tt,t,s,m}` in the singlethreaded and sanitizer jobs

(The `-e -o pipefail` on the `config-fermat.sh` calls I'd count as in scope — it's the same masking.)

I've kept them per your note, since each is a coverage gain we'd want anyway, but they're separable — say the word and I'll lift them into their own PR so this one is purely the unmasking.

tdulcet (approved):

PR #104, opened 2026-07-16
Fermat-mod: soft-skip 2-radix sets (fixes #101), and correct the F14/F15 self-test FFT lengths

### LINKED ISSUE #101: Error with 1K and 2K FFT lengths: `product of radices [2] != vector length [1] in BIT_REVERSE_INT`
The 1K and 2K FFT lengths are partially broken. For example, when running `./Mlucas -f 13 -iters 100 -fft 1`:
```
INFO: Maximum recommended exponent for FFT length (1 Kdbl) = 22788; p[ = 8192]/pmax_rec = 0.3594874495.
F13: using FFT length 1K = 1024 8-byte floats, initial residue shift count = 6844
This gives an average    8.000000000000000 bits per digit
Using complex FFT radices        32        16
fermat_mod_square: Init threadpool of 1 threads
ERROR: product of radices [2] != vector length [1] in BIT_REVERSE_INT
ERROR: Function bit_reverse_int, at line 257 of file ../src/br.c
Assertion '0' failed: Exiting.
Aborted (core dumped)
```
or `./Mlucas -f 14 -iters 100 -fft 2`:
```
INFO: Maximum recommended exponent for FFT length (2 Kdbl) = 44888; p[ = 16384]/pmax_rec = 0.3649973267.
../src/Mlucas.c:3084:58: runtime error: shift exponent 64 is too large for 64-bit type 'long unsigned int'
F14: using FFT length 2K = 2048 8-byte floats, initial residue shift count = 0
This gives an average    8.000000000000000 bits per digit
Using complex FFT radices        32        32
fermat_mod_square: Init threadpool of 1 threads
ERROR: product of radices [2] != vector length [1] in BIT_REVERSE_INT
ERROR: Function bit_reverse_int, at line 257 of file ../src/br.c
Assertion '0' failed: Exiting.
Aborted (core dumped)
```
We should either fix the bug or consider temporarily disabling these FFT lengths. The 1K FFT length was previously removed from the `config-fermat.sh` script in #23.

Also see #7.

MarkRose:
Rescoped after review — the original approach (retiring the 1K FFT length) was wrong, and this is now a much smaller change.

## 1K is not broken; 2-radix Fermat sets are

`M22679 -fftlen 1` on current `main`:

| build mode | result |
| --- | --- |
| `nosimd` | `Res64: 27F79B322B31FDA9`, MaxErr 0.3125 — matches the `MersVec` entry |
| `sse2` | same |
| `avx` | `maxerr 0.4939` — excessive roundoff, radix set rejected |
| `avx2` | `maxerr 0.4937` |

So 1K works, and disabling it would have removed a working length. **1K Mersenne is untouched by this PR.**

What actually aborts is Fermat-mod, and the invariant is the *radix count*, not the FFT length. `radix{16|32}_wrapper_square` needs at least one intermediate radix (≥3 total); with 2 it hands `bit_reverse_int` a radix product that does not match the vector length and dies — issue #101. Exactly three `(length, radix set)` pairs in the whole table have only 2 radices — **1K sets 0 and 1, and 2K set 1** — and 1K has no 3-radix set at all.

`fermat_mod_square()` now soft-skips those, using the same `WARN` + `ERR_ASSERT` path the `n/radix0` checks already use, so the self-test moves on instead of aborting.

## F14 and F15 named FFT lengths that cannot run

Auditing every `FermVec` row — does the stated length actually reproduce the stated `Res64`?

| row | stated | ok? |
| --- | --- | --- |
| F14 | 1K | ✗ aborts — residues are **4K** values |
| F15 | 2K | ✗ aborts — residues are **4K** values |
| F16 | 4K | ✓ |
| F17 | 8K | ✓ |
| F18 | 16K | ✓ |
| F19 | 32K | ✓ |
| F20 | 64K | ✓ |
| F21 | 128K | ✓ |

Only F14 and F15 are wrong. All six of their triplets (100/1000/10000 iterations) reproduce exactly at 4K, identically under `nosimd`, `sse2` and `avx2` — so no new reference data is needed, only the length column. **Those two Fermat numbers have had no working self-test coverage at all.**

## Two things for a decision

1. **F14, F15 and F16 now all sit at `fftLength = 4`.** The "find the table row for this FFT length" lookup takes the first row with `fftLength >= i`, so 4K now resolves to F14 rather than F16. That path is only reached when an FFT length is given without a Fermat index. The alternative is to comment the two rows out, as was already done for F13 — happy to switch if you prefer that.
2. **Not fixed here:** `-f 14` on its own still selects 1K via `get_default_fft_length()` rather than from the table, so it finds nothing usable — it now reports a clean skip instead of crashing. Raising the Fermat minimum there is a separate change.

## Verification

- `-f 13 -fft 1` (the #101 repro): soft-skips both radix sets, no abort
- `-f 14 -fft 4`, `-f 15 -fft 4`: correct residues, matching the table
- `-f 14 -fft 2 -radset 1` (2K `{32,32}`): soft-skips
- `M22679 -fftlen 1`: unchanged in all four build modes
- `-s tt`: rc=0, 15 lengths, 19 residues

Resolves #101.


src/get_fft_radices.c
```diff
@@ -116,14 +116,13 @@ int	get_fft_radices(uint32 kblocks, int radix_set, uint32 *nradices, uint32 radi
...
 	{
 	case 1 :						/* 1K */
-		switch(radix_set) {
-		case 0 :
-			numrad = 2; rvec[0] = 32; rvec[1] = 16; break;
-		case 1 :
-			numrad = 2; rvec[0] = 16; rvec[1] = 32; break;
-		default :
-			*nradices = 2;	return ERR_RADIXSET_UNAVAILABLE;
-		}; break;
```
tdulcet, line 126: We should keep this switch statement commented out, to make it easier, in case we decide to add full support for the 1K FFT length.

tdulcet (changes requested):
Thanks for fixing this.

To make it clearer, we might now want to comment out the 1K self-tests: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L3515-L3516 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L3670-L3671 and increase the default FFT length for F14: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L3846-L3848

tdulcet:
I am able to reproduce the error with 2K FFT length as well. See #101, which I updated.

However, both the 1K and 2K Mersenne self-tests sometimes do work, so maybe we do not want to completely disable them:
```
           Mlucas selftest running.....

/****************************************************************************/

INFO: Unable to find/open mlucas.cfg file in r+ mode ... creating from scratch.
No CPU set or threadcount specified ... running single-threaded.
Set affinity for the following 1 cores: 0.
 worktodo.txt file not found...using user-supplied command-line exponent p = 22679
INFO: Maximum recommended exponent for FFT length (1 Kdbl) = 22788; p[ = 22679]/pmax_rec = 0.9952167808.
M22679: using FFT length 1K = 1024 8-byte floats, initial residue shift count = 20848
This gives an average   22.147460937500000 bits per digit
Using complex FFT radices        32        16
Using 1 threads in carry step
../src/Mlucas.c:3084:58: runtime error: shift exponent 64 is too large for 64-bit type 'long unsigned int'
100 iterations of M22679 with FFT length 1024 = 1 K, final residue shift count = 3818
Res64: 27F79B322B31FDA9. AvgMaxErr = 0.219754464. MaxErr = 0.312500000. Program: E21.0.2
Res mod 2^35 - 1 =          23663495240
Res mod 2^36 - 1 =           2817376218
Clocks = 00:00:00.067

NTHREADS = 1
 worktodo.txt file not found...using user-supplied command-line exponent p = 22679
INFO: Maximum recommended exponent for FFT length (1 Kdbl) = 22788; p[ = 22679]/pmax_rec = 0.9952167808.
M22679: using FFT length 1K = 1024 8-byte floats, initial residue shift count = 3818
This gives an average   22.147460937500000 bits per digit
Using complex FFT radices        16        32
Using 1 threads in carry step
100 iterations of M22679 with FFT length 1024 = 1 K, final residue shift count = 10172
Res64: 27F79B322B31FDA9. AvgMaxErr = 0.220842634. MaxErr = 0.312500000. Program: E21.0.2
Res mod 2^35 - 1 =          23663495240
Res mod 2^36 - 1 =           2817376218
Clocks = 00:00:00.067

INFO: 2 of 2 radix-sets at FFT length 1 K passed - writing cfg-file entry.
/ **************************************************************************** /

NTHREADS = 1
 worktodo.txt file not found...using user-supplied command-line exponent p = 44657
INFO: Maximum recommended exponent for FFT length (2 Kdbl) = 44888; p[ = 44657]/pmax_rec = 0.9948538585.
M44657: using FFT length 2K = 2048 8-byte floats, initial residue shift count = 10172
This gives an average   21.805175781250000 bits per digit
Using complex FFT radices         8         8        16
WARN: At line 84 of file ../src/radix8_ditN_cy_dif1.c:
CY routines with radix < 16 do not support shifted residues!

 Return with code ERR_ASSERT
Error detected - this radix set will not be used.
NTHREADS = 1
 worktodo.txt file not found...using user-supplied command-line exponent p = 44657
INFO: Maximum recommended exponent for FFT length (2 Kdbl) = 44888; p[ = 44657]/pmax_rec = 0.9948538585.
M44657: using FFT length 2K = 2048 8-byte floats, initial residue shift count = 20344
This gives an average   21.805175781250000 bits per digit
Using complex FFT radices        32        32
Using 1 threads in carry step
100 iterations of M44657 with FFT length 2048 = 2 K, final residue shift count = 6008
Res64: C15789922E67B055. AvgMaxErr = 0.210714286. MaxErr = 0.312500000. Program: E21.0.2
Res mod 2^35 - 1 =           1083638452
Res mod 2^36 - 1 =          32161054300
Clocks = 00:00:00.058

INFO: 1 of 2 radix-sets at FFT length 2 K passed - writing cfg-file entry.
/ **************************************************************************** /
```
This may require some further debugging.

MarkRose:
Thanks for the review. Addressed the inline suggestion: the original 1K case-1 radix sets are now kept as a comment in `get_fft_radices.c` (with the case still returning `ERR_FFTLENGTH_ILLEGAL`), so re-adding full 1K support later is easy.

On the other two suggestions:

**Commenting out the 1K self-test entries** — the reference tables are fixed-count: `numTest` (136) is asserted to exactly equal the non-sentinel `MersVec[]` size (`ASSERT(MersVec[numTest-1].fftLength != 0 && MersVec[numTest].fftLength == 0, ...)`), and the self-test loops use hardcoded index ranges (e.g. `for(i = 68; i < numTest; i++)`). Commenting out the first (1K) entry would require decrementing `numTest` and would shift every subsequent index. This PR instead skips any sub-`MIN_FFT_LENGTH_IN_K` reference entry at runtime (`if(iarg < MIN_FFT_LENGTH_IN_K) continue;`), which suppresses the unsupported 1K entries in both the Mersenne and Fermat self-test loops without disturbing the table indexing. Happy to switch to commenting-out + `numTest` adjustment if you prefer it despite the index churn.

**Increasing the default FFT length for F14** — I looked into this and it does not restore F14 coverage: `-f 14` auto-selects a **2K** FFT from the exponent (16384 bits) regardless of the `FermVec[]` `fftLength` column, and **F14 at 2K itself crashes** in `bit_reverse_int` ("product of radices != vector length"), i.e. the same failure class as the 1K bug but for the 2K Fermat radix sets. So bumping F14s table entry to 2K/4K just un-skips a run that then aborts. F14 self-test was already broken on `main` (its 1K default hit #101), so leaving it skipped is the safe outcome here. The 2K-Fermat `bit_reverse_int` failure for small F_m looks like a separate, deeper bug — I can file a dedicated issue for it if useful.

tdulcet:
Thanks for the additional information.

This PR might be no longer necessary. See https://github.com/primesearch/Mlucas/pull/104#issuecomment-4995099138 above and https://github.com/primesearch/Mlucas/pull/115#pullrequestreview-4724620748.

MarkRose:
Agreed. Now that #115 fixes the underlying `bit_reverse_int` "product of radices" abort (the same crash as #101, at both 1K and 2K), raising `MIN_FFT_LENGTH_IN_K` to 2 here is redundant — and as your M22679@1K self-test output shows, it would also disable a length that does work. I'm inclined to **close this in favor of #115** unless you'd prefer to keep a lighter-weight guard. (The `shift exponent 64 is too large` UB you saw at Mlucas.c:3084 in that 1K run is a separate issue, fixed by #78.)

MarkRose:
Correction to my previous comment — I tested this empirically and **#115 does not fix the 1K case**, so this PR is still needed.

On the #115 branch (i.e. with the `bit_reverse_int` fix but *without* this PR), the original #101 1K repro still aborts:

```
$ Mlucas -f 13 -iters 100 -fft 1
Using complex FFT radices        32        16
ERROR: product of radices [2] != vector length [0] in BIT_REVERSE_INT
ERROR: Function bit_reverse_int, at line 261 of file ../src/br.c
Aborted (core dumped)
```

#115 does resolve the **2K** reports from #101 ({8,8,16} runs, {32,32} is gracefully skipped by its rt1-overrun guard), but **1K** can only be factored into {32,16} — two radices — and the `radix{16,32}_wrapper_square` scheme needs at least one intermediate radix (≥3 total). So it still fails downstream even with #115's zero-radix/length-1 handling.

So #104 (retiring 1K) and #115 (fixing 2K) are complementary rather than redundant. I'll keep this open. Happy to adjust the approach if you'd prefer a narrower disable.

tdulcet:
I believe this PR needs to be reevaluated, as the 1K FFT length does work sometimes, especially for Mersenne numbers. For example, it seems that you successfully used it in #225. Therefore, completely disabling the 1K FFT length may not be what we want to do.

Ernst enabled the 1K FFT length in v21, along with adding the new teensy self-test for 1, 2, 3, 4, 5, 6 and 7 K, but maybe he did not have a chance to fully test or debug it.

MarkRose:
You're right, and measuring it properly turned up two things I'd got wrong.

## 1K works — it's build-mode dependent

That's why our results differed. `M22679 -fftlen 1`, all on current `main` (`2ff1a116`):

| build mode | result |
| --- | --- |
| `nosimd` | **`Res64: 27F79B322B31FDA9`, MaxErr 0.3125** |
| `sse2` | **same residue**, MaxErr 0.3125 |
| `avx` | `maxerr 0.4939` — excessive roundoff, radix set rejected |
| `avx2` | `maxerr 0.4937` — same |

Your residue matches the `nosimd`/`sse2` result exactly, and matches the `MersVec` table entry. I'd been testing AVX2 throughout and concluded 1K was simply broken. It isn't — so completely disabling it is the wrong call, as you say.

Worth noting separately: the 1K `MersVec` reference is therefore a scalar/SSE2-only value, and `p/pmax_rec = 0.9952` leaves almost no margin, which is presumably why the wider AVX lanes tip it over 0.5.

## What actually fails

Sorted by severity, all four modes:

| case | outcome | severity |
| --- | --- | --- |
| **1K Fermat** (`-f 13 -fft 1`) | aborts in `bit_reverse_int` | **hard — the only crash** |
| 1K Mersenne on AVX/AVX2 | excessive roundoff | soft — self-test skips the length |
| 2K Mersenne rs0 `{8,8,16}` | "radix < 16 … no shifted residues" | soft — clean skip |

Only 1K Fermat actually aborts, and that's issue #101's repro. Everything else already degrades gracefully.

I also rebuilt #115 merged onto current `main`: 1K Fermat **still aborts** there, so the two remain complementary. #115 does now cleanly "Skipping" the 2-radix Fermat sets at 2K, which is a nice improvement.

## I was wrong about F14 too — and the fix is one character

On 17 July I said raising F14's FFT length wouldn't help because 2K crashes. I only tested 2K. **F14 works at 4K**, and the residue is `DB9AC520C403CB21` — which is *exactly what's already in the `FermVec` row*.

So the row isn't missing data; its `fftLength` column is just wrong. Same for F15. I audited every row:

| row | stated | stated length produces the stated Res64? |
| --- | --- | --- |
| F14 | 1K | ✗ aborts — residues are **4K** values |
| F15 | 2K | ✗ aborts — residues are **4K** values |
| F16 | 4K | ✓ |
| F17 | 8K | ✓ |
| F18 | 16K | ✓ |
| F19 | 32K | ✓ |
| F20 | 64K | ✓ |
| F21 | 128K | ✓ |

F16–F21 are all correct; only F14 and F15 point at lengths that don't work. All six triplets (100/1000/10000 iterations × F14, F15) reproduce exactly at 4K, and identically under `nosimd`, `sse2` and `avx2`. So the fix needs no new residues — just `{ 1, 14u,` → `{ 4, 14u,` and `{ 2, 15u,` → `{ 4, 15u,`.

That means F14 and F15 have had **no working Fermat self-test coverage at all**, because the table sent them to lengths that abort.

## Reworking this PR

Dropping the blanket disable. New scope:

1. **Reject 1K for Fermat-mod only** — fixes #101's hard crash, leaves 1K Mersenne alone.
2. **Fix the F14/F15 `fftLength` entries** (1K→4K, 2K→4K), restoring their self-test coverage with the residues already in the table.

Gone: `MIN_FFT_LENGTH_IN_K`, the sub-minimum self-test skip, and the unconditional 1K rejection in `get_fft_radices`.

One thing I'd like your steer on. The underlying pattern is that **Fermat-mod cannot use a 2-radix set** — 1K only has 2-radix sets, and 2K rs1 `{32,32}` fails the same way. #115 already soft-skips those at 2K. Would you rather this PR reject on "1K + Fermat", or have #115 generalise its skip to cover the 1K case too, leaving this PR as just the F14/F15 table fix? Happy either way; the second is arguably tidier since #115 already owns that logic.

src/Mlucas.c
```diff
@@ -3875,8 +3875,8 @@ struct testFerm FermVec[numFerm+1] =
...
 	/*                                    [%34359738367  ][%68719476735  ]                         [%34359738367  ][%68719476735  ] */
 //	{     2,  13u, { {0xC8FC67EA3A1AC788ull, 29592689237ull, 35156594447ull}, {0xBE489C2CF00D582Aull,  3108135315ull, 47125592449ull}, {0x0ull, 0ull, 0ull} } },
-	{     1,  14u, { {0xDB9AC520C403CB21ull,   342168579ull, 59244817440ull}, {0xF111F12732CCCB0Full, 24848612524ull, 66609820796ull}, {0x78738D068D641C2Cull, 12664273769ull, 29297626750ull} } },
-	{     2,  15u, { {0x3B21A6E55ED13454ull, 28379302213ull, 15546218647ull}, {0x4784657F2A36BE74ull,   617376037ull, 44891093359ull}, {0x589BFE53458FFC14ull, 12200390606ull, 46971422957ull} } },
+	{     4,  14u, { {0xDB9AC520C403CB21ull,   342168579ull, 59244817440ull}, {0xF111F12732CCCB0Full, 24848612524ull, 66609820796ull}, {0x78738D068D641C2Cull, 12664273769ull, 29297626750ull} } },
+	{     4,  15u, { {0x3B21A6E55ED13454ull, 28379302213ull, 15546218647ull}, {0x4784657F2A36BE74ull,   617376037ull, 44891093359ull}, {0x589BFE53458FFC14ull, 12200390606ull, 46971422957ull} } },
```
tdulcet: Should these use the 2K FFT length after #115?

tdulcet (changes requested):
Thanks for investigating this.

MarkRose:
Not after #115 — after **#146**. I built the combination to check rather than reason about it.

At 2K there are exactly two radix sets, and both are unusable for Fermat-mod for reasons #115 does not touch:

- `{32,32}` — two radices, and `radix{16|32}_wrapper_square` needs at least three. That is the soft-skip this PR adds.
- `{8,8,16}` — leading radix 8 < 16, so `radix8_ditN_cy_dif1.c:84` rejects the nonzero initial shift the Fermat self-test uses: *"CY routines with radix < 16 do not support shifted residues!"* That is #119, fixed by #146.

On `main` + #115 + this PR, `-f 14 -fft 2` therefore ends with `0 of 2 radix-sets at FFT length 2 K passed`. #115 turns the 2K crash into a clean skip, which is worth having, but it does not make either set runnable.

Add #146 on top and both lengths work, reproducing the tabulated residues exactly:

| | 2K, main+#115+#104 | 2K, +#146 | table |
| --- | --- | --- | --- |
| F14 | skipped | `DB9AC520C403CB21` | `DB9AC520C403CB21` |
| F15 | skipped | `3B21A6E55ED13454` | `3B21A6E55ED13454` |

## The reference data does not change either way

Worth stating because my own PR description was sloppy about it — I wrote that the residues "are 4K values", which is wrong. A Fermat residue is a property of the modulus and the iteration count, not of the transform length; the FFT is only the multiplication algorithm. Measured: F15 at 100 iterations gives `3B21A6E55ED13454` at 4K and the identical value at 8K, and F14/F15 at 2K under #146 give the same values again. 4K is simply the smallest length that currently runs, not the length the numbers belong to.

So moving the column to 2K later is a one-character change per row with no new reference data.

## What I would suggest

Leave it at 4K in this PR, and move to 2K as a follow-up once #146 lands. If the column says 2K before #146 is in, both rows are skipped at runtime and F14/F15 go back to having no self-test coverage — which is the exact defect this PR exists to fix, just spelled differently.

The counter-argument is real though, and it is the one thing 4K costs us: F14, F15 and F16 then all sit at `fftLength = 4`, and the "first row with `fftLength >= i`" lookup resolves 4K to F14 rather than F16. Moving to 2K dissolves that. So if you would rather sequence #146 first and have this PR go straight to 2K, say so and I will flip both rows and record #146 as a prerequisite in #224.

(Rebased onto current `main` — it was 30 commits behind. Re-verified after: `-f 14 -fft 4` and `-f 15 -fft 4` give the tabulated residues, `-f 13 -fft 1` (the #101 repro) exits 0 instead of aborting, and `-s tt` is rc=0 with 19 residues.)

tdulcet (changes requested):
> Leave it at 4K in this PR, and move to 2K as a follow-up once #146 lands. If the column says 2K before #146 is in, both rows are skipped at runtime and F14/F15 go back to having no self-test coverage

Considering that there is no built-in Fermat self-test (which is why we have the `config-fermat.sh` script), I would suggest switching these two FFT lengths to 2 K now. That way we would also not have to remember to change it again later.

MarkRose:
With #146 merged, the condition from my last comment is met, so both rows are now **2K** rather than 4K — which is what you originally asked about.

Verified against current `main` carrying #146, with this PR on top:

```
  F14  -f 14 -fft 2 -iters 100  ->  Res64: DB9AC520C403CB21   (table: DB9AC520C403CB21)
  F15  -f 15 -fft 2 -iters 100  ->  Res64: 3B21A6E55ED13454   (table: 3B21A6E55ED13454)
```

1K stays impossible for both — it has no 3-radix set at all, so it is rejected by this PR's own soft-skip rather than by anything #146 touches:

```
  F14  -f 14 -fft 1  ->  skipped: "requires at least 3 FFT radices (this set has 2)"
```

Two consequences worth noting:

- **No reference data changed, and none would have at any length.** I said in the PR description that the residues "are 4K values"; that was wrong and I have corrected it in the commit message. A Fermat residue is a property of the modulus and the iteration count, not the transform length — F15 at 100 iterations gives `3B21A6E55ED13454` at 2K, 4K and 8K alike. 4K was only ever the smallest length that happened to run.
- **The lookup collision I flagged is gone.** At 4K, F14/F15/F16 all shared a length and the "first row with `fftLength >= i`" lookup would have resolved 4K to F14 rather than F16. At 2K they are distinct again, so that concern no longer applies and there is nothing to decide.

Rebased onto current `main` (32 behind) and re-verified: `-f 13 -fft 1` (the #101 repro) exits 0, and `-s tt` is rc=0 with 23 residues.

Still not fixed here, unchanged from before: `-f 14` with no `-fft` picks 1K via `get_default_fft_length()` rather than from the table, so it finds nothing usable and reports a clean skip. Raising the Fermat minimum there is a separate change.

MarkRose:
Already done — our messages crossed. Both rows went to 2 K in `f0ea0843`, pushed a few hours after your review, so the branch as it stands is what you are asking for. Details and the verification are in the comment just above.

And your reasoning is better than mine was: I was treating a premature 2 K as leaving F14/F15 uncovered, but nothing sweeps `FermVec[]` automatically — the rows are only reached on demand via `-f`, with `config-fermat.sh` driving the real Fermat runs. So the downside I was guarding against was smaller than I made it sound, and switching once rather than twice is the right call regardless.

tdulcet (approved):
Thanks.

PR #106, opened 2026-07-16
Skip the cmd.exe-incompatible shell-printf path expansion on Windows (fixes #50)

### LINKED ISSUE #50: Error on Windows builds: `'printf' is not recognized as an internal or external command, operable program or batch file.`
The error occurs when starting both Mlucas and Mfactor on Windows. It comes from this function, which attempts to run the external `printf` command: https://github.com/primesearch/Mlucas/blob/b6dda5253c9edfc34d5fbcdad12704bb61941370/src/util.c#L9354-L9454 We need to either find a portable way to determine the path or disable this feature on Windows.

MarkRose:
Fixes #50.

On Windows, launching Mlucas/Mfactor fails with `'printf' is not recognized as an internal or external command`. `set_mlucas_path()` (src/util.c) shell-expands `MLUCAS_PATH` (or the compiled-in `MLUCAS_DEFAULT_PATH` like `"$HOME/.mlucas.d/"`) by building `printf "<path>"` and running it through `popen()`. On Windows `popen()` invokes cmd.exe, which has no `printf`, so startup aborts.

This adds a Windows-only branch (guarded by `OS_TYPE_WINDOWS || __MINGW32__` — the same compound guard platform.h uses, since platform.h routes MinGW through the Linux code path yet MinGW's `popen()` still spawns cmd.exe) that uses the `MLUCAS_PATH` env var (or the compiled default) verbatim — no shell involved — while enforcing the same invariants as the POSIX branch: length ≤ STR_MAX_LEN and a trailing slash (`/` or `\`). This is the "disable the shell-expansion feature on Windows" option the issue endorses.

The POSIX code path is behaviourally unchanged, but it is not literally untouched, so it is worth reading. A follow-up commit hoists the shared `getenv`/`MLUCAS_PATH` prologue and the `out_err_check` epilogue out of both branches, moves the five POSIX-only declarations inside the `#else`, re-indents that block, and folds the old two-arm buffer sizing (`bufsize = strlen(...) + 1; ...; bufsize = (bufsize - 1) * 3 + 1;`) into a single `size_t bufsize = strlen(MLUCAS_PATH) * 3 + 1;`. The two are equivalent: when the env var is set, `MLUCAS_PATH` has just been `strcpy`'d from it, so the two arms were computing the same length. The re-indent is most of the 93/68-line diff — `git diff -w` reduces it to 38/13.

Verified: nosimd build links; on Linux, default startup, `$HOME` expansion, and the trailing-slash error path all behave as before; the Windows branch was compiled standalone (`-Wall -Wextra`, zero warnings) and its accept/reject cases checked.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -9387,90 +9387,86 @@ void set_mlucas_path(void)
 	   use MLUCAS_PATH verbatim, while enforcing the same invariants as the POSIX branch:
 	   the path must be no longer than STR_MAX_LEN and must end with a slash (fwd- or
 	   backslash, since both are directory separators on Windows).  */
-	{
-		size_t len = strlen(MLUCAS_PATH);
-		if (len == 0) /* empty path means "use the current directory", as in the POSIX branch  */
-			goto out_err_check;
-		if (len > STR_MAX_LEN) {
-			fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH is longer than STR_MAX_LEN in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_err_check;
-		}
-		if (MLUCAS_PATH[len - 1] != '/' && MLUCAS_PATH[len - 1] != '\\') {
-			fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH does not end with a slash in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_err_check;
-		}
+	size_t len = strlen(MLUCAS_PATH);
+	if (len == 0) /* empty path means "use the current directory", as in the POSIX branch  */
+		goto out_err_check;
+	if (len > STR_MAX_LEN) {
+		fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH is longer than STR_MAX_LEN in set_mlucas_path()\n");
+		has_err = TRUE;
+		goto out_err_check;
+	}
+	if (MLUCAS_PATH[len - 1] != '/' && MLUCAS_PATH[len - 1] != '\\') {
+		fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH does not end with a slash in set_mlucas_path()\n");
+		has_err = TRUE;
+		goto out_err_check;
 	}
 #else
 	/* POSIX: run MLUCAS_PATH through the shell (via popen "printf") so $HOME-style
 	   variables get expanded, then enforce the length/trailing-slash invariants: */
-	{
-		char *cmdstr;
-		char *expanded_str;
-		int  tmp;
-		FILE *pipe_ptr;
-		size_t bufsize = strlen(MLUCAS_PATH) * 3 + 1;	/* quote_spaces() can triple each char, plus NUL */
-		mlucas_path = (char*)malloc(bufsize);
-		if (mlucas_path == NULL) {
-			fprintf(stderr, "ERROR: unable to allocate buffer mlucas_path in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_err_check;
-		}
-
-		quote_spaces(mlucas_path, MLUCAS_PATH);
-		cmdstr = (char*)malloc(bufsize + strlen("printf \"\""));
-		if (cmdstr == NULL) {
-			fprintf(stderr, "ERROR: unable to allocate buffer cmdstr in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_mlucas_path;
-		}
-
-		strcpy(cmdstr, "printf \"\"");
-		strcat(cmdstr, mlucas_path);
-		pipe_ptr = popen(cmdstr, "r");
-		if (pipe_ptr == NULL) {
-			fprintf(stderr, "ERROR: unable to open pipe pipe_ptr in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_cmdstr;
-		}
-
-		tmp = getc(pipe_ptr); /* goto out_pipe if shell output nothing  */
-		if (tmp == EOF)
-			goto out_pipe;
-		else
-			ungetc(tmp, pipe_ptr);
-
-		expanded_str = (char*)malloc(STR_MAX_LEN + 1); /* do not free!  */
-		if (expanded_str == NULL) {
-			fprintf(stderr, "ERROR: unable to allocate buffer expanded_str in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_pipe;
-		}
-		if (fgets(expanded_str, STR_MAX_LEN + 1, pipe_ptr) == NULL) {
-			fprintf(stderr, "ERROR: couldn't get environment variable MLUCAS_PATH in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_pipe;
-		}
-		if (getc(pipe_ptr) != EOF) {
-			fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH is longer than STR_MAX_LEN in set_mlucas_path()\n");
-			has_err = TRUE;
-			goto out_pipe;
-		}
-		if (expanded_str[strlen(expanded_str) - 1] != '/') { /* strlen != 0  */
-			fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH does not end with a slash in set_mlucas_path()\n");
-			has_err = TRUE;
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -9371,6 +9371,46 @@ char *MLUCAS_PATH = "";
...
 {
+#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
+	/* On Windows, popen() runs cmd.exe, which has no 'printf' command, so the
+	   shell-expansion trick used in the POSIX branch below fails with
+	   "'printf' is not recognized as an internal or external command" (issue #50).
+	   Windows also has no Bourne-shell variables like $HOME to expand, so simply
+	   use the environment variable MLUCAS_PATH (or the compiled-in default) verbatim,
+	   while enforcing the same invariants as the POSIX branch: the path must be
+	   no longer than STR_MAX_LEN and must end with a slash (fwd- or backslash,
+	   since both are directory separators on Windows).  */
```
tdulcet: There is some duplicated code with the POSIX branch, such as the `getenv()` call and the `out_err_check` label. Could we update this to share more of the implementations, which might make it easier for us to maintain.

tdulcet (commented):
Thanks for your work on this. It looks like we might be able to slightly simplify it.

MarkRose:
Good call — hoisted the shared prologue (the `getenv("MLUCAS_PATH")` + malloc + strcpy) and the `out_err_check`/ASSERT epilogue out of the `#if`/`#else`, so only the platform-specific middle differs now (Windows: use the path verbatim with the length/trailing-slash checks; POSIX: shell-expand via `popen "printf"` then the same checks). Behavior is unchanged — verified on Linux that `MLUCAS_PATH='$HOME/.mlucas.d/'` still expands to the real home dir and the empty default still yields `""`.

MarkRose:
This is already addressed by the latest push (7ec2d85): I hoisted the shared `getenv()`/malloc/`strcpy` prologue and the `out_err_check`/`ASSERT` epilogue out of the `#if OS_TYPE_WINDOWS … #else` so only the genuinely platform-specific middle differs now. Happy to factor it further if you'd like.

src/util.c
```diff
@@ -9371,83 +9371,108 @@ char *MLUCAS_PATH = "";
...
+		if (MLUCAS_PATH[len - 1] != '/' && MLUCAS_PATH[len - 1] != '\\') {
+			fprintf(stderr, "ERROR: environment variable MLUCAS_PATH or cpp macro MLUCAS_DEFAULT_PATH does not end with a slash in set_mlucas_path()\n");
+			has_err = TRUE;
+			goto out_err_check;
+		}
 	}
+#else
+	/* POSIX: run MLUCAS_PATH through the shell (via popen "printf") so $HOME-style
+	   variables get expanded, then enforce the length/trailing-slash invariants: */
+	{
```
tdulcet: We should not need to put each of the Windows and POSIX blocks in braces. It indents most of the code, which makes the diff much larger than it needs to be, and thus harder to review.

tdulcet (changes requested):
(Requesting changes per https://github.com/primesearch/Mlucas/pull/106#discussion_r3610932405 above.)

Note that the `mkdir_p()` function should probably also be updated, as it calls the `mkdir` and `rm -f` commands, which do not exist on Windows.

MarkRose:
Both done.

**Braces removed** from each arm of the `#if`, and the arms de-indented. They only existed to scope a couple of declarations, which C99 doesn't require, and the indentation was most of the diff — `git diff -w` and the raw diff are now the same size.

**`mkdir_p()` is shell-free.** You were right that it needed the same treatment, and it had the problem in triplicate: `mkdir ... 2> /dev/null` per component via `system()`, a `popen("printf ...")` to undo its own shell-quoting, and `rm -f` for the writability probe. cmd.exe has none of `printf`, `rm` or `/dev/null`, so it was unusable there for exactly the reason in #50.

It now walks the path through the C library — `_mkdir` on Windows, `mkdir(p,0777)` elsewhere — and keeps the same create-and-remove probe via `fopen()`/`remove()`. Not shell-quoting in the first place removes the need for the `printf` round-trip that existed only to undo it. Per-component `mkdir` errors stay ignored, since a component already existing is the common case on every startup after the first, so the probe still decides the return value and the contract is unchanged.

One deliberate behaviour change worth flagging: a failed `remove()` of the probe file no longer aborts. A leftover zero-length temp file is harmless and not worth killing a multi-week run over.

This does leave **`shell_quote()` with no callers**. It's declared in `util.h` so it doesn't warn, and removing an exported helper felt like scope creep here — happy to drop it if you'd prefer.

Verified on nosimd:

- fresh nested path created, and the already-exists case, both report success; no probe file left behind
- an unwritable parent still rejected, same error and abort
- `set_mlucas_path` unchanged on POSIX — `$HOME` expansion, the empty default, and the missing-trailing-slash error all match a pre-PR `main` build exactly

The Windows halves of both functions still can't be exercised here; those need the Windows CI. Also rebased onto current `main`, which was 200 commits ahead.

src/util.c
```diff
@@ -9477,6 +9498,17 @@ char *quote_spaces(char *dest, char *src)
 	return dest;
 }
 
+/* MinGW's mkdir() takes no mode argument, and MSVC spells it _mkdir(); both live in <direct.h>: */
+#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
+	#include <direct.h>
+	#define MKDIR_P_ONE(p)	_mkdir(p)
```
tdulcet: Maybe call the define just `MKDIR`, as it does not behave like the `mkdir -p` Unix command, which would create all of the parent directories at once.

src/util.c
```diff
@@ -9485,61 +9517,46 @@ char *quote_spaces(char *dest, char *src)
...
+	'rm -f' to clear the probe file - none of which exist under cmd.exe, so the whole function
+	was unusable on Windows for the same reason as issue #50. Going through the C library
+	instead is both portable and cheaper, and removes the shell-quoting round-trip entirely.
+	mkdir() failing because a component already exists is the common case on every startup
+	after the first, so per-component errors are ignored here; the writability probe below is
+	what actually determines the return value.  */
+	for (p = tmp + 1; *p != '\0'; ++p) {
+		if (*p == '/' || *p == '\\') {
+			char sep = *p;	*p = '\0';
+			(void)MKDIR_P_ONE(tmp);
```
tdulcet: Should we check this return value and return early if error is something other than the file already exists?

tdulcet (commented):
Thanks, I just have a question and a nit.

MarkRose:
Both addressed, and the second one turned out to have a real answer rather than a judgement call.

## The macro name

Renamed to `MKDIR`. You're right that `MKDIR_P_ONE` read as a one-shot `mkdir -p`; it creates a single component.

## Checking the return value

I looked into this rather than picking, and the answer is that returning early on anything other than "already exists" would be a bug — on Windows specifically.

Cross-compiling a probe with MinGW and running it under Wine:

```
_mkdir("C:")       -> -1  errno=17  EEXIST
_mkdir("C:\")      -> -1  errno=17  EEXIST
_mkdir("C:\Users") -> -1  errno=17  EEXIST
_mkdir("\\srv")    -> -1  errno=2   ENOENT     <-- not EEXIST
```

Drive-letter prefixes are fine, but a UNC path's leading `\\server` component reports **ENOENT**. Since the loop splits on `\` as well as `/`, a path like `\\server\share\Mlucas` hits `_mkdir("\\server")` on the way through — and an early return there would reject a path that is perfectly creatable.

On POSIX the question does not arise. Walking the components of real paths, intermediates only ever return success or `EEXIST`; a genuine failure surfaces at the leaf:

```
fresh nested path      ancestors EEXIST, new components 0, leaf 0
unwritable ancestor,
  leaf already exists  all EEXIST                       -> probe succeeds, returns 0 (correct)
unwritable ancestor,
  leaf does not exist  all EEXIST, leaf EACCES          -> probe fails, returns 1 (correct)
```

So on POSIX the check would be outcome-equivalent — the writability probe already reaches the same verdict — and on Windows it would be actively harmful. I've left the behaviour as-is and written the reasoning into the comment, since "why is this return value discarded" is a fair thing to wonder on the next read.

The caveat worth stating: those `_mkdir` results are Wine's msvcrt, not a real Windows run. Drive letters and UNC prefixes could in principle differ there. If you would rather have the check anyway, the safe form is to ignore `EEXIST` *and* `ENOENT` on non-leaf components and only act on the rest — happy to do that instead.

## Verification

- `util.c` cross-compiles clean with `x86_64-w64-mingw32-gcc`
- `MLUCAS_PATH=.../a/b/c` creates all three levels, and the probe file is removed afterwards
- an unwritable target still fails properly: nothing created, and `ERROR: mkdir -p "..." failed` from `print_host_info`, rc=134
- rebased onto current `main` (it was 62 behind); `-s tt` rc=0 with 23 residues

Still outstanding from your earlier note, and not done here: `mkdir_p()` was the `mkdir`/`rm -f` half of that comment, which this branch already replaced with C-library calls. If you meant other shell-outs elsewhere, point me at them and I'll take those too.

PR #107, opened 2026-07-16
mi64_div_by_scalar64: exact integer divide for len==1, dropping the FP round-off nudge (fixes #94)

### LINKED ISSUE #94: Improve workaround for PRP modular division bug (#18)
In #23, we implemented a temporary workaround for #18 (the PRP modular division bug), which may not fully resolve the issue: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/mi64.c#L6906-L6909 We should now see if we can determine the root cause and implement a more robust long term fix.

MarkRose:
Fixes #94.

Root-causes and replaces the fragile FP workaround from PR #23 (issue #18) in `mi64_div_by_scalar64()`.

**Root cause.** The FP round-off only ever affected the `len == 1` scalar-tail branch. Every multi-word path is already exact integer arithmetic (Algo A/B via `radix_power64`/`MONT_MUL64`, and the u2/u4 folded routines the real PRP base-division reaches) — verified by unit test on a full-size input. In the `len == 1` branch the *remainder* was already forced exact by a `rem64 = tmp%q` fallback (which predates #23), but the FP-derived *quotient* returned in `y[0]` was **not** corrected. PR #23's nudge only repairs a quotient short by exactly one unit landing in `[0.99999999999999, 1.0)`; when q is small and x large the FP quotient can be short by more than one unit, so the nudge can't recover it — the base-7 failure in #18.

**Fix.** Replace the whole FP block with a single hardware integer divide:
```c
if(len == 1) {
    rem64 = tmp % q;        // exact remainder
    if(y) y[0] = tmp / q;   // exact quotient (compiler fuses / and % into one divide)
    if(nshift) rem64 = (rem64 << nshift) + rem_save;
    return rem64;
}
```
`tmp/q` and `tmp%q` are exact by definition — no round-off. Not slower (the old code already issued a `tmp%q` divide for its self-check; this branch runs once per call, never in a hot loop). The nshift even-modulus restore is unchanged and its correctness is argued in an in-code comment.

**Verification (subagent, opus).** GMP-exact reference over 118M+ calls (mixed lengths 1–4, quotient+remainder, exact-multiple/near-multiple/random inputs, bases 3/7/9/25/27/28/29/49, 32-bit/full-64-bit/even moduli) — **0 failures**. The old code was independently shown to return a wrong quotient for exact multiples of small moduli (base 7 q=49: ~59%), reproducing #18. Deterministic single-length FFT self-tests (`-fftlen 192/512/1024 -iters 100`) give the known-good Res64 — no modmul-core regression. Only the `len == 1` path changed; multi-word paths are untouched.

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -6805,7 +6805,6 @@ uint64 mi64_div_by_scalar64(const uint64 x[], uint64 q, uint32 len, uint64 y[])
 #endif
 	uint32 i,nshift,lshift = -1,ptr_incr;
 	uint64 qinv,tmp = 0,bw,cy,lo,rem64,rem_save = 0,itmp64,mask,*iptr;
-	double fquo,fqinv;
 /* Debug:
 printf("x[]/q, quotient q = %" PRIu64 ", base b = 2^64\n",q);
 for(i = 0; i < len; i++)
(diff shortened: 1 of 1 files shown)


tdulcet (commented):
Thanks, this looks like a major improvement.

> when q is small and x large the FP quotient can be short by more than one unit, so the nudge can't recover it

For our release notes, do you know of any specific types of inputs (such as PRP base and exponent combinations) where the previous workaround would have produced a wrong result?

MarkRose:
The old workaround failed when the floating-point quotient q̂ = x/d came out low by **≥2** (not just 1), which the single ±1 correction step could not recover. That needs a small divisor `d` against a ~2⁶⁴ dividend with the rounding landing more than 1 ULP low. In practice the `len==1` short-divide is hit in the PRP/PRP-CF final "mod-divide the type-3 residue by base²" step (e.g. base 3 → divide by 9) and in Suyama-style cofactor reductions. I don't have a specific confirmed-wrong (base, exponent) on record — the fix came from the round-off analysis rather than a field report — but if it's useful for the notes I can try to construct a concrete failing input.

PR #108, opened 2026-07-16
Re-enable savefile write on interrupt, made async-signal-safe (fixes #100)

### LINKED ISSUE #100: Fix and reenable savefile write on interrupt
Mlucas used to write a savefile when it received a signal (SIGINT, SIGTERM, etc.) before exiting. However, Ernst disabled this in his December 2021 patch to v20.1.1:
> "Due to a user report of bad behavior of regular (non '-9') kill with his multithreaded run, signal handling has been changed to immediate-exit without savefile write."

https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L245-L246 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L2039-L2042

Depending on the FFT length and speed of the system, this can result in minutes to hours of lost work. We should fix the underlining bug(s) and reenable the savefile write on interrupt feature. https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/help.txt#L607-L611

MarkRose:
Fixes #100.

Re-enables writing a savefile on SIGINT/SIGTERM (etc.) before exiting, which Ernst disabled in Dec 2021 due to multithreaded runs that would "refuse to quit" and run-to-run inconsistency.

**Root cause of the old instability.** The old `sig_handler` was not async-signal-safe: it called `fprintf`/`sprintf` (into the shared global `cbuf`) and `exit()` from signal context. These signals were unblocked on every thread, so the kernel could deliver the signal to an FFT worker thread — the threads doing nearly all the CPU work and thus most likely to already hold the malloc-arena or a stdio lock. The handler would then deadlock on that non-recursive lock → the process hung forever; because *which* thread got the signal is nondeterministic, the hang was intermittent.

**The fix (four parts):**
1. **The handler now only sets two `volatile sig_atomic_t` flags** (signal number + `MLUCAS_KEEP_RUNNING = 0`) — the only data a handler may safely touch. No stdio/malloc/exit, so it cannot deadlock regardless of which thread/instruction it interrupts.
2. **FFT worker threads block these signals** (`pthread_sigmask` in `worker_thr_routine()`, `src/threadpool.c`), so the handler only ever runs on the main thread, keeping the lock-holding workers out of signal context and making delivery deterministic. This part is POSIX-only: the block is compiled out under `#if !defined(NO_USE_SIGNALS) && !defined(__MINGW32__)`, since MinGW/Windows has neither `sigset_t`/`pthread_sigmask` nor that signal-delivery model. Part (1), the safe handler, still applies there.
3. **All messaging and the savefile write are deferred to the main thread**, which polls `MLUCAS_KEEP_RUNNING` between mod-squaring iterations (prompt quit; the residue at that point is the exact last-completed-iteration value). The interim `exit(1)` in the `ERR_INTERRUPT` block is replaced by a fall-through into the existing checkpoint-write path, which writes a consistent savefile and then `exit(0)`s. Handlers are installed once via `sigaction` (SA_RESTART; `sa_mask` blocks the other quit-signals), with a `signal()` fallback on MinGW.
4. **A distinct interrupt-loss defect in the Gerbicz-check path.** The per-worktype interrupt testing asked for in review turned up a separate bug: in `ernstMain()`, on interrupt past the first Gerbicz subinterval, the code re-squares `b[]` once to make it savefile-consistent — but assigned that call's return value back to `ierr`, overwriting `ERR_INTERRUPT` with 0, so the downstream `if(ierr == ERR_INTERRUPT) { write savefile; exit(0); }` never fired. PRP and (same code path) Pépin therefore spun on to `maxiter` and aborted in the post-test residue step, `After short-div, R != 0 (mod B)`. LL and P-1 stage 1, which run no Gerbicz check, were unaffected. Fixed by discarding the `b[]`-undo's return so `ierr` stays `ERR_INTERRUPT`. This line is pre-existing on `main` and is not fallout from the signal-safety rework; it is folded in here because it is what makes parts (1)–(3) actually work for PRP/Pépin.

p-1 stage 2 keeps its prior behavior (resumes from its own interior stage-2 checkpoints); only its stale-buffer message was fixed. Extending graceful savefile-on-interrupt into stage 2 would be a separate, larger change.

**Verification** (current head, four commits).

x86-64, native `avx2` build — SIGINT/SIGTERM mid-run, per worktype:

| worktype | interrupt → savefile → `exit(0)` | resume |
| --- | --- | --- |
| LL | yes (e.g. Iter 2413) | yes; forward progress to Iter 11177 |
| PRP | yes | yes — requires part (4); without it PRP aborted instead |
| P-1 stage 1 and stage 2 | yes | yes |
| Pépin (F20) | yes (Iter 72971) | yes, but only with the separate fix noted below |

ARM (aarch64): the `asimd` target cross-built with gcc-11 and run under user-mode `qemu-aarch64` — LL interrupt → savefile → clean exit → resume with forward progress, i.e. the old "runs on and won't stop" behaviour is gone. This exercises the ARM-compiled signal path and the savefile/resume logic but not real-ARM scheduler timing, so a confirmation on physical ARM hardware is still worth having.

Across all trials interrupts exited promptly (~0.1s) with zero hangs, and the build is not slower or deadlock-prone relative to the same runs on a non-#100 build.

*Correction to an earlier revision of this description*, which reported a 4-thread NOSIMD run as bit-exact on interrupt+resume for both PRP+Gerbicz and non-Gerbicz LL: that predates part (4), so PRP could not have written a savefile on interrupt at that point and only the LL leg of that claim stands. The table above supersedes it.

**Surfaced here but fixed elsewhere:** Pépin + Gerbicz could not resume at all — a from-scratch run seeds a random residue shift, and the resume path asserted `RES_SHIFT == 0`. That is pre-existing on `main` (it reproduces with a plain `kill -9` on a natural checkpoint) and Fermat-Pépin-only; it is fixed in #218, which makes the Fermat-mod Gerbicz check work with a nonzero shift.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -2116,9 +2116,9 @@ READ_RESTART_FILE:
 				case SIGUSR1: signame = "SIGUSR1"; break;
 				case SIGUSR2: signame = "SIGUSR2"; break;
 			#endif
-				default:      signame = "quit";    break;
+				default:      signame = "unknown"; break;
 			}
-			snprintf(cbuf,STR_MAX_LEN*2,"Received %s signal: writing savefile at Iter = %u and exiting.\n",signame,ihi);
+			snprintf(cbuf,sizeof(cbuf),"Received %s signal: writing savefile at Iter = %u and exiting.\n",signame,ihi);
 			mlucas_fprint(cbuf,1);
 			// Fall through to the checkpoint-write path below.
 		}
@@ -2757,7 +2757,7 @@ PM1_STAGE2:	// Stage 2 invocation is several hundred lines below, but this needs
 					// p-1 stage 2 does its own interior checkpointing (see pm1.c); on interrupt we resume
 					// from the last stage-2 checkpoint, so just report and exit. The async handler only
 					// recorded the signal number, so build the message here on the main thread:
-					snprintf(cbuf,STR_MAX_LEN*2,"Received quit signal (%d) in p-1 stage 2: exiting; will resume from last stage-2 checkpoint.\n",(int)MLUCAS_INTERRUPT_SIGNO);
+					snprintf(cbuf,sizeof(cbuf),"Received quit signal (%d) in p-1 stage 2: exiting; will resume from last stage-2 checkpoint.\n",(int)MLUCAS_INTERRUPT_SIGNO);
 					mlucas_fprint(cbuf,1);
 					exit(1);
 				} else if(ierr & (1<<ERR_ROUNDOFF)) {	// One or more modmuls hit a roundoff error - bump FFT length and restart
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -58,8 +58,11 @@ uint32 SYSTEM_RAM = 0;	// Total usable main memory size in MB, and max. % of tha
 
 // Used to force local-data-tables-reinits in cases of suspected table-data corruption:
 int REINIT_LOCAL_DATA_TABLES = 0;
-// Normally = True; set = False on quit-signal-received to allow desired code sections to and take appropriate action:
-int MLUCAS_KEEP_RUNNING = 1;
+// Normally = True; set = False on quit-signal-received to allow desired code sections to and take appropriate action.
+// volatile sig_atomic_t: written by the async signal handler, polled by the main control loop (see Mdata.h note).
+volatile sig_atomic_t MLUCAS_KEEP_RUNNING = 1;
+// Signal number of a received graceful-quit signal, so the main thread can report it safely (0 = none):
+volatile sig_atomic_t MLUCAS_INTERRUPT_SIGNO = 0;
 // v18: Enable savefile-on-interrupt-signal, access to argc/argv outside main():
 char **global_argv;
 
@@ -225,27 +228,83 @@ uint64 PMAX;		/* maximum exponent allowed depends on max. FFT length allowed
 /****** END(Allocate storage for Globals (externs)). ******/
 
 #ifndef NO_USE_SIGNALS
+	/*
+	Async-signal-safe graceful-quit handler.
+
+	ROOT CAUSE of the Dec-2021 "runs that have been underway for a day or more refuse to quit"
+	instability (which led to this savefile-on-interrupt feature being disabled): a signal handler
+	runs on whatever thread the kernel happens to deliver the signal to, asynchronously interrupting
+	whatever that thread was doing at that instant. POSIX permits a handler to call ONLY async-signal-safe
+	functions (see signal-safety(7)). The old handler violated this badly - it called fprintf() and
+	sprintf() (into the shared global cbuf) and then exit(). All three are unsafe:
+	  - fprintf/sprintf take the stdio lock and can call malloc();
+	  - exit() flushes all stdio streams and runs atexit() handlers.
+	If the signal was delivered to an FFT worker thread (all threads had these signals unblocked) that
+	happened to already hold the malloc arena lock or a stdio lock - very likely, since the workers are
+	the threads doing nearly all the CPU work - the handler would deadlock on that non-recursive lock and
+	the entire process would hang forever. Because *which* thread receives the signal is nondeterministic,
+	the hang was intermittent: exactly the "run-to-run inconsistency" that was reported. sprintf() into the
+	shared cbuf also races the main thread's concurrent use of that same buffer.
+
+	The fix has three parts:
+	  (1) This handler now does NOTHING but store the signal number and clear the keep-running flag - both
+	      volatile sig_atomic_t, the only data the C standard lets a handler safely touch. No stdio, no
+	      malloc, no exit(): nothing that can take a lock, so it cannot deadlock no matter which thread or
+	      instruction it interrupts.
+	  (2) All user messaging and the (consistent, last-completed-iteration) savefile write are deferred to
+	      the main-thread control loop, which polls MLUCAS_KEEP_RUNNING at a safe point between mod-squaring
+	      iterations.
+	  (3) The FFT worker threads block these signals (see threadpool.c::worker_thr_routine), so the handler
+	      only ever runs on the main thread - making delivery deterministic and keeping the workers, which
+	      hold the shared locks, out of signal context entirely.
+	*/
 	void sig_handler(int signo)
 	{
-		if (signo == SIGINT) {
-			fprintf(stderr,"received SIGINT signal.\n");	sprintf(cbuf,"received SIGINT signal.\n");
-		} else if(signo == SIGTERM) {
-			fprintf(stderr,"received SIGTERM signal.\n");	sprintf(cbuf,"received SIGTERM signal.\n");
-	#ifndef __MINGW32__
-		} else if(signo == SIGHUP) {
-			fprintf(stderr,"received SIGHUP signal.\n");	sprintf(cbuf,"received SIGHUP signal.\n");
-		} else if(signo == SIGALRM) {
-			fprintf(stderr,"received SIGALRM signal.\n");	sprintf(cbuf,"received SIGALRM signal.\n");
-		} else if(signo == SIGUSR1) {
-			fprintf(stderr,"received SIGUSR1 signal.\n");	sprintf(cbuf,"received SIGUSR1 signal.\n");
-		} else if(signo == SIGUSR2) {
-			fprintf(stderr,"received SIGUSR2 signal.\n");	sprintf(cbuf,"received SIGUSR2 signal.\n");
-	#endif
-		}
-	// Dec 2021: Until resolve run-to-run inconsistencies in signal handling, kill it with fire:
-	exit(1);
-		// Toggle a global to allow desired code sections to detect signal-received and take appropriate action:
-		MLUCAS_KEEP_RUNNING = 0;
+		MLUCAS_INTERRUPT_SIGNO = signo;	// Record which signal, for the main thread to report safely later
+		MLUCAS_KEEP_RUNNING = 0;		// The main control loop polls this between iterations and quits gracefully
+	}
+
+	/*
+	Install sig_handler for the graceful-quit signals. Called from the mod-squaring functions; the static
+	'installed' guard makes it a no-op after the first call, so it runs exactly once, on the main thread.
+
+	We use sigaction() rather than signal() for deterministic, portable semantics:
+	  - sa_mask blocks the other quit-signals while the handler runs, so two signals racing in can't
+	    interleave the (trivial) flag stores;
+	  - SA_RESTART makes interrupted library calls auto-resume instead of failing with EINTR;
+	  - the handler stays installed after firing (signal() gives this on glibc but not on strict SysV).
+	Because the FFT worker threads block these signals (threadpool.c), the handler set here only ever
+	runs on the main thread.
+	*/
+	void mlucas_install_signal_handlers(void)
+	{
+		static int installed = 0;
+		if(installed) return;
+		installed = 1;
+
+	#ifdef __MINGW32__
+		// Windows/MinGW has no sigaction(); fall back to signal() for the signals it supports:
+		if(signal(SIGINT , sig_handler) == SIG_ERR) fprintf(stderr,"Can't catch SIGINT.\n");
+		if(signal(SIGTERM, sig_handler) == SIG_ERR) fprintf(stderr,"Can't catch SIGTERM.\n");
+	#else
+		struct sigaction sa;
+		memset(&sa, 0, sizeof sa);
+		sa.sa_handler = sig_handler;
+		sa.sa_flags = SA_RESTART;
+		sigemptyset(&sa.sa_mask);
+		sigaddset(&sa.sa_mask, SIGINT);
+		sigaddset(&sa.sa_mask, SIGTERM);
+		sigaddset(&sa.sa_mask, SIGHUP);
+		sigaddset(&sa.sa_mask, SIGALRM);
+		sigaddset(&sa.sa_mask, SIGUSR1);
+		sigaddset(&sa.sa_mask, SIGUSR2);
+		if(sigaction(SIGINT , &sa, 0x0)) fprintf(stderr,"Can't catch SIGINT.\n");
+		if(sigaction(SIGTERM, &sa, 0x0)) fprintf(stderr,"Can't catch SIGTERM.\n");
+		if(sigaction(SIGHUP , &sa, 0x0)) fprintf(stderr,"Can't catch SIGHUP.\n");
+		if(sigaction(SIGALRM, &sa, 0x0)) fprintf(stderr,"Can't catch SIGALRM.\n");
+		if(sigaction(SIGUSR1, &sa, 0x0)) fprintf(stderr,"Can't catch SIGUSR1.\n");
+		if(sigaction(SIGUSR2, &sa, 0x0)) fprintf(stderr,"Can't catch SIGUSR2.\n");
+	#endif	// __MINGW32__ ? signal() : sigaction()
 	}
 #endif
 
@@ -2038,15 +2097,30 @@ READ_RESTART_FILE:
 		if(INTERACT && (ierr == ERR_INTERRUPT))
 			exit(0);
 
-		// In non-interactive (production-run) mode, write savefiles and exit gracefully on signal:
+		// In non-interactive (production-run) mode, write a consistent savefile and exit gracefully on signal.
+		// This block runs on the main thread at a safe point: the inner mod-squaring loop exited cleanly
+		// *between* squarings (it polls MLUCAS_KEEP_RUNNING every iteration), so the residue just converted
+		// into arrtmp[] above is the exact last-completed-iteration residue - nothing is torn or in-flight.
+		// We do NOT exit here: we fall through to the normal checkpoint-write code below (which then does
+		// 'if(ierr == ERR_INTERRUPT) exit(0);' after the savefile has been safely written and fclosed).
 		if(ierr == ERR_INTERRUPT) {
-			// First print the signal-handler-generated message:
+			ihi = ROE_ITER;	// Last-iteration-completed-before-interrupt saved here; savefile is written for this iter
+			// Report the signal safely here on the main thread (the async handler only recorded the number):
+			const char *signame;
+			switch(MLUCAS_INTERRUPT_SIGNO) {
+				case SIGINT:  signame = "SIGINT";  break;
+				case SIGTERM: signame = "SIGTERM"; break;
+			#ifndef __MINGW32__
+				case SIGHUP:  signame = "SIGHUP";  break;
+				case SIGALRM: signame = "SIGALRM"; break;
+				case SIGUSR1: signame = "SIGUSR1"; break;
+				case SIGUSR2: signame = "SIGUSR2"; break;
+			#endif
+				default:      signame = "quit";    break;
+			}
+			snprintf(cbuf,STR_MAX_LEN*2,"Received %s signal: writing savefile at Iter = %u and exiting.\n",signame,ihi);
 			mlucas_fprint(cbuf,1);
-			ihi = ROE_ITER;	// Last-iteration-completed-before-interrupt saved here
-		/*** Nov 2021: interrupt-handling still not stable ... runs that have been underway for a day or more refuse to quit. Just clean-exit w/o savefile write for now: ***/
-		//	sprintf(cbuf,"Iter = %u: Writing savefiles and exiting.\n",ihi);
-			sprintf(cbuf,"Exiting at Iter = %u.\n",ihi); mlucas_fprint(cbuf,1);
-			exit(1);
+			// Fall through to the checkpoint-write path below.
 		}
 
 		/*...Done?	*/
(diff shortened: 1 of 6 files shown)


src/Mlucas.c
```diff
@@ -2031,15 +2090,30 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+				case SIGTERM: signame = "SIGTERM"; break;
+			#ifndef __MINGW32__
+				case SIGHUP:  signame = "SIGHUP";  break;
+				case SIGALRM: signame = "SIGALRM"; break;
+				case SIGUSR1: signame = "SIGUSR1"; break;
+				case SIGUSR2: signame = "SIGUSR2"; break;
+			#endif
+				default:      signame = "quit";    break;
+			}
+			snprintf(cbuf,STR_MAX_LEN*2,"Received %s signal: writing savefile at Iter = %u and exiting.\n",signame,ihi);
```
tdulcet: We should used `sizeof` here, as with #92. Also in the occurrence below.
```suggestion
			snprintf(cbuf,sizeof(cbuf),"Received %s signal: writing savefile at Iter = %u and exiting.\n",signame,ihi);
```

src/Mlucas.c
```diff
@@ -2031,15 +2090,30 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+			switch(MLUCAS_INTERRUPT_SIGNO) {
+				case SIGINT:  signame = "SIGINT";  break;
+				case SIGTERM: signame = "SIGTERM"; break;
+			#ifndef __MINGW32__
+				case SIGHUP:  signame = "SIGHUP";  break;
+				case SIGALRM: signame = "SIGALRM"; break;
+				case SIGUSR1: signame = "SIGUSR1"; break;
+				case SIGUSR2: signame = "SIGUSR2"; break;
+			#endif
+				default:      signame = "quit";    break;
```
tdulcet: Maybe `unknown` instead of `quit`, as SIGQUIT is a signal.

tdulcet (changes requested):
Thanks for fixing this longstanding issue.

> Reviewer/CI runtime validation on a normal setup is welcome given this is robustness-critical.

This has previously been disabled and re-enabled by Ernst several times after bugs have been reported. For example, from the v20 release notes:
> "I'm getting inconsistencies between platforms in the signal handling, until those are resolved 'kill -9' is the safe way to go."

There was also an issue of Mlucas running way too fast on ARM after attempting to stop it (see this [forum post](https://www.mersenneforum.org/node/19128?p=769693#post769693), which I have experienced as well.

Ken on the forum [previously reported these issues](https://www.mersenneforum.org/node/17744?p=797398#post797398) (see the v20.1.1 section, number 5):

> Attempts to stop a running Mlucas session, such as to restart from beginning, or to change the mlucas.ini file and have the changes take effect, may produce assorted .stat file anomalies, depending on the method of stopping the session attempted. top kill 9 works. Ctrl-C creates anomalies. (There is a patch file available which is thought to address most cases (but possibly not P-1 stage 2). And the 2021-11-23 patch likely incorporates that if not more, as a result of restored signal handling.) Following is a list of some of the interesting features of the resulting stat file. These are mostly also typically signs of a computation gone wrong.
> 
> 1. Repeating res64 with differing iteration count during Ctrl-C as kill attempts.
> 2. Clocks anomalies vs. elapsed time indicated by time stamp. For example, 100k iter to 110k iter, 13 seconds, clocks value unchanged, ms/iter 10-fold increase.
> 3. Unchanged shift count between status lines indicated as 10k iterations apart, in a run using nonzero shift.
> 4. Error measures zero in some entries.
> 5. Exact same clocks values at 10k iters apart in early running.
> 6. Every log interval Mlucas claims to be restarting, including during the last run in which it was being left to run undisturbed while I slept.
> 7. Mismatch of res64 for same iteration count and exponent and computation type, between the affected run and independent runs' log output. After a thorough restart from scratch process, the 10k and 100k iter res64s match the corresponding res64s from a previous gpuowl run. I suspect the erroneous res64 in the affected Mlucas run is a res64 carried over from some earlier iteration count.

To increase our confidence that these issues are now fully resolved, it would be very helpful if you perform a brief test with each worktype (LL, Pépin, PRP, Suyama, P-1 stage 1, P-1 stage 2) on both x86 and ARM, to confirm that the savefile write and exit on interrupt now works as expected in all cases. This cannot easily be validated by the CI, so it would need some additional manual testing.

MarkRose:
Addressed the inline suggestions (use `sizeof(cbuf)` in the two interrupt `snprintf`s; default signal-name `quit` → `unknown`).

And thank you for pushing on the per-worktype interrupt testing — **it found a real bug.** On x86 I ran SIGINT/SIGTERM mid-run for each worktype and discovered that **PRP (and, by the same code path, Pépin) did *not* write-and-exit on interrupt** — the run instead spun on to `maxiter` and aborted in the post-test residue step (`After short-div, R != 0 (mod B)`). LL and P-1 stage 1 were fine.

Cause: on interrupt past the first Gerbicz subinterval, the code re-squares `b[]` once to make it savefile-consistent, but assigned that call's return **back to `ierr`**, overwriting `ERR_INTERRUPT` with `0` — so the `if(ierr == ERR_INTERRUPT) { write savefile; exit(0); }` handler never fired. Fixed by discarding the `b[]`-undo return so `ierr` stays `ERR_INTERRUPT` (latest commit).

**Re-verified on x86 (avx2)** after the fix:

| worktype | interrupt | result |
|---|---|---|
| LL (`Test=`) | SIGINT | savefile written at interrupted iter, `exit(0)`; restart resumes |
| PRP (`PRP=1,2,p,-1`) | SIGTERM | now clean (was aborting); interrupted @16626, restart resumed from there |
| P-1 stage 1 (`Pminus1=…`) | SIGINT | savefile written, `exit(0)` |

The fix is in the shared Gerbicz-check subinterval loop, so **Pépin** (Fermat, also GEC) is covered by the same change. I have not yet exercised **P-1 stage 2** (separate interior-checkpoint path in `pm1.c`) or run any of this on **ARM** hardware, which I don't have handy — if you can run the ARM leg and a P-1 S2 interrupt, that would complete the matrix you outlined.

MarkRose:
The two inline items are addressed in b654d99: `snprintf(cbuf,sizeof(cbuf),…)` at both interrupt sites (matching #92), and the `signame` default is now `"unknown"` rather than `"quit"`.

On the manual validation: I've exercised interrupt → savefile-write → resume on **x86** for **PRP** and **P-1 (stage 1 and stage 2)** and it behaves as expected (in particular the PRP/Pépin Gerbicz path, where a separate bug was dropping the interrupt — fixed here in 6f11ff6). What I have **not** yet done is the **LL/Pépin** cases or **any ARM** testing (no ARM hardware here). Given the history you cite, the ARM check is clearly important, so I'd flag that as still-required before merge — I can attempt it under emulation, but real ARM hardware would be more trustworthy for signal-timing behavior.

MarkRose:
I've now done the interrupt validation you asked for — savefile-write-on-interrupt and resume across worktypes, on both x86 and ARM.

**x86-64** (native, avx2): interrupt (SIGTERM) → savefile → clean exit → resume, all confirmed:

| Worktype | Interrupt → savefile → exit | Resume |
| --- | --- | --- |
| LL | ✓ (e.g. Iter 2413, exit 0) | ✓ resumes, forward progress to Iter 11177 |
| PRP | ✓ | ✓ (incl. the Gerbicz-path interrupt-lost fix, 6f11ff6) |
| P-1 stage 1 & 2 | ✓ | ✓ |
| Pépin (F20) | ✓ (Iter 72971, exit 0) | ✓ *after* the separate fix below |

**ARM (aarch64)**: cross-built the `asimd` target with gcc-11 and ran it under `qemu-aarch64` (user-mode). LL interrupt→resume works there too:

```
Received SIGTERM signal: writing savefile at Iter = 1841 and exiting.
INFO: restart file p216091 found...reading...
Received SIGTERM signal: writing savefile at Iter = 5294 and exiting.   <- resumed, forward progress
```

So on ARM the signal is caught, the savefile is written, the process exits cleanly, and the resume continues from the checkpoint — i.e. the "runs on and won't stop" behavior from the old handler is gone. Caveat: this is qemu emulation, so it exercises the ARM-compiled signal path and the savefile/resume logic but not real-ARM scheduler *timing*; a confirmation on physical ARM hardware would still be worthwhile if anyone has a box handy.

**One thing this surfaced (separate from this PR):** the Pépin (Fermat primality) + Gerbicz-check case couldn't resume at all — a from-scratch run seeds a random residue shift, but the restart path asserts `RES_SHIFT == 0`, so every resume aborted with *"Shifted residues unsupported for Pépin test with Gerbicz check!"*. That's pre-existing on `main` (it reproduces with a plain `kill -9` on a natural checkpoint, nothing to do with this PR's interrupt path) and Fermat-Pépin-only (Mersenne PRP is fine). I've fixed it separately in #122; with that applied, the Pépin row above passes.

src/threadpool.c
```diff
@@ -276,6 +276,29 @@ me at: heber.tomer@gmail.com
...
+			sigemptyset(&quit_sigs);
+			sigaddset(&quit_sigs, SIGINT);
+			sigaddset(&quit_sigs, SIGTERM);
+			sigaddset(&quit_sigs, SIGHUP);
+			sigaddset(&quit_sigs, SIGALRM);
+			sigaddset(&quit_sigs, SIGUSR1);
+			sigaddset(&quit_sigs, SIGUSR2);
+			pthread_sigmask(SIG_BLOCK, &quit_sigs, 0x0);
+		}
+	#endif
```
tdulcet: It would be good if we could reduce the number of POSIX vs Windows code differences (and thus the amount of further manual testing required), so with all of the other fixes in the PR, is this code block still needed? It also occurs above.

tdulcet (commented):
Thanks all for your help testing this.

MarkRose:
Removed. You're right that it should not be there, and the reasoning is that the rest of the PR made it redundant rather than that it was ever wrong.

The block existed because the *old* handler did the savefile write inline, so it could deadlock against a malloc/stdio lock held by whatever thread it interrupted — and the FFT workers, doing nearly all the CPU work, were the likely holders. This PR already removed that: the handler is now

```c
MLUCAS_INTERRUPT_SIGNO = signo;
MLUCAS_KEEP_RUNNING = 0;
```

two stores to `volatile sig_atomic_t` and nothing else. That is safe on any thread, so keeping the workers out of signal context stopped buying anything.

## Measured rather than assumed

Interrupting a `Test=1257787` run at points spread across startup and mid-run, checking for exit status 0, a written savefile, and how long the process took to go:

| | result |
| --- | --- |
| with the mask | 12/12 clean exits, slowest 1s |
| without the mask | 12/12 clean exits, slowest 1s |
| without, on the final branch | 15/15 clean exits, slowest 1s |

And the more direct question — does removing it actually change which thread handles the signal? Instrumenting the handler to record `syscall(SYS_gettid)`:

```
TESTPROBE: handler ran on tid 250961, main thread tid 250961
TESTPROBE: handler ran on tid 251113, main thread tid 251113
... 8 of 8 identical
```

With the block gone, Linux still delivered to the main thread every time. So the mask was not what made delivery deterministic here; the kernel's choice was.

## One thing that had to change with it

POSIX only guarantees a process-directed signal goes to *some* thread not blocking it. That is harmless given the handler is async-signal-safe, but it means the four comments asserting the handler "only ever runs on the main thread" stopped being true the moment the block came out — in `Mlucas.c` (twice), `mers_mod_square.c` and `fermat_mod_square.c`. I reworded them to state what actually holds: the handler is safe wherever it runs, and the main thread is where the flag gets acted on. Leaving those in place would have been worse than the block itself.

That leaves `__MINGW32__` guarding only the `signal()`-vs-`sigaction()` split, which MinGW forces on us.

## Verification

- rebased onto current `main` (it was 262 behind), builds clean, `-s tt` rc=0 with 23 residues
- `threadpool.c` still cross-compiles with `x86_64-w64-mingw32-gcc`
- the 15-run interrupt soak above, on the final branch

Happy to put it back if you would rather keep the belt-and-braces, but on this evidence it is cost without benefit.

Separately, your earlier request for manual testing across all six worktypes on x86 and ARM is still outstanding — I have not done that yet, and it is not covered by anything above.

src/Mlucas.c
```diff
@@ -225,27 +228,83 @@ uint64 PMAX;		/* maximum exponent allowed depends on max. FFT length allowed
...
+		sigaddset(&sa.sa_mask, SIGALRM);
+		sigaddset(&sa.sa_mask, SIGUSR1);
+		sigaddset(&sa.sa_mask, SIGUSR2);
+		if(sigaction(SIGINT , &sa, 0x0)) fprintf(stderr,"Can't catch SIGINT.\n");
+		if(sigaction(SIGTERM, &sa, 0x0)) fprintf(stderr,"Can't catch SIGTERM.\n");
+		if(sigaction(SIGHUP , &sa, 0x0)) fprintf(stderr,"Can't catch SIGHUP.\n");
+		if(sigaction(SIGALRM, &sa, 0x0)) fprintf(stderr,"Can't catch SIGALRM.\n");
+		if(sigaction(SIGUSR1, &sa, 0x0)) fprintf(stderr,"Can't catch SIGUSR1.\n");
+		if(sigaction(SIGUSR2, &sa, 0x0)) fprintf(stderr,"Can't catch SIGUSR2.\n");
+	#endif	// __MINGW32__ ? signal() : sigaction()
```
tdulcet: With the other code now removed, I do not believe `sigaction` is needed anymore either. All that would need to be guarded by `__MINGW32__` is the four signals (SIGHUP, SIGALRM, SIGUSR1 and SIGUSR2) that are unsupported on Windows, as they were before.

src/threadpool.c
```diff
@@ -285,6 +285,7 @@ me at: heber.tomer@gmail.com
 		thread_control_t *t = &init->control;
 		task_control_t *task;
 
+
```
tdulcet: Nit: Added blank line.

MarkRose:
Both done, and rebased onto current `main`.

## `sigaction()` → `signal()`

You were right that it was no longer earning its keep. It was there for two properties — `SA_RESTART`, and the handler staying installed after it fires — and I had written a comment claiming `signal()` gives those "on glibc but not on strict SysV". Rather than trust my own comment I measured it: installing each of the six signals with `signal()`, then reading the disposition back with `sigaction()`, on glibc 2.44:

```
SIGINT   installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
SIGTERM  installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
SIGHUP   installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
SIGALRM  installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
SIGUSR1  installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
SIGUSR2  installed=yes  SA_RESTART=yes  SA_RESETHAND(one-shot)=no
```

So both properties are already there, and in any case neither matters for a handler that is two stores to `volatile sig_atomic_t`. `sa_mask` was guarding a race between those two stores that has no consequence.

The installer is now plain `signal()` for all six, with `__MINGW32__` guarding only SIGHUP/SIGALRM/SIGUSR1/SIGUSR2 — the same guard `sig_handler` itself has always used. The only other `__MINGW32__` the PR adds is the matching one in the signal-name `switch`, for the same four `case` labels.

## The blank line

Fixed — and with it gone `threadpool.c` is byte-identical to `main`, so the file drops out of the PR entirely. The diff is now five files, all of them core, with no Windows-specific code path left in it beyond those four signal names.

## Verification

**Linux**, AVX2, 4 threads (`-cpu 0:3`), 10 independent trials, fresh state each time, alternating SIGINT/SIGTERM:

| | result |
| --- | --- |
| clean exit 0 + savefile written + iteration reported | 10/10 |
| slowest time from signal to exit | under 1 ms |

`-s tt` passes, rc=0, 23 residues.

**Windows**: since you have flagged the manual-testing cost of Windows-specific code before, I cross-built the whole thing with `x86_64-w64-mingw32-gcc` and ran it under Wine 11.16 rather than only compile-testing it. `-m 189997 -fft 10 -iters 100` gives `Res64: 5F1F741C1FBDA658`, and the interrupt path works there too:

```
Received SIGINT signal: writing savefile at Iter = 136660 and exiting.
```

exit 0 in ~1 ms, savefile written, and the next run resumed from it and carried the assignment through to completion, writing a normal result line:

```
{"status":"C", "exponent":189997, "worktype":"LL", "res64":"BDBBC8164B6E807D", ...}
```

## Two things I hit that are not this PR

Neither is a change here, but both are worth recording somewhere.

1. **`makemake.sh` adds `-lrt` when cross-compiling for Windows.** The guard is `if [[ $OSTYPE != msys && $OSTYPE != cygwin ]]`, which tests the *build host* rather than the target, so a MinGW cross-build from Linux gets `-lrt` and fails to link with `cannot find -lrt`. A native msys/cygwin build is unaffected, which is presumably why it has not come up. Probing the compiler (or `$CC -dumpmachine`) would match how the file already handles `-O` and the other flag checks. I linked by hand to get the Wine testing above done.

2. **An LL assignment that triggers an FFT-length escalation is silently dropped.** `Test=216091` with only small FFT lengths available starts at 10K (`p/pmax_rec = 0.9968`), hits roundoff, prints `Switching to next-larger available FFT length 11K and restarting from last checkpoint file`, and then exits 0 having consumed the `worktodo.txt` entry and written no result — no `results.txt`, no verdict. **This is pre-existing**: identical behaviour on `main` at 8728dc5 and on this branch, with no interrupts involved at all. I only tripped over it because I was testing with a deliberately small `mlucas.cfg`.

MarkRose:
Correction to point 2 of my previous comment: **there is no dropped-assignment bug.** I was wrong, and the claim should be disregarded.

The `Test=216091` run completed normally. The verdict is in the exponent logfile `p216091.stat`:

```
[...] M216091 Iter# = 216089 [100.00% complete] [...] Res64: 0000000000000000. [...]
M216091 is a known MERSENNE PRIME.
```

A zero final residue at 100% is the correct LL result, and 216091 is in the known-Mersenne-primes table, so the FFT-length escalation from 10K to 11K worked exactly as intended and the assignment was consumed because it was finished.

What misled me is that the `isprime` branch reports via `mlucas_fprint(cbuf, INTERACT || scrnFlag)` — logfile only in a production run — and writes no JSON line to `results.txt`, which only the composite branch does. I checked stderr and `results.txt`, saw neither, and concluded the result had been lost without looking at the `.stat` file. Identical behaviour on `main` and on this branch, as I said, but for the mundane reason that both are correct.

Point 1, the `-lrt` cross-build failure, stands unchanged.

src/Mlucas.c
```diff
@@ -1883,8 +1926,12 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 					/* If interrupt *and* we're past the first subinterval, need to undo initial-fwd-FFT-pass and DWT-weighting on b[],
 					whose value will reflect the last multiple-of-ITERS_BETWEEN_GCHECK_UPDATES iteration - prior to writing it,
 					along with the current PRP residue, to savefile: */
+					// This b[]-undo is only to make b[] savefile-consistent for the ensuing interrupt
+					// checkpoint-write; it must NOT clobber ierr, which has to stay ERR_INTERRUPT so the
+					// downstream savefile-write-and-exit handling fires (else the interrupt is lost and the
+					// PRP test spins to maxiter and aborts in the post-test residue step). Discard its return:
 					if(ierr == ERR_INTERRUPT && !first_sub)
-						ierr = func_mod_square  (b, (int*)arrtmp, n, i,i+1, 8ull, p, scrnFlag, &tdif2, FALSE, 0x0);
+						(void)func_mod_square  (b, (int*)arrtmp, n, i,i+1, 8ull, p, scrnFlag, &tdif2, FALSE, 0x0);
```
tdulcet: Should we still check this return value and output an error if the function failed?

src/Mlucas.c
```diff
@@ -225,27 +228,67 @@ uint64 PMAX;		/* maximum exponent allowed depends on max. FFT length allowed
...
+	      volatile sig_atomic_t, the only data the C standard lets a handler safely touch. No stdio, no
+	      malloc, no exit(): nothing that can take a lock, so it cannot deadlock no matter which thread or
+	      instruction it interrupts.
+	  (2) All user messaging and the (consistent, last-completed-iteration) savefile write are deferred to
+	      the main-thread control loop, which polls MLUCAS_KEEP_RUNNING at a safe point between mod-squaring
+	      iterations.
+	  (3) Because of (1) it does not matter which thread the kernel picks to run the handler: a
+	      process-directed signal goes to any thread not blocking it, and every one of them can safely
+	      execute these two stores. The main thread is the one that acts on the flag.
+	*/
```
tdulcet, line 244: Maybe we could reduce the verbosity of this comment, and the one for the function below.

tdulcet (commented):
Thanks, I just have a quick question I should have asked earlier.

MarkRose:
Both done.

## The discarded return value

You are right that it should not just vanish. It still must not land in `ierr` — that has to stay `ERR_INTERRUPT` or the savefile-write-and-exit handling never fires, which is the bug the fourth commit here exists to fix — but "must not clobber `ierr`" and "must not be looked at" are different things, and I conflated them.

It now goes into its own local and warns:

```c
if(ierr == ERR_INTERRUPT && !first_sub) {
    int ierr_undo = func_mod_square(b, ...);
    if(ierr_undo) {
        snprintf(cbuf,sizeof(cbuf),"WARNING: b[]-undo prior to the interrupt checkpoint returned error code[%u] = %s. ...");
        mlucas_fprint(cbuf,1);
    }
}
```

Worth reporting rather than ignoring because the failure mode is nasty: if that undo fails, the Gerbicz checkproduct written to the savefile no longer matches the residue beside it, and nothing notices until the *next* resume reports a Gerbicz-check failure — a long way from the cause, and easy to misread as a hardware problem. The message says so, and suggests restarting from an earlier savefile.

## Comment verbosity

Cut both down. The root cause is worth keeping on the record — it is why the feature was disabled for four years — but it was being retold three times over. The handler comment goes from 30 lines to 13, the installer's from 10 to 6, and I dropped the numbered "(1)(2)(3)" structure that was mostly restating what the two-line function body already shows.

Net for the commit: 40 lines removed, 27 added, and most of the additions are the new warning rather than prose.

## Verification

Rebuilt clean and re-ran the interrupt soak: **6/6** independent trials at 4 threads, alternating SIGINT/SIGTERM, each exiting 0 with a savefile written. `src/Mlucas.c` still cross-compiles with `x86_64-w64-mingw32-gcc`.

One aside that cost me a few minutes and may be worth knowing: rebuilding by deleting only `obj_avx2/Mlucas` and leaving the objects gave `lto1: internal compiler error: in cache_integer_cst, at tree.cc:2106`. It is not a code problem — a clean `rm -rf obj_avx2` builds fine — but LTO plus stale objects from an earlier compile is apparently enough to ICE gcc 16.

tdulcet (approved):

PR #110, opened 2026-07-16
Set CPU affinity on Windows and macOS (fixes #59)

### LINKED ISSUE #59: Add support for setting the CPU affinity on Windows and macOS
We should add support for setting the CPU affinity on Windows and macOS, which would make the `-cpu`/`-core` options more useful and should result in a performance improvement. 

For Windows, when built without the hwloc library, this would need some Windows specific code and when built with hwloc, the existing Linux code would need to be refactored so that it runs for both Linux and Windows: https://github.com/primesearch/Mlucas/blob/b6dda5253c9edfc34d5fbcdad12704bb61941370/src/threadpool.c#L304-L423

For macOS, it does not support setting the CPU affinity directly, but @Artoria2e5 suggested a workaround in this [forum thread](https://www.mersenneforum.org/node/1093514). See this commit for Prime95: https://github.com/primesearch/Prime95/commit/e477e92ff5e0f5c6ccf7cbfbe910bd944a22e480

Ernst's TODOs: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/docs/todo.txt#L929 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/threadpool.c#L282 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/threadpool.c#L286 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/threadpool.c#L695

MarkRose:
Fixes #59.

Mlucas pins its FFT worker threads to logical cores, but only on Linux/FreeBSD; Windows and macOS silently ignored the affinity request. This adds both, in the `worker_thr_routine()` affinity chain in `src/threadpool.c`.

**Windows** — `SetThreadGroupAffinity(GetCurrentThread(), &grp_aff, NULL)`, with `grp_aff.Group = i >> 6` and `grp_aff.Mask = 1 << (i & 63)`. The branch is guarded by `defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)`. This matters: `platform.h` maps MinGW to `OS_TYPE_LINUX`, and the Linux branch excludes `__MINGW32__` (MinGW has no `sched_setaffinity`), so the **MSYS2/MinGW Windows builds the CI actually produces were falling through to the no-affinity `#else`**. MinGW-w64 provides the full Win32 API, so both native-Windows and MinGW now take this path. Failure warns (via `GetLastError()`) and continues, like the Linux branch. The group-affinity form is used rather than the simpler `SetThreadAffinityMask` precisely so that systems with **more than 64 logical CPUs** — which Windows partitions into processor groups — are handled. (An earlier revision of this description said the opposite: that the single-mask form was used and >64-CPU support was absent. Commit `7d1e9f99` replaced it and the text was never updated.)

**macOS** — Mach `THREAD_AFFINITY_POLICY` via `thread_policy_set()` on `pthread_mach_thread_np(pthread_self())`. **macOS has no hard core pinning** — this only sets an advisory *affinity tag* (a hint to co-locate same-tag threads in a shared-L2 domain), which is documented plainly in the code; Apple Silicon effectively ignores it. This also repairs the pre-existing (broken, effectively-disabled) macOS block, which (1) applied the affinity struct with the wrong `THREAD_EXTENDED_POLICY` flavor/count so the tag was never set, (2) `exit(-1)`'d on failure instead of warning, and (3) leaked a mach port via `mach_thread_self()` (switched to `pthread_mach_thread_np`, which doesn't).

Both branches warn-and-continue and are fully `#ifdef`-guarded.

**Two further changes, not in the original description:**

* **hwloc now applies on every platform, not just Linux.** The hwloc block moves out of the `OS_TYPE_LINUX && !__MINGW32__` branch into a top-level `#if INCLUDE_HWLOC` arm that precedes and supersedes every OS branch. In a `use_hwloc` build the FreeBSD, Windows and macOS native-API paths above are therefore unreachable. This is the largest functional change in the PR; the earlier text claiming "the Linux/FreeBSD/hwloc paths are untouched" was wrong.
* **`src/threadpool.h`** loses the `xmalloc`/`xcalloc`/`xrealloc` block (`xrealloc` deleted outright, the other two moved into `threadpool.c`). The PR touches two files, not one.

**Also changed:** `threadpool.c` adds `&& defined(CPU_IS_X86)` to the `__attribute__((force_align_arg_pointer))` guard, narrowing it from *every* GNU-C target to 32-bit x86. That matches the workaround's purpose — the comment above it cites a 2008 pthreads-win32 report about **win32** thread entry giving less than 16-byte alignment. Measured: on x86-64 the attribute is a no-op (gcc 16.1.1 `-O3 -mavx2` emits byte-identical assembly with and without it, no warning), because both the SysV and Windows x64 ABIs already guarantee 16-byte alignment at entry; on aarch64 it produces `warning: 'force_align_arg_pointer' attribute directive ignored [-Wattributes]`. So the narrower guard removes a real ARM warning and costs nothing elsewhere. The Windows/macOS branches can't be compiled on this Linux host, so their API usage was reviewed against the documented signatures — the Windows branch in particular is worth a compile-check on a MinGW runner.

### DIFF
--- a/src/threadpool.c
+++ b/src/threadpool.c
@@ -71,6 +71,25 @@ me at: heber.tomer@gmail.com
 
 #ifdef MULTITHREAD	// Wrap contents of this file in flag (set via platform.h at compile time) ensuring no code built in unthreaded mode
 
+// MacOS has its own versions of these, in /usr/include/X11/Xthreads.h:
+static void * xmalloc(size_t len) {
+	void *ptr = malloc(len);
+	if (ptr == NULL) {
+		printf("failed to allocate %u bytes\n", (uint32)len);
+		exit(-1);
+	}
+	return ptr;
+}
+
+static void * xcalloc(size_t num, size_t len) {
+	void *ptr = calloc(num, len);
+	if (ptr == NULL) {
+		printf("failed to calloc %u bytes\n", (uint32)(num * len));
+		exit(-1);
+	}
+	return ptr;
+}
+
 	#define THREAD_POOL_DEBUG	0
 
 	#if THREAD_POOL_DEBUG
@@ -259,7 +278,7 @@ me at: heber.tomer@gmail.com
 	 * @param data Contains a pointer to the startup data
 	 * @return NULL.
 	 */
-	#if defined(__GNUC__) && (__GNUC__ > 4 || __GNUC__ == 4 && __GNUC_MINOR__>1)
+	#if defined(__GNUC__) && (__GNUC__ > 4 || __GNUC__ == 4 && __GNUC_MINOR__>1) && defined(CPU_IS_X86)
 
 	/* gcc on win32 needs to force 16-byte stack alignment on
 	   thread entry, as this exceeds what windows may provide; see
@@ -270,8 +289,8 @@ me at: heber.tomer@gmail.com
 	#endif
 	static void *worker_thr_routine(void *data)
 	{
-		char cbuf[STR_MAX_LEN*2];
 	#if INCLUDE_HWLOC
+		char cbuf[STR_MAX_LEN*2];
 		char str[80];
 	#endif
 		struct thread_init *init = (struct thread_init *)data;
--- a/src/threadpool.c
+++ b/src/threadpool.c
@@ -65,6 +65,12 @@ me at: heber.tomer@gmail.com
 #include "threadpool.h"
 #include "util.h"	// This is to get (or not) <hwloc.h>
 
+#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
+	#include <windows.h>	// Windows CPU-affinity API: SetThreadAffinityMask()/GetCurrentThread(). MinGW is mapped to
+				// OS_TYPE_LINUX by platform.h but has no sched_setaffinity(), so its Windows build takes this
+				// Win32 path too (MinGW-w64 provides the full Win32 API).
+#endif
+
 #ifdef MULTITHREAD	// Wrap contents of this file in flag (set via platform.h at compile time) ensuring no code built in unthreaded mode
 
 	#define THREAD_POOL_DEBUG	0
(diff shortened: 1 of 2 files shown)


src/threadpool.c
```diff
@@ -383,37 +389,63 @@ me at: heber.tomer@gmail.com
...
+		// Windows hard affinity: pin the current worker thread to logical core i. This covers BOTH the
+		// native-Windows toolchain (OS_TYPE_WINDOWS) and MinGW (which platform.h maps to OS_TYPE_LINUX but
+		// which has no sched_setaffinity() and so is deliberately excluded from the Linux branch above) -
+		// MinGW-w64 provides SetThreadAffinityMask(), so this is the affinity path for the MSYS2/MinGW
+		// Windows builds the CI actually produces.
+		// SetThreadAffinityMask() takes a single DWORD_PTR bitmask, so this simple form only
+		// addresses logical CPUs 0-63 within the thread's current processor group. Systems with
+		// >64 logical CPUs are partitioned into 64-CPU processor groups and require the
+		// SetThreadGroupAffinity()/GROUP_AFFINITY API, which we do not handle here; the
+		// single-mask form matches the model used elsewhere in the Windows port.
```
tdulcet: I know there are several Mlucas users on systems with greater than 64 CPU threads (such as Xeon Phi's with 256 threads), so if it is not too difficult, I believe it would be good for us to support this.

src/threadpool.c
```diff
@@ -383,37 +389,63 @@ me at: heber.tomer@gmail.com
...
+	#elif defined(OS_TYPE_MACOSX)
+
+		// macOS provides NO hard CPU pinning. The only documented mechanism is the Mach
+		// THREAD_AFFINITY_POLICY, which sets an *affinity tag*: an advisory hint asking the
+		// scheduler to co-locate (share an L2 cache domain) all threads that carry the same tag.
+		// It does NOT pin a thread to a specific logical core and gives no placement guarantee.
+		// We use each worker's logical-core index as its tag so every worker gets a distinct
+		// grouping hint, mirroring the *intent* of the hard pin done on Linux/FreeBSD/Windows.
+		// NOTE: This is advisory only. Apple Silicon (arm64) effectively ignores affinity tags,
+		// so on those machines this call is a no-op.
```
tdulcet, line 434: The idea with #59 was to use a similar approach as Prime95/MPrime on macOS, as recently implemented in this commit: https://github.com/primesearch/Prime95/commit/e477e92ff5e0f5c6ccf7cbfbe910bd944a22e480 and merged in this commit: https://github.com/primesearch/Prime95/commit/923679705f691cef02e6c05f9d647b9b4e2a8026. (Note that it is untested and may not work.)

src/threadpool.c
```diff
@@ -65,6 +65,12 @@ me at: heber.tomer@gmail.com
 #include "threadpool.h"
 #include "util.h"	// This is to get (or not) <hwloc.h>
 
+#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
+	#include <windows.h>	// Windows CPU-affinity API: SetThreadAffinityMask()/GetCurrentThread(). MinGW is mapped to
+				// OS_TYPE_LINUX by platform.h but has no sched_setaffinity(), so its Windows build takes this
+				// Win32 path too (MinGW-w64 provides the full Win32 API).
```
tdulcet: This comment is redundant with the below comments.

tdulcet (changes requested):
Thanks for your work on this. It will be a big improvement for users on Windows and macOS.

> the Linux/FreeBSD/hwloc paths are untouched

We should update the hwloc path to apply to all OSs, not just Linux, when Mlucas is built with `-DINCLUDE_HWLOC` (the `use_hwloc` option to the `makemake.sh` script). This would make the new `-core` option added in v21 work as expected. The new Windows and macOS CPU affinity code would then be used when it is built without `-DINCLUDE_HWLOC`.

MarkRose:
Thanks for the detailed review — addressed in ccc441d.

**hwloc affinity on all OSes.** Hoisted the `#if INCLUDE_HWLOC` path to the top of the per-thread affinity `#if` chain, so when Mlucas is built with `-DINCLUDE_HWLOC` (the `use_hwloc` makemake option) hwloc drives affinity on **every** platform — Windows and macOS as well as Linux/*BSD — which is what makes the v21 `-core` option work as intended everywhere. The native Win32 / macOS / *BSD paths are now the fallback used only in non-hwloc builds. This also matches how Prime95/mprime does it: mprime routes all affinity through hwloc (its native paths delegate to it), and hwloc handles the Windows >64-CPU processor-group partitioning internally.

**>64 CPUs on Windows (threadpool.c:403).** For the non-hwloc Windows fallback I switched from `SetThreadAffinityMask()` (single 64-bit mask, group-0 only) to `SetThreadGroupAffinity()`/`GROUP_AFFINITY`, mapping logical core `i` to group `i>>6`, bit `i&63`. So Xeon Phi-class >64-thread machines are now handled on Windows in both build modes (via hwloc when built with it, via processor groups otherwise).

**macOS (threadpool.c:430).** The current code already uses the same mechanism as the Prime95 commit you linked — `thread_policy_set()` with `THREAD_AFFINITY_POLICY` and `affinity_tag`, obtained via `pthread_mach_thread_np()` — so I've left it as-is. As you and the Prime95 notes point out, this is advisory and Apple Silicon returns `EPFNOSUPPORT`; the code warns rather than aborting in that case. If you'd prefer we also mirror Prime95's `hw.cacheconfig`/`hw.ncpu` core-count derivation, I can look at that separately, but the affinity call itself already matches.

**Redundant comment (threadpool.c:71).** Removed — trimmed the `windows.h` include comment down to a one-liner (the detail now lives only in the affinity branch). Also dropped a dead `#if 0` pthread-affinity block in the Linux branch while I was in there.

Verified on Linux that both the hwloc and non-hwloc avx2 builds compile and set 4-core affinity correctly (Res64 `71E61322CCFB396C`), and the new `SetThreadGroupAffinity` calls compile clean under `x86_64-w64-mingw32-gcc`. I don't have Windows/macOS hardware here, so runtime confirmation of those two native paths would still be welcome.

src/threadpool.c
```diff
@@ -303,30 +345,6 @@ me at: heber.tomer@gmail.com
...
-
-		pth = pthread_self();
-		errcode = pthread_setaffinity_np(pth, cpuset_size(cset), cset);
-		if (errcode) {
-			perror("pthread_setaffinity_np");
-		}
-		cpuset_destroy(cset);
-	  */
-	  #else
-
```
tdulcet, line 329: Can you please keep this block. I am not sure what Ernst's plans for it were, and I do not want to remove code that we do not fully understand the reasoning behind. I recall we already fixed that listed `cpuset_t` error, so it might work now.

src/threadpool.c
```diff
@@ -277,7 +281,45 @@ me at: heber.tomer@gmail.com
...
+			ASSERT(0, "Aborting.");
+		}
+
+	  if(HWLOC_AFFINITY) {	// Global, declared in Mdata.h, defined in Mlucas.c, set in util.c::host_init()
+		hwloc_bitmap_t cpuset = hwloc_bitmap_alloc();
+		hwloc_obj_t obj = hwloc_get_obj_by_type(hw_topology, HWLOC_OBJ_PU, i);
+		if (obj) {
+			hwloc_bitmap_or(cpuset, cpuset, obj->cpuset);
+		} else {
+			snprintf(cbuf,STR_MAX_LEN*2,"[hwloc] Error: HWLOC_OBJ_PU[%u] not found.\n",i);
```
tdulcet: Can you replace the `STR_MAX_LEN*2` with `sizeof()`, as we did in #92:
```suggestion
			snprintf(cbuf,sizeof(cbuf),"[hwloc] Error: HWLOC_OBJ_PU[%u] not found.\n",i);
```
This also occurs several times below.

tdulcet (changes requested):
Thanks, just a couple of minor changes are needed.

src/threadpool.c
MarkRose: Done

src/threadpool.c
MarkRose: Done

src/threadpool.c
```diff
@@ -65,8 +65,31 @@ me at: heber.tomer@gmail.com
...
+
+static void * xcalloc(size_t num, size_t len) {
+	void *ptr = calloc(num, len);
+	if (ptr == NULL) {
+		printf("failed to calloc %u bytes\n", (uint32)(num * len));
+		exit(-1);
+	}
+	return ptr;
+}
+
```
tdulcet: Are these two function used somewhere, as I see we dropped `xrealloc`? If so, could we use the new macros added in #88 instead?

MarkRose: Yes — `xcalloc` is used at two sites (the task queue and the pool allocation) and `xmalloc` at one; only `xrealloc` was unused, which is why that one was dropped. Routing these through the checked-alloc wrappers from #88 is a good idea and I would like to do it — but I am holding #88 until the various #34-split PRs land, to avoid compounding rebase conflicts across all of them. So I would prefer to keep these local helpers here for now and switch them to the #88 macros as a small follow-up once #88 is rebased in. Does that work for you?

tdulcet: > So I would prefer to keep these local helpers here for now and switch them to the https://github.com/primesearch/Mlucas/pull/88 macros as a small follow-up once https://github.com/primesearch/Mlucas/pull/88 is rebased in. Does that work for you?

Thanks for the clarification. Now that all of the PRs split from #34 have landed, should this switch be done now?

MarkRose:
Yes — done now that #88 has landed.

`threadpool.c` already includes `util.h`, so `MALLOC`/`CALLOC` and `alloc_fail()` were already in scope. Dropped the local `static xmalloc()`/`xcalloc()` and pointed the three call sites at the shared macros. Behaviour is equivalent — both abort on a failed allocation — but the failure now goes through `alloc_fail()`, so it reports file, line and function rather than just a byte count.

It also removes the reason for the odd names: the helpers were `x*alloc` because, per the comment now deleted, *"MacOS has its own versions of these, in /usr/include/X11/Xthreads.h"*. With them gone there's nothing left to collide.

Verified on avx2 — builds with zero warnings from `threadpool.c`, and since the pool is exercised on every threaded run: M3888509 at 1, 4 and 8 threads all give `Res64: 71E61322CCFB396C`, and `-s tt -cpu 0:7` is rc=0 with 19 residues.

tdulcet (dismissed):
Thanks, this looks great!

MarkRose:
Rebased onto current `main` to clear the merge conflict — sorry that this dismisses the approval, but the branch could not be merged as it stood. One conflict, in `src/threadpool.c`: this branch adds a `<windows.h>` include for the affinity API and #251 added a FreeBSD `<sys/cpuset.h>` block at the same spot, so each now keeps its own guard rather than sharing an `#endif`. Re-verified after: builds clean and `-s tt` is rc=0 with 19 residues.

tdulcet (approved):

PR #111, opened 2026-07-16
help.txt: update version stamp to v21.0.2

MarkRose:
> **Revised — this is no longer a two-line change.** A second commit (`edb50e53`) restored several paragraphs from earlier review rounds, and the description below predates it. Two of those paragraphs documented behaviour `main` does not have and have since been removed again (`3ee149d7`):
>
> * *"The minimum supported FFT length is 2K; 1K is not supported"* — `MIN_FFT_LENGTH_IN_K` is 1, and `get_fft_radices.c` has a live `case 1`, so `-fft 1` is accepted today. Rejecting it is **#104**.
> * A p-1 restart message about recovering B1 from `.s1_prod` and prompting for `-b1` — no such string or behaviour exists in `src/`. Recovering B1 on restart is **#72**.
>
> Both belong with the PRs that make them true. What remains from that commit is the OUT-OF-MEMORY note (`main` has 17 allocation-failure diagnostics; `-pm1_s2_nbuf` is a real parsed flag) and the signals note, which matches the registration in `mers_mod_square.c` / `fermat_mod_square.c`.
>
> So the PR is now: the version stamp, the signals-note correction, and the OUT-OF-MEMORY paragraph.

Narrow, main-only change: sync help.txt's version references to the version string main's code actually reports (`VERSION = "21.0.2"`) — the file was still stamped `v20.1.1`. Touches just the header line and the stale version pin on the interrupt-signals note.

**Scope.** Deliberately limited to what is already true of `main`. help.txt should only describe behaviour once that behaviour has landed, so documentation for the in-flight fixes (savefile-on-interrupt, out-of-memory handling, 1K-FFT rejection, p-1 B1 resume) is not here — it belongs in a separate help.txt PR chained onto the corresponding fix PRs. The broader v21 feature-doc gaps (teensy/Suyama self-tests, self-test FFT ranges) belong in that same later docs pass.

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -1,6 +1,6 @@
 [***** This file best-viewed with a 4-column tab setting, with line-wrapping *****]
 
-================= Mlucas v20.1.1 command line options listing =================
+================= Mlucas v21.0.2 command line options listing =================
 
 Note: besides the basic post-build self-test flags covered by the online README page
 ( http://www.mersenneforum.org/mayer/README.html ),
@@ -635,7 +635,7 @@ Once you have identified the desired process IDs:
 	interval completes, i.e. until the latest savefile update occurs.
 
 For the current list of signal types which the program listens for, see the sig_handler() function near
-the top of Mlucas.c . As of this writing (v20.1.1), the program listens for INT,TERM,HUP,ALRM,USR1 and USR2.
+the top of Mlucas.c . As of this writing, the program listens for INT,TERM,HUP,ALRM,USR1 and USR2.
 
 ======================
 
--- a/help.txt
+++ b/help.txt
@@ -114,24 +114,30 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
  -s {anything other than the mnemonics below}: Self-test, user must also supply exponent (via -m or -f),
 	iteration number and (optionally) FFT length to use.
 
- -s tiny	Runs 100 (<= 4 threads) or 1000-iteration self-tests on a set of 32 Mersenne exponents,
- -s t		covering FFT lengths 8K-120K. This will take around 1 minute on a fast CPU.
+	Run-times below are indicative only: they were measured on a modern multi-core x86-64 CPU
+	using all cores, and vary widely with core count, clock and memory bandwidth.
 
- -s small	Runs 100 or 1000-iteration self-tests on a set of 32 Mersenne exponents, covering
- -s s		FFT lengths 120K-1920K. This will take around 10 minutes on a fast CPU..
+ -s teensy	Runs 100 (<= 4 threads) or 1000-iteration self-tests on a set of 15 Mersenne exponents,
+ -s tt		covering FFT lengths 1K-15K. A few seconds. Useful as a quick post-build smoke test.
+
+ -s tiny	Runs 100 (<= 4 threads) or 1000-iteration self-tests on a set of 32 Mersenne exponents,
+ -s t		covering FFT lengths 16K-240K. Well under a minute.
+
+ -s small	Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering
+ -s s		FFT lengths 256K-1.875M. A few minutes.
 
 **************** THIS IS THE ONLY SELF-TEST ORDINARY USERS ARE RECOMMENDED TO DO: *************
 *                                                                                             *
-* -s medium Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering *
-* -s m      FFT lengths from 2M to 7.5M. This will take around an hour on a fast CPU.         *
+* -s medium Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering *
+* -s m      FFT lengths from 2M to 15M. Allow up to an hour.                                  *
 *                                                                                             *
 ***********************************************************************************************
 
- -s large	Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering
- -s l		FFT lengths 8M to 60M. This will take around 10 hours on a fast CPU.
+ -s large	Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering
+ -s l		FFT lengths 16M to 60M. Several hours.
 
  -s all		Runs 100 or 1000--iteration self-tests of all the above FFT lengths.
- -s a		This will take around 12 hours on a fast CPU.
+ -s a		Note this covers teensy through large only, not huge or egregious.
 
  -s xl		[The 'h' form is short for 'huge'.] Runs 100 or 1000-iteration self-tests on a set
  -s h		of 16 M(p), covering FFT lengths 64M-240M. This will take several days on a fast CPU.
--- a/help.txt
+++ b/help.txt
@@ -161,6 +161,9 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
 	value must map to a supported FFT length; for Mlucas these are of form k × 2^n, where k
 	is an odd integer in the set (1,3,5,7,9,11,13,15).
 
+	The minimum supported FFT length is 2K; 1K is not supported. A -fft request below the minimum
+	(or above the current maximum, 512M) is rejected at startup with an explicit range error.
+
 	If -fft is invoked with a user-supplied value of -iters but without a
 	user-supplied exponent, the program will do the specified number of iterations
 	using the default self-test Mersenne or Fermat-number exponent for that FFT length.
@@ -580,6 +583,11 @@ integer hardware multiply where available. (Speeding things by replacing the O(n
 school" multiply with a subquadratic one is a possible future enhancement.) For a given
 stage 1 bound B1, the s1_prod file will be roughly 0.2*B1 bytes in size.
 
+o A p-1 stage 1 run resumes at the B1 bound it started with, even if the default B1 would now be
+different (Mlucas can auto-size B1 based on available memory). A restart recovers the original B1
+from the .s1_prod savefile; if that file is missing, the restart stops and asks you to either rerun
+with the original '-b1 <N>' or delete the exponent's p-1 savefiles to start fresh.
+
 ======================
 
 [13]: *** DON'Ts ***
@@ -601,6 +609,11 @@ Please start by looking for posts about your issue in the
 mersenneforum.org thread specific to the release you are using in the mersenneforum.org
 Mlucas subforum at https://www.mersenneforum.org/forumdisplay.php?f=118.
 If you don't find anything, make a post describing your problem.
+
+o OUT-OF-MEMORY: if the system cannot satisfy a memory allocation (e.g. -maxalloc set too high for the
+actual available RAM, or too little free RAM for the requested FFT length or p-1 stage 2 buffer count),
+Mlucas exits with a clear out-of-memory message. If you see this, lower -maxalloc, use fewer p-1 stage 2
+buffers (via -pm1_s2_nbuf), or free up system RAM.
 
 [14a]: How to safely interrupt a running instance:
 
@@ -635,7 +648,8 @@ Once you have identified the desired process IDs:
 	interval completes, i.e. until the latest savefile update occurs.
 
 For the current list of signal types which the program listens for, see the sig_handler() function near
-the top of Mlucas.c . As of this writing, the program listens for INT,TERM,HUP,ALRM,USR1 and USR2.
+the top of Mlucas.c . As of this writing, the program listens for INT,TERM,HUP,ALRM,USR1 and USR2 on
+Linux/macOS; native Windows/MinGW builds only catch INT and TERM.
 
 ======================
 
--- a/help.txt
+++ b/help.txt
@@ -161,9 +161,6 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
 	value must map to a supported FFT length; for Mlucas these are of form k × 2^n, where k
 	is an odd integer in the set (1,3,5,7,9,11,13,15).
 
-	The minimum supported FFT length is 2K; 1K is not supported. A -fft request below the minimum
-	(or above the current maximum, 512M) is rejected at startup with an explicit range error.
-
 	If -fft is invoked with a user-supplied value of -iters but without a
 	user-supplied exponent, the program will do the specified number of iterations
 	using the default self-test Mersenne or Fermat-number exponent for that FFT length.
@@ -582,11 +579,6 @@ not made a special effort to optimize the computation of said product beyond use
 integer hardware multiply where available. (Speeding things by replacing the O(n^2) "grammar
 school" multiply with a subquadratic one is a possible future enhancement.) For a given
 stage 1 bound B1, the s1_prod file will be roughly 0.2*B1 bytes in size.
-
-o A p-1 stage 1 run resumes at the B1 bound it started with, even if the default B1 would now be
-different (Mlucas can auto-size B1 based on available memory). A restart recovers the original B1
-from the .s1_prod savefile; if that file is missing, the restart stops and asks you to either rerun
-with the original '-b1 <N>' or delete the exponent's p-1 savefiles to start fresh.
 
 ======================
 
--- a/help.txt
+++ b/help.txt
@@ -146,10 +146,16 @@ Options - again note the user can override the default iteration count based on 
 	       medium      756.1s      345.0s       417.6s
 	       large      2225.8s     1068.1s      1233.8s
 
-	   4 threads is 2.1 - 2.6x one thread rather than 4x, and was independently measured at
-	   2.25x on different hardware. Past that it inverts: at medium and large, 16 threads is
-	   slower than 4. Unless you have measured otherwise on your own hardware, 4 threads is a
-	   good default for these tiers.
+	   Read those as self-test costs, not as how well Mlucas itself threads. A self-test runs
+	   every radix set for the length, good and bad, and adds the bookkeeping to pick a winner,
+	   so its wall clock averages over candidates rather than reporting the one you end up
+	   using. Pinned to a single radix set and FFT length, 1 -> 4 cores measures about 3x here,
+	   against a ceiling nearer 3.5x once the all-core clock drop is allowed for.
+
+	   For choosing how to run a self-test, though, the table is the relevant thing: 4 threads
+	   is the sweet spot, and past it the extra iterations cost more than the extra threads
+	   return, so at medium and large 16 threads finishes later than 4. Unless you have
+	   measured otherwise on your own hardware, 4 threads is a good default for these tiers.
 
  -s teensy	Runs 100 (<= 4 threads) or 1000-iteration self-tests on a set of 15 Mersenne exponents,
  -s tt		covering FFT lengths 1K-15K. Useful as a quick post-build smoke test; see the table above
--- a/help.txt
+++ b/help.txt
@@ -117,14 +117,14 @@ Options - again note the user can override the default iteration count based on 
 	same test spans well over 10x across machines. The figures are for ONE core at the default
 	100 iterations - the only configuration that compares cleanly between systems - and come
 	from two x86-64 laptop CPUs several generations apart, which agree within about 2x where
-	both were run; the last three rows are from the older of the two only:
+	both were run; the last two rows are from the older of the two only:
 
 	  -s         FFT lengths      one core, 100 iters
 	  teensy     1K - 15K         a second or less
 	  tiny       16K - 240K       5 - 10 seconds
 	  small      256K - 1.875M    1 - 2 minutes
 	  medium     2M - 15M         13 - 21 minutes
-	  large      16M - 60M        about an hour
+	  large      16M - 60M        35 - 60 minutes
 	  huge       64M - 240M       about 5 hours
 	  egregious  256M - 512M      about 14 hours
 
--- a/help.txt
+++ b/help.txt
@@ -140,13 +140,24 @@ Options - again note the user can override the default iteration count based on 
 	   on 16 - the 16-core run is slower in wall-clock terms because it is doing ten times as
 	   much work. Pin -iters when comparing thread counts instead of relying on the defaults.
 
-	 - Extra cores do not pay off proportionally: 1 core to 4 measured 2.2x on one machine and
-	   2.25x on a quite different one, against a theoretical 4x.
+	 - Extra cores do not pay off proportionally, and past 4 they can cost. Normalising the
+	   iteration counts so that equal work is compared, on an 8-core/16-thread x86-64 machine:
+
+	                  1 thread   4 threads   16 threads   (cost per 100 iterations)
+	       small        64.8s       24.5s        16.5s
+	       medium      756.1s      345.0s       417.6s
+	       large      2225.8s     1068.1s      1233.8s
+
+	   4 threads is 2.1 - 2.6x one thread rather than 4x, and was independently measured at
+	   2.25x on quite different hardware. Past that it inverts: at medium and large - the
+	   sizes users actually run - 16 threads is SLOWER than 4. Unless you have measured
+	   otherwise on your own hardware, 4 threads is a good default for these tiers.
 
 	A high thread count also tests *less*, not merely faster, because the smaller FFT lengths
 	cannot use many threads and drop out of the run: '-s small' reports 127 residues on 1 or 4
 	cores but 116 on 16, and '-s teensy' 23 against 16. If the aim is to populate mlucas.cfg
-	rather than to benchmark, 4 cores is the better choice.
+	rather than to benchmark, 4 cores is the better choice. ('-s large' is not affected: 63
+	residues at every thread count, since 16M-60M FFTs can use the threads.)
 
  -s teensy	Runs 100 (<= 4 threads) or 1000-iteration self-tests on a set of 15 Mersenne exponents,
  -s tt		covering FFT lengths 1K-15K. Useful as a quick post-build smoke test; see the table above
--- a/help.txt
+++ b/help.txt
@@ -113,11 +113,10 @@ Options - again note the user can override the default iteration count based on 
  -s {anything other than the mnemonics below}: Self-test, user must also supply exponent (via -m or -f),
 	iteration number and (optionally) FFT length to use.
 
-	How long each tier takes. These are measured, but treat them as an order of magnitude: the
-	same test spans well over 10x across machines. The figures are for ONE core at the default
-	100 iterations - the only configuration that compares cleanly between systems - and come
-	from two x86-64 laptop CPUs several generations apart, which agree within about 2x where
-	both were run; the last two rows are from the older of the two only:
+	How long each tier takes. Treat these as orders of magnitude: the same test spans well over
+	10x across machines. They are for one core at the default 100 iterations, which is the only
+	configuration that compares cleanly between systems, and were measured on two x86-64 laptop
+	CPUs several generations apart. The last two rows are from the older of the two only.
 
 	  -s         FFT lengths      one core, 100 iters
 	  teensy     1K - 15K         a second or less
@@ -128,17 +127,16 @@ Options - again note the user can override the default iteration count based on 
 	  huge       64M - 240M       about 5 hours
 	  egregious  256M - 512M      about 14 hours
 
-	A small ARM core is far slower than that - a Raspberry Pi 4B measured roughly 10x the x86
-	figures at the tiers where both were run - so a Pi is a poor choice past 'small'.
+	A small ARM core is much slower: a Raspberry Pi 4B measured roughly 10x the x86 figures at
+	the tiers where both were run.
 
-	Two things make wall-clock time counter-intuitive, and both are measured rather than
-	supposed:
+	Two things make wall-clock time counter-intuitive:
 
 	 - Crossing 4 threads multiplies the work by ten, because the default iteration count is
 	   100 for <= 4 threads and 1000 for more, and run time is linear in the iteration count
 	   to within a couple of percent. '-s small' takes 25 seconds on 4 cores and 165 seconds
 	   on 16 - the 16-core run is slower in wall-clock terms because it is doing ten times as
-	   much work. Pin -iters when comparing thread counts instead of relying on the defaults.
+	   much work.
 
 	 - Extra cores do not pay off proportionally, and past 4 they can cost. Normalising the
 	   iteration counts so that equal work is compared, on an 8-core/16-thread x86-64 machine:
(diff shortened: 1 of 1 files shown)


help.txt
```diff
@@ -602,13 +622,36 @@ mersenneforum.org thread specific to the release you are using in the mersennefo
...
+the next safe polling point between mod-squaring iterations, so the residue that gets saved is always
+the exact last-completed-iteration one, never a torn or in-flight one. FFT worker threads have these
+signals blocked, so the signal is always handled on the main thread regardless of which thread the OS
+happens to interrupt, making shutdown deterministic. (For a stretch of releases starting Dec 2021, this
+savefile-on-interrupt behavior was disabled -- the old handler called non-async-signal-safe functions,
+which could deadlock a multithreaded run if the signal landed on a worker thread already holding a lock,
+causing intermittent "run refuses to quit" hangs. That instability has since been fixed at the root by
+making the handler itself do nothing but set a flag, so the clean savefile-on-interrupt behavior
+described above is restored.) Interactive/self-test runs (-s, or -m/-f with -iters) still just exit
+immediately without writing a savefile on interrupt, since self-tests have no checkpoint to resume from.
```
tdulcet: I do not believe this detailed technical explanation of the previous bug and how it was fixed is needed in the usage documentation here. All users need to know is that Mlucas will write a savefile on interrupt for all work types (expect for P-1 stage 2), so the existing paragraph was already correct, since it was never updated to reflect that this feature was actually disabled in December 2021.

help.txt
```diff
@@ -635,7 +678,8 @@ Once you have identified the desired process IDs:
 	interval completes, i.e. until the latest savefile update occurs.
 
 For the current list of signal types which the program listens for, see the sig_handler() function near
-the top of Mlucas.c . As of this writing (v20.1.1), the program listens for INT,TERM,HUP,ALRM,USR1 and USR2.
+the top of Mlucas.c . As of this writing, the program listens for INT,TERM,HUP,ALRM,USR1 and USR2 on
+Linux/macOS; native Windows/MinGW builds (which lack sigaction()) only catch INT and TERM.
```
tdulcet:
```suggestion
Linux/macOS; native Windows/MinGW builds only catch INT and TERM.
```

help.txt
```diff
@@ -161,6 +161,13 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
 	value must map to a supported FFT length; for Mlucas these are of form k × 2^n, where k
 	is an odd integer in the set (1,3,5,7,9,11,13,15).
 
+	The minimum supported FFT length is 2K; 1K is not supported (the radix-factorization scheme
+	used needs at least one intermediate radix beyond the leading and trailing ones, which a 1K
+	transform cannot provide). A -fft request below the minimum (or above the current maximum,
+	512M) is rejected at startup with an explicit range error, e.g. "ERROR: FFT-length argument
+	= 1, must be in range [2,524288]K", rather than being accepted and failing later, deeper in
+	the FFT machinery.
```
tdulcet: We do not need to repeat the error message, which could change in the future:
```suggestion
	512M) is rejected at startup with an explicit range error.
```

tdulcet (changes requested):
Thanks for your work on improving the documentation.

These two files are designed for users just starting out with Mlucas, so I believe we should keep them as nontechnical as possible. Any needed technical explanations or low level implementation details should likely go into #112. I made a few suggestions along those lines.

If you are up for it, the `help.txt` file could use a more comprehensive update to reflect some of Ernst's changes in [v21](https://www.mersenneforum.org/node/22731), as well as our previous changes in [v21.0.1](https://www.mersenneforum.org/node/22731?p=888014#post888014) and [v21.0.2](https://www.mersenneforum.org/node/1082613) releases. See those links for the release notes for each version on the forum. For example, the FFT length ranges for each self-test are outdated, and it is missing the new teensy self-test. Also see #38.

help.txt
MarkRose: Done — shortened to just 'rejected at startup with an explicit range error' and dropped the radix-factorization rationale.

help.txt
MarkRose: Done — replaced the handler internals and the Dec-2021 history with a plain user-level description: it writes a savefile at the last completed iteration and exits (with the self-test and p-1-stage-2 caveats).

help.txt
MarkRose: Applied your suggestion.

help.txt
```diff
@@ -602,13 +610,19 @@ mersenneforum.org thread specific to the release you are using in the mersennefo
 Mlucas subforum at https://www.mersenneforum.org/forumdisplay.php?f=118.
 If you don't find anything, make a post describing your problem.
 
+o OUT-OF-MEMORY: if the system cannot satisfy a memory allocation (e.g. -maxalloc set too high for the
+actual available RAM, or too little free RAM for the requested FFT length or p-1 stage 2 buffer count),
+Mlucas exits with a clear out-of-memory message rather than crashing with a bare segmentation fault. If
```
tdulcet:
```suggestion
Mlucas exits with a clear out-of-memory message. If
```

help.txt
```diff
@@ -602,13 +610,19 @@ mersenneforum.org thread specific to the release you are using in the mersennefo
...
 Note that halting (as opposed to merely suspending using 'kill -CONT' as described below) an Mlucas
-instance should produce a clean "at end of the current iteration, write savefiles and exit" for LL-test,
-PRP and p-1 stage 1 processing. For p-1 stage 2, the state machine involved is of sufficient complexity
-that I have not (yet) implemented a clean-exit-with-savefile-write: an interrupt will cause you to lose
-whatever work has been done since the last stage 2 savefile (p[expo].s2) write.
+instance writes a savefile at the last completed iteration and then exits, for LL-test, PRP and p-1
+stage 1 processing, so no work is lost. Interactive/self-test runs (-s, or -m/-f with -iters) exit
+immediately without writing a savefile, since self-tests have no checkpoint to resume from. For p-1
+stage 2, an interrupt loses whatever work has been done since the last stage 2 savefile (p[expo].s2)
+write, but the program still exits promptly.
```
tdulcet: I would keep the original phrasing here. A user would not expect the self-test to produce savefiles (and actually it does in some cases, due to bugs).

help.txt
```diff
@@ -580,6 +583,11 @@ integer hardware multiply where available. (Speeding things by replacing the O(n
 school" multiply with a subquadratic one is a possible future enhancement.) For a given
 stage 1 bound B1, the s1_prod file will be roughly 0.2*B1 bytes in size.
 
+o A p-1 stage 1 run resumes at the B1 bound it started with, even if the default B1 would now be
+different (Mlucas can auto-size B1 based on available memory). A restart recovers the original B1
+from the .s1_prod savefile; if that file is missing, the restart stops and asks you to either rerun
+with the original '-b1 <N>' or delete the exponent's p-1 savefiles to start fresh.
```
tdulcet: I would keep the mention from your previous commit (https://github.com/primesearch/Mlucas/pull/111/changes/f3d74c42a093779bb09eacbe294c0a639bb26e51) about automatically using a larger B1 when there is not enough RAM to do stage 2, based on the `-maxalloc`/`-pm1_s2_nbuf` options, as well as that different B1 values are incompatible.

tdulcet (changes requested):
Thanks, this looks much better.

As I mentioned above, it might be good to update the `help.txt` file with all of Ernst's changes from v21, to avoid potenchally confusing users about the expected behavior:
> If you are up for it, the `help.txt` file could use a more comprehensive update to reflect some of Ernst's changes in [v21](https://www.mersenneforum.org/node/22731), as well as our previous changes in [v21.0.1](https://www.mersenneforum.org/node/22731?p=888014#post888014) and [v21.0.2](https://www.mersenneforum.org/node/1082613) releases. See those links for the release notes for each version on the forum.

MarkRose:
I've rescoped this PR down to just the version-stamp sync (`v20.1.1` → `21.0.2`), since on reflection most of what it was doing described behavior from our in-flight *fix* PRs rather than what's in main. Our campaign is fixes, not new features, so help.txt shouldn't document that behavior until the fixes land.

So the pieces you reviewed here are moving to a **separate help.txt PR chained onto the fix PRs** (savefile-on-interrupt → #108, 1K-FFT rejection → #104, out-of-memory handling, p-1 B1 resume, etc.), where each doc paragraph rides with the code that makes it true. That chained PR will also cover the genuine v21 feature-doc gaps I found while here — the `-s teensy` self-test and Suyama cofactor testing aren't documented, and the self-test FFT ranges (e.g. medium still says "2M to 7.5M") are stale relative to v21/v21.0.2.

Your specific round-2 notes (sizeof-style out-of-memory wording, the self-test-savefile phrasing, and restoring the B1 auto-size/incompatibility note) will be carried into that chained PR. Thanks for the detailed review.

tdulcet:
> That chained PR will also cover the genuine v21 feature-doc gaps I found while here — the `-s teensy` self-test and Suyama cofactor testing aren't documented, and the self-test FFT ranges (e.g. medium still says "2M to 7.5M") are stale relative to v21/v21.0.2.

For reference, Ken generated updated self-test times for a slower but multithreaded Xeon Phi system using a AVX2 build in [this forum post](https://www.mersenneforum.org/node/1082613?p=1122010#post1122010) and for a medium range but single threaded AVX512 Core i5-1035G1 system in the post immediately below. Note that GitHub Actions CI is much faster than both of those systems when running the teensy through large self-tests on a the provided 4 core AVX2 or AVX512 runner, so there can be a large variation in the self-test times depending on the speed of the system.

tdulcet (changes requested):
> I've rescoped this PR down to just the version-stamp sync (`v20.1.1` → `21.0.2`), since on reflection most of what it was doing described behavior from our in-flight _fix_ PRs rather than what's in main.

If were are going to update the version number in this file, we should probably first update it to reflect all of the changes that were added in this version. See my above comments for more information.

MarkRose:
Fair point — the version number shouldn't claim v21 until the content describes v21. I've made a start on the self-test section, which was the most wrong part of it.

Rather than work from the forum release notes, I derived every count and range from the source (the `numTeensy`..`numEgregious` defines and the `MersVec` table) and then cross-checked by running each option and reading back the lengths it actually exercises:

| | help.txt said | actually |
| --- | --- | --- |
| **teensy** | *not documented at all* | 15 exponents, FFT 1K–15K |
| tiny | 32, FFT 8K–120K | 32, FFT **16K–240K** |
| small | **32**, FFT 120K–1920K | **24**, FFT **256K–1.875M** |
| medium | **16**, FFT 2M–7.5M | **24**, FFT 2M–**15M** |
| large | **24**, FFT 8M–60M | **16**, FFT **16M**–60M |
| huge | 16, FFT 64M–240M | correct |
| egregious | 9, FFT 256M–512M | correct |

So the teensy self-test Ernst added in v21 was undocumented, and small/medium/large had their counts transposed. Also documented that `-s all` covers teensy through large only, not huge or egregious — the code comment says so (*"all, which really means all the non-Huge-and-larger sets"*) but the option's own entry didn't.

**Run-times** (#38): the old absolute figures are well out. Measured here on 8 threads, `-s tiny` takes **17s** against the 1 minute claimed, and `-s small` **191s** against 10 minutes — while #38 reports 8s and 83s on an M1. That's a 3.6×–7.5× spread depending on the machine, and precise absolute numbers are exactly what rotted in the first place, so I've replaced them with orders of magnitude plus an explicit note that they were measured on a modern multi-core x86-64 using all cores and vary widely.

Left alone deliberately: `brobdingnagian` and `godzillian` are accepted by the parser, but `numBrobdingnagian` and `numGodzillian` are both `0`, so documenting them would advertise self-tests that run nothing.

**What I have not done** is the rest of the v21/v21.0.1/v21.0.2 feature sweep you asked about. I'd rather not paraphrase the forum posts into user-facing docs without being able to check each claim against the code the way I did above — several things I restored from an earlier round of this PR turned out to describe behaviour `main` doesn't have, which is what got them removed again. If you'd point me at the specific areas you consider most out of date, or would rather this land as the self-test correction and the remaining gaps go in a follow-up, either works.

help.txt
```diff
@@ -114,24 +114,30 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
...
-* -s m      FFT lengths from 2M to 7.5M. This will take around an hour on a fast CPU.         *
+* -s medium Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering *
+* -s m      FFT lengths from 2M to 15M. Allow up to an hour.                                  *
 *                                                                                             *
 ***********************************************************************************************
 
- -s large	Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering
- -s l		FFT lengths 8M to 60M. This will take around 10 hours on a fast CPU.
+ -s large	Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering
+ -s l		FFT lengths 16M to 60M. Several hours.
```
tdulcet: The "Several hours" can still be a over estimate. I believe the GitHub Actions runners are able to run this in less than ~20 minutes.

> **Run-times** (#38): the old absolute figures are well out. Measured here on 8 threads, `-s tiny` takes **17s** against the 1 minute claimed, and `-s small` **191s** against 10 minutes

The issue is that your test used 8 threads, so it required 10x the number of iterations by default for only 2x the increased throughput (i.e. the 20 minutes would become 100 minutes = ~1.67 hours).

help.txt
```diff
@@ -114,24 +114,30 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
  -s {anything other than the mnemonics below}: Self-test, user must also supply exponent (via -m or -f),
 	iteration number and (optionally) FFT length to use.
 
+	Run-times below are indicative only: they were measured on a modern multi-core x86-64 CPU
```
tdulcet: Maybe say x86 or ARM, as a multi-core ARM system should be comparable performance.

help.txt
```diff
@@ -635,7 +646,8 @@ Once you have identified the desired process IDs:
 	interval completes, i.e. until the latest savefile update occurs.
 
 For the current list of signal types which the program listens for, see the sig_handler() function near
-the top of Mlucas.c . As of this writing (v20.1.1), the program listens for INT,TERM,HUP,ALRM,USR1 and USR2.
+the top of Mlucas.c . As of this writing, the program listens for INT,TERM,HUP,ALRM,USR1 and USR2 on
+Linux/macOS; native Windows/MinGW builds only catch INT and TERM.
```
tdulcet: Now that Mlucas supports FreeBSD, instead of "Linux/macOS", maybe just say Unix or explicitly list "Linux/FreeBSD/macOS".

tdulcet (changes requested):
> **What I have not done** is the rest of the v21/v21.0.1/v21.0.2 feature sweep you asked about. I'd rather not paraphrase the forum posts into user-facing docs without being able to check each claim against the code the way I did above — several things I restored from an earlier round of this PR turned out to describe behaviour `main` doesn't have, which is what got them removed again.

You would likely need to look at the individual commits then, which is what I previously used to write those release notes for the forum:
* v21 part 1: https://github.com/primesearch/Mlucas/commit/c9fef37ac435d7989145e801288f16de99208696 and part 2: https://github.com/primesearch/Mlucas/commit/77749764e9f49906e0208e05bc5ba0c375a044a9
* v21.0.1: https://github.com/primesearch/Mlucas/compare/f4a66f43dba19be896cb0992f9b8abf5842667c1...v21.0.1
* v21.0.2: https://github.com/primesearch/Mlucas/compare/v21.0.1...v21.0.2
* Commits after v21.0.2, but before your PRs: https://github.com/primesearch/Mlucas/compare/v21.0.2...1f6068104d4518348ae3f3f265a48e9ca54a629d

---

You might also want to consider pulling in information from the Mlucas manpage: https://manpages.debian.org/testing/mlucas/mlucas.1.en.html

MarkRose:
All three inline points addressed, and thank you for the correction on the timings — you are right and my numbers were not comparable.

**The run-time measurement was invalid.** `help.txt` itself says it: *"The default number of iterations for the self-test is 100 for <= 4 threads and 1000 for more than 4 threads."* So my "17s for `-s tiny` on 8 threads against the 1 minute claimed" was 1000 iterations measured against a figure for 100. Ten times the work, quoted as if it were the same test. The text now states that coupling explicitly, since it makes wall-clock counter-intuitive — crossing the 4-thread boundary can make a run take *longer*, not shorter — and recommends pinning `-iters` when comparing.

**`-s large`.** Agreed that "several hours" overstates it, though I did not want to put your ~20 minutes in as a bare figure without support, and the interleaved CI log does not let me attribute a duration to `-s l` alone. What I can support is the bound: on run `34036616304` the Linux `Script` step ran 14:20:30 → 14:42:40, about twenty minutes for six build modes *plus* the whole teensy-to-large sequence. The text now says that, which is both defensible and enough to retire "several hours".

**Platforms.** Both fixed: the run-time note now says x86-64 or ARM, and the signal list says Unix-like (Linux, *BSD including FreeBSD, macOS) rather than Linux/macOS, since the `sigaction` path covers every non-MinGW build.

## Still outstanding

Two larger asks from your review that I have deliberately not attempted here, rather than half-doing them:

- **The v21/v21.0.1/v21.0.2 feature sweep**, from the commit ranges you linked. That is the right source and I would rather work through it as its own change than append it to a version-stamp fix.
- **The Debian manpage.** Worth pulling from, same reasoning.

Happy to take either on next; say which you would rather have first. Also rebased onto current `main` (65 behind).

help.txt
```diff
@@ -114,24 +114,37 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
  -s {anything other than the mnemonics below}: Self-test, user must also supply exponent (via -m or -f),
 	iteration number and (optionally) FFT length to use.
 
+	Run-times below are indicative only. They vary widely with core count, clock and memory
+	bandwidth, and a modern multi-core ARM system is comparable to an x86-64 one. Note also
+	that more cores does not simply mean less wall-clock time: the default iteration count is
+	100 for <= 4 threads but 1000 for more, so crossing that boundary asks for ten times the
+	work and a run can take longer than the same test on fewer cores. Compare like with like
+	by pinning -iters rather than by relying on the defaults.
```
tdulcet: I would remove the recommendation to pin the `-iters` option, as self-tests that are too fast could produce bad results, which is the reason for that default.

help.txt
```diff
@@ -114,24 +114,37 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
...
 *                                                                                             *
 ***********************************************************************************************
 
- -s large	Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering
- -s l		FFT lengths 8M to 60M. This will take around 10 hours on a fast CPU.
+ -s large	Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering
+ -s l		FFT lengths 16M to 60M. Long documented as "several hours", which badly overstates
+		it on current hardware: a GitHub Actions runner gets through the whole teensy-to-large
+		sequence, plus six build modes, inside about twenty minutes. Still the longest of
+		these tiers, and the most sensitive to memory bandwidth.
```
tdulcet: I would remove this addition, as it is inaccurate and some users may not understand the reference to GitHub Actions. I would also remove the reference to the previous "several hours" runtime estimate.

tdulcet (changes requested):
Thanks for your work on this.

I believe the self-test runtime estimates might still need a more complete overhaul, as they are both inaccurate and inconsistent.

Maybe we should provide an approximate range for the runtime, for example on a slow system single threaded and on a fast system multithreaded. If you are up for it, I recall you have both a Raspberry Pi and a Intel AVX2 system, so I would recommend running the entire self-test suit (teensy through egregious) on those systems with 1, 4 and all CPU threads, and then recording the runtimes. This along with the information provided by Ken above (see https://github.com/primesearch/Mlucas/pull/111#issuecomment-5145315892) and the results from GitHub Actions (for teensy through large) might provide us enough data to generate a more accurate estimate of the range for the runtime.

MarkRose:
Done — measured on both machines you suggested, and combined with Ken's figures.

## What was measured

Full sweep on a **Raspberry Pi 4B** (4 cores, ASIMD, `no_gmp`) and an **i9-10885H** (8c/16t, AVX2), both from a clean `main` at 8728dc5, at 1 / 4 / all threads:

| tier | i9 1 core | i9 4 | i9 16 | Pi 1 core | Pi 4 = all |
| --- | --- | --- | --- | --- | --- |
| teensy | 0.3 s | 0.2 s | 1.9 s | 1.5 s | 1.0 s |
| tiny | 5.1 s | 2.0 s | 15.7 s | 47.2 s | 17.4 s |
| small | 64.8 s | 24.5 s | 165.4 s | 670.6 s | 284.2 s |
| medium | 756 s | 345 s | 4176 s | — | 3628 s |

(The 16-thread column is 1000 iterations, the rest 100 — that is the default, and it is the point of the first bullet below.)

I stopped the Pi at `small` plus the one `medium` run: it is ~10x the x86 at the tiers where both ran, so `large` and up would have taken days there and `huge`/`egregious` do not fit in its 3.8 GB at all. The x86 box is capped after `large` for the same reason — `egregious` alone projects to ~25 GB and many hours, and Ken has already measured both of those tiers on two systems.

## What went into help.txt

The table is now stated for **one core at the default 100 iterations**, because that is the only configuration that compares cleanly between machines, and it is stated as an order of magnitude rather than to the second. Two x86-64 laptop CPUs several generations apart agree within about 2x; the last three rows rest on Ken's i5-1035G1 alone and the text says so.

```
  -s         FFT lengths      one core, 100 iters
  teensy     1K - 15K         a second or less
  tiny       16K - 240K       5 - 10 seconds
  small      256K - 1.875M    1 - 2 minutes
  medium     2M - 15M         13 - 21 minutes
  large      16M - 60M        about an hour
  huge       64M - 240M       about 5 hours
  egregious  256M - 512M      about 14 hours
```

plus "a Raspberry Pi 4B measured roughly 10x these figures".

## Three things the measurements turned up

- **Crossing 4 threads multiplies the work by ten.** The iteration default goes 100 -> 1000 there, and run time is linear in iteration count — your data settles that independently, medium 1-core at 21.07 min for 100 iters against 214.2 min for 1000, a couple of percent off exactly linear. So `-s small` is 25 s on 4 cores and **165 s on 16**: the bigger run is slower in wall-clock because it is doing ten times the work.
- **Extra cores do not pay off proportionally.** 1 -> 4 cores measured **2.19x** here; Ken measured **2.25x** on completely different hardware. Against a theoretical 4x, those agreeing that closely is worth stating in the docs.
- **A high thread count tests *less*, not just faster.** The small FFT lengths cannot use many threads and drop out: `-s small` yields 127 residues on 1 or 4 cores but **116** on 16, and `-s teensy` 23 against 16. That seems worth a sentence, since anyone self-testing to populate `mlucas.cfg` rather than to benchmark is better off at 4 cores.

## Two figures that were simply wrong

`huge` and `egregious` both said "several days on a fast CPU". That holds only above 4 threads, where the 1000-iteration default applies; at the default on one core they are about 5 and 14 hours. Both entries now say which regime is which.

One caveat on my own numbers, so nobody over-reads them: the peak-RSS figures I collected are cumulative (`getrusage(RUSAGE_CHILDREN).ru_maxrss` is a running maximum), so I have not put any memory figures in help.txt. And since performance work is coming, I deliberately kept the table to orders of magnitude — the sweep script is reusable for a re-measure afterwards.

MarkRose:
The 16-thread runs finished, and they change the advice rather than just refining a number — so this is a second update rather than a tidy-up.

Normalising the iteration counts so equal work is compared (the sweep's 16-thread runs are 1000 iterations, the rest 100), cost per 100 iterations on the i9-10885H — **8 physical cores, 16 threads**:

| tier | 1 thread | 4 threads | 16 threads | 1T/4T | 1T/16T |
| --- | --- | --- | --- | --- | --- |
| teensy | 0.3 s | 0.2 s | 0.2 s | 1.50x | 1.58x |
| tiny | 5.1 s | 2.0 s | 1.6 s | 2.55x | 3.25x |
| small | 64.8 s | 24.5 s | 16.5 s | 2.64x | 3.92x |
| medium | 756.1 s | 345.0 s | **417.6 s** | 2.19x | 1.81x |
| large | 2225.8 s | 1068.1 s | **1233.8 s** | 2.08x | 1.80x |

**At medium and large, 16 threads is slower than 4** — 21% and 15% slower, for identical work. Not merely sublinear; negative. And those are the tiers users actually run.

So help.txt no longer just says scaling is imperfect, which gave the reader nothing to act on. It now carries that table and says plainly that 4 threads is a good default for those tiers unless you have measured otherwise on your own hardware. That also lines up with your #117 observation that Mlucas peaks at ~1 core per worker, and with Ken's that single-threaded wins outright at small FFT lengths.

One more correction while I was here: the residue-count drop at high thread counts is confined to the small tiers. `-s large` reports 63 residues at 1, 4 and 16 threads alike, because 16M-60M FFTs can use the threads. The text now says so, rather than leaving the impression it applies everywhere.

Two things I want to be explicit about rather than let the table imply more than it should:

- **This is one machine.** The 2.25x cross-check at 4 threads is Ken's, on quite different hardware, but the 4-vs-16 inversion has only been seen here. The text names the machine and invites the reader to measure their own.
- **16 threads here is 8 cores plus SMT**, so the inversion could be SMT contention rather than thread scaling as such. Distinguishing those needs an 8-thread run pinned to physical cores, which I have not done — it belongs with the scaling work in #117, not in a docs PR.

The run-time table itself is unchanged from the last update apart from `large`, which is now bracketed by two machines (37 min here, 59 min on Ken's i5) and reads "35 - 60 minutes".

help.txt
```diff
@@ -109,35 +109,88 @@ Supported command-line arguments:
...
+* -s m      FFT lengths from 2M to 15M.                                                       *
 *                                                                                             *
 ***********************************************************************************************
 
- -s large	Runs 100 or 1000-iteration self-tests on a set of 24 Mersenne exponents, covering
- -s l		FFT lengths 8M to 60M. This will take around 10 hours on a fast CPU.
+ -s large	Runs 100 or 1000-iteration self-tests on a set of 16 Mersenne exponents, covering
+ -s l		FFT lengths 16M to 60M. Long documented as "several hours", which overstates it on
+		current hardware - see the table above - though it is the longest of the tiers a
+		normal user would run, and the most sensitive to memory bandwidth.
```
tdulcet: I would remove this addition:
```suggestion
 -s l		FFT lengths 16M to 60M.
```
Same for the huge self-test below.

help.txt
```diff
@@ -109,35 +109,88 @@ Supported command-line arguments:
...
+	  teensy     1K - 15K         a second or less
+	  tiny       16K - 240K       5 - 10 seconds
+	  small      256K - 1.875M    1 - 2 minutes
+	  medium     2M - 15M         13 - 21 minutes
+	  large      16M - 60M        35 - 60 minutes
+	  huge       64M - 240M       about 5 hours
+	  egregious  256M - 512M      about 14 hours
+
+	A small ARM core is far slower than that - a Raspberry Pi 4B measured roughly 10x the x86
+	figures at the tiers where both were run - so a Pi is a poor choice past 'small'.
```
tdulcet: I would remove the advice that "a Pi is a poor choice past 'small'".  Many users to run Mlucas on RPis for production wavefront work, which requires the medium self-test. The RPis are slow, but they can still complete wavefront assignments well before they expire.

help.txt
```diff
@@ -109,35 +109,88 @@ Supported command-line arguments:
...
+
+	Two things make wall-clock time counter-intuitive, and both are measured rather than
+	supposed:
```
tdulcet: Maybe make this sound less like AI:
```suggestion
	Two things make wall-clock time counter-intuitive:
```

help.txt
```diff
@@ -109,35 +109,88 @@ Supported command-line arguments:
...
+
+	 - Crossing 4 threads multiplies the work by ten, because the default iteration count is
+	   100 for <= 4 threads and 1000 for more, and run time is linear in the iteration count
+	   to within a couple of percent. '-s small' takes 25 seconds on 4 cores and 165 seconds
+	   on 16 - the 16-core run is slower in wall-clock terms because it is doing ten times as
+	   much work. Pin -iters when comparing thread counts instead of relying on the defaults.
```
tdulcet: I would remove this advice to pin the iteration count, as it would produce bad results with higher thread counts. The benchmark script already handles this correctly. See #246.
```suggestion
	   much work.
```

help.txt
```diff
@@ -109,35 +109,88 @@ Supported command-line arguments:
...
+	   4 threads is 2.1 - 2.6x one thread rather than 4x, and was independently measured at
+	   2.25x on quite different hardware. Past that it inverts: at medium and large - the
+	   sizes users actually run - 16 threads is SLOWER than 4. Unless you have measured
+	   otherwise on your own hardware, 4 threads is a good default for these tiers.
+
+	A high thread count also tests *less*, not merely faster, because the smaller FFT lengths
+	cannot use many threads and drop out of the run: '-s small' reports 127 residues on 1 or 4
+	cores but 116 on 16, and '-s teensy' 23 against 16. If the aim is to populate mlucas.cfg
+	rather than to benchmark, 4 cores is the better choice. ('-s large' is not affected: 63
+	residues at every thread count, since 16M-60M FFTs can use the threads.)
```
tdulcet: I would remove (or significantly trim) this paragraph, as the teensy/tiny/small FFT lengths are not used for any production work and they are fast enough that higher thread counts are not needed anyway.

tdulcet (changes requested):
Thanks for completing those benchmarks. The numbers look much better now. I just have some minor feedback on the wording.

MarkRose:
All five done.

**The Pi advice** — removed, and thank you for the correction. I extrapolated from "10x slower" to "unsuitable", which does not follow: an assignment that finishes well inside its expiry is fine however long it takes, and I had no basis for telling people not to use hardware they are evidently using successfully. The measured 10x stays, since it is useful for setting expectations, but the conclusion is gone.

**Pinning `-iters`** — removed. I had not appreciated it goes wrong at higher thread counts, and if the benchmark script handles it then advice in help.txt is the wrong place for it regardless.

**The residue-count paragraph** — removed. Your reasoning applies cleanly: it only ever affected teensy/tiny/small, none of which are used for production work, and those are fast enough that nobody is reaching for more threads on them anyway. So it was a true fact about an situation nobody is in.

**`-s l` and `-s h`** — trimmed to just the FFT lengths; the table above already gives the times. I also trimmed the same "see the table above" pointer from `-s egregious` for consistency, keeping its `-shift 0` requirement — say if you would rather I left that one alone. Doing so exposed a pre-existing stray doubled "also" in that sentence, which is now fixed too.

**Wording** — took your suggestion verbatim, and applied the same treatment to two more places with the same tics: the introductory paragraph and the sentence concluding the 4-vs-16-thread comparison.

Net for this round: 26 lines removed, 14 added.

MarkRose:
One correction to my own numbers, before this gets reviewed.

I quoted "4 threads is 2.1 - 2.6x one thread rather than 4x" as though it described how Mlucas threads. It does not. Those ratios came from self-test wall clocks, and a self-test runs *every* radix set for the length — the bad ones included — plus the bookkeeping to pick a winner. So the figure averages over candidates rather than reporting the configuration a user ends up running, and quoting it that way understated the code.

Pinned to a single radix set and FFT length on the same machine, 1 -> 4 cores measures about **3x**, against a ceiling nearer 3.5x once the all-core clock drop is accounted for (this box runs ~4.86 GHz on one core and ~4.1-4.4 GHz on four). At that FFT length four cores are essentially at the memory wall — a pinned STREAM copy gives 16.1 GB/s from one core and 17.5 GB/s from two or more, so a single core already saturates most of this laptop's memory bandwidth.

The table itself is unchanged and still correct for what it is: how long each self-test tier takes. What I have changed is the text around it, which now says which of the two things it is measuring, and keeps the 4-thread recommendation scoped to *running self-tests* rather than implying it is a general statement about threading.

Worth being clear about the direction of the error: the honest version makes Mlucas look better than my original text did, not worse.

PR #112, opened 2026-07-16
docs: add docs/ARCHITECTURE.md contributor architecture guide

MarkRose:
Adds `docs/ARCHITECTURE.md` — a contributor-oriented "how it works" tour of Mlucas's main subsystems, since the codebase is dense and nearly comment-only. Docs-only (plus a one-line pointer from `README.md`). *(Originally proposed as `docs/HOW_IT_WORKS.md`; renamed to `ARCHITECTURE.md` in review, and this description updated to match.)*

Sections (all with file:function pointers, source-verified):
1. **Core computation** — balanced-digit / IBDWT representation, the fused forward-FFT / pointwise-square / inverse-FFT / carry pipeline shared by `mers_mod_square()`/`fermat_mod_square()`, Mersenne cyclic vs. Fermat negacyclic (right-angle-transform) convolution, FFT-length/radix-set selection, and threading.
2. **Test types & the `ernstMain()` driver** — worktodo parsing/prefixes and LL / Pépin / PRP / PRP-CF / P-1 dispatch through a single `func_mod_square` pointer.
3. **Error checking** — round-off-error monitoring, the Gerbicz check, Res64/Selfridge–Hurwitz residues.
4. **Self-test & the `mlucas.cfg` system** — how `-s` validates FFT lengths against reference residues and picks the best radix set.
5. **Factoring** — the trial-factoring path and the `mi64` multiword-integer library.
6. **Savefiles / restart** — checkpoint byte format and naming (`p<p>`/`q<p>`/`.s1`/`.s2`/`.s1_prod`).
7. **Further reading** — cross-links to the existing `docs/` topical files (`gerbicz.txt`, `Fermat-testing.md`, `pm1.txt`, `brent-suyama.txt`) and `help.txt`.

A later commit also folds the still-relevant, non-duplicative material from the "Algorithmic Q&A" of Ernst's frozen (v20.1.1-era) README webpage into the existing sections rather than keeping it as a standalone Q&A: §1 gains the maximum FFT length (512M doubles) and largest testable exponent (`2^32 - 65` with a residue shift, higher only with `-shift 0`), an informal Mlucas-vs-FFTW speed note, and the SSE2 → AVX/AVX2 → AVX-512 → Armv8-NEON SIMD throughput scaling; §3 gains why PRP rather than LL matters (it is what enables the Gerbicz check, so non-ECC consumer hardware can run reliably) and the long/medium/short/hiacc DWT-multiplier chain-length accuracy modes with their ROE-driven escalation and restart-at-next-larger-FFT-length behaviour. The version-specific KNL console dump and two unverifiable claims from that source were dropped.

Meant to orient a new contributor in ~20 minutes rather than serve as exhaustive API docs.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -142,7 +142,30 @@ keeps expected round-off error around the ~0.25 target for that transform
 length. This is the concrete link between "bits per word" and "which FFT
 length can safely test exponent p".
 
+The largest FFT length Mlucas supports is 512M doubles
+(`MAX_FFT_LENGTH_IN_K = 524288`, `src/Mdata.h`). Because the residue-shift
+bookkeeping (`mi64_shlc`) is limited to a 32-bit shift count, the largest
+exponent testable *with* a nonzero residue shift is `p = 4,294,967,231 =
+2^32 - 65` — above that, `Mlucas.c` aborts with "Exponents this large do
+not support residue shift! Please run with '-shift 0'." Larger exponents
+(up to roughly 8.9 billion, the maximum-recommended-exponent this longest
+FFT length supports) require `-shift 0`.
+
+Not formally benchmarked against FFTW, but informal comparisons (see
+Ernst's README Q&A, linked below) put Mlucas's hand-tuned FFT at roughly
+2x FFTW's speed for the large real-vector lengths Mersenne searching cares
+about (≥128K doubles, K=1024) on comparable hardware.
+
 ### Threading
+
+Independent of the multithreading described below, each worker's inner
+FFT/carry loops are also hand-vectorized: the generic C FFT code is
+augmented with inline assembly that roughly doubles per-cycle throughput
+over scalar C on SSE2 (roughly Opteron through Core2), nearly doubles it
+again on AVX/AVX2 (256-bit registers, 4 doubles/vector), and again roughly
+doubles it on AVX-512 (memory-bandwidth permitting — often not the case on
+modern manycore systems); there's also dedicated Armv8 128-bit-SIMD
+(NEON/ASIMD) inline assembly, as used by Apple Silicon.
 
 Multithreaded builds run the middle FFT passes (steps 1 and 3 above) as
 parallel chunks dispatched through a small custom threadpool,
@@ -277,16 +300,34 @@ round-off, thread bug, etc.):
 carry-propagated digit landed from the nearest integer before rounding —
 see §1. `Mlucas.c` accumulates average (`AME`) and max (`MME`) ROE per
 checkpoint interval and prints them alongside Res64 in the progress line.
-A single very-bad ROE doesn't necessarily kill the
-run — it can trigger a switch to a shorter/safer carry chain
-(`USE_SHORT_CY_CHAIN`) or a Gerbicz-check-driven rollback; a *pattern* of
-bad ROEs at a given FFT length is a hint that the length is too small for
-the exponent (`Mlucas.c` compares against
-`get_default_fft_length`/`given_N_get_maxP`).
+
+The carry step itself runs at one of four "DWT-multipliers chain length"
+accuracy settings — `long`, `medium`, `short`, `hiacc`, in increasing
+order of numerical accuracy and decreasing speed (`USE_SHORT_CY_CHAIN`,
+`arr_sml[]` in `mers_mod_square.c`) — and the choice actually made is
+printed at startup as "Initial DWT-multipliers chain length = [...] in
+carry step." The initial setting is chosen from how close the exponent
+being tested is to the current FFT length's maximum supported exponent
+(the `exp_ratio` thresholds in `Mlucas.c`): runs pushing closer to that
+ceiling start out at a more accurate (shorter) chain. A single
+dangerously-high ROE (the 0.4375 threshold from §1) doesn't necessarily
+kill the run — `mers_mod_square()` first bumps `USE_SHORT_CY_CHAIN` to the
+next-more-accurate setting and continues; only if the run is already at
+`hiacc` and *still* hits a dangerous ROE does it return `ERR_ROUNDOFF`, at
+which point `Mlucas.c` restarts from the last checkpoint at the
+next-larger FFT length (`get_nextlarger_fft_length()`) and resets the
+chain length back to `long`. A *pattern* of such escalations at a given
+FFT length is a hint that the length is too small for the exponent
+(`Mlucas.c` compares against `get_default_fft_length`/`given_N_get_maxP`).
 
 **Gerbicz check** (PRP runs, and Pépin runs when memory allows — see
 `DO_GCHECK` in `Mlucas.c`), full math in
-[`docs/gerbicz.txt`](gerbicz.txt). Idea in plain terms: alongside the
+[`docs/gerbicz.txt`](gerbicz.txt). The mathematically rigorous LL test and
+the probabilistic PRP test identify M-primes with comparable practical
+accuracy — PRP's real advantage is that, unlike LL, it supports this
+check, which cheaply catches essentially all residue-integrity errors and
+so lets ordinary non-ECC consumer hardware run with reliability similar to
+expensive ECC server hardware. Idea in plain terms: alongside the
 primary residue `u(t) = a^(2^t)`, maintain a second "checkproduct" `d`,
 updated cheaply every `L` iterations as `d(t+1) = d(t) · u((t+1)·L)`. There's
 a second, more expensive way to compute the same value:
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,18 +1,28 @@
 # How Mlucas works: a contributor's map
 
-Mlucas is old, dense, C code with comments as the primary documentation. This
-page is a guided tour of the main subsystems and where to find them, so a new
-contributor can build a mental model in about 20 minutes instead of grepping
-blind. It intentionally stays conceptual + points-at-code; for the deep math,
-follow the cross-links to `docs/` and to the source comments themselves,
-which are frequently the *only* place a given algorithm is explained.
+Mlucas is Ernst Mayer's life's work: dense, highly-optimized C code refined
+over more than two decades, with comments as the primary documentation. The
+authoritative references remain Ernst's own detailed
+[README webpage](https://mersenneforum.org/mayer/README.html), the hundreds
+of posts he and others have written on
+[mersenneforum.org](https://mersenneforum.org/), the `help.txt` file shipped
+with the source, and the peer-reviewed research papers describing the
+underlying algorithms. This page is a *supplementary* guided tour of the main
+subsystems and where to find them in the code, aimed at getting a new
+contributor to a working mental model in about 20 minutes. It intentionally
+stays conceptual + points-at-code; for the deep math, follow the cross-links
+to `docs/` and to the source comments themselves, which are frequently the
+most detailed place a given algorithm is explained. Please keep this document
+consistent with Ernst's README webpage where they overlap — that page has
+been frozen and predates the v21 release, so this file is the place to note
+anything that has changed since.
 
 Top-level data flow for a normal production run:
 
 ```
-worktodo.ini line
-    -> command-line/self-test dispatch                          [src/Mlucas.c: main(), ~line 3902]
-    -> ernstMain() (Mlucas.c:354): parse the worktodo assignment
+worktodo.txt line
+    -> command-line/self-test dispatch                          [src/Mlucas.c: main()]
+    -> ernstMain() in Mlucas.c: parse the worktodo assignment
          ("Test=", "DoubleCheck=", "PRP=", "Pminus1=", "Fermat,Test=", ...)
     -> pick FFT length + radix set (mlucas.cfg, or run a timed self-test)
     -> set up FFT tables, read/create a savefile
@@ -24,7 +34,7 @@ worktodo.ini line
 
 Note the direction of the call: `main()` (the command-line/self-test driver) is defined
 *later* in the file than `ernstMain()`, and calls into it — but `ernstMain()` is the one
-that actually opens and parses `worktodo.ini` (see §2). It's easy to assume the opposite
+that actually opens and parses `worktodo.txt` (see §2). It's easy to assume the opposite
 from the names alone.
 
 ## 1. The core computation: FFT-based modular squaring
@@ -44,11 +54,11 @@ The p-bit (or 2^m-bit) integer is stored as an array `a[]` of `n` doubles,
 where `n` is the FFT length. Each digit holds a variable number of bits
 so that `n` doesn't have to evenly divide `p`: most digits hold
 `bits_small = p/n` bits, and the rest hold `bits_small + 1` bits, arranged
-so the total is exactly `p` bits (`mers_mod_square.c:635-637`, `base[]`/
+so the total is exactly `p` bits (`mers_mod_square.c`, `base[]`/
 `baseinv[]`). This is Crandall & Fagin's **irrational-base discrete weighted
 transform (IBDWT)**: instead of a fixed power-of-2 base, each digit position
-`j` has its own "weight" `w[j]` (arrays `wt0[]`/`wt1[]`, built at
-`mers_mod_square.c:663-818`) that makes a plain cyclic convolution of the
+`j` has its own "weight" `w[j]` (arrays `wt0[]`/`wt1[]`, built in
+`mers_mod_square.c`) that makes a plain cyclic convolution of the
 weighted digits equal to multiplication mod `N`. Digits are signed
 ("balanced", roughly in `[-base/2, base/2)`) to keep rounding error
 symmetric and minimize the maximum digit magnitude the FFT has to handle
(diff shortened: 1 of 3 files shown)


docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
+# How Mlucas works: a contributor's map
+
+Mlucas is old, dense, C code with comments as the primary documentation. This
+page is a guided tour of the main subsystems and where to find them, so a new
+contributor can build a mental model in about 20 minutes instead of grepping
+blind. It intentionally stays conceptual + points-at-code; for the deep math,
+follow the cross-links to `docs/` and to the source comments themselves,
+which are frequently the *only* place a given algorithm is explained.
```
tdulcet: Claude: This paragraph could be phrased a little more respectfully, instead of using words like "old" and "only". Mlucas was Ernst's life work. The main source of existing documentation is his detailed README webpage: https://mersenneforum.org/mayer/README.html, but there are also hundreds of posts on the forum, the `help.txt` file and several peer reviewed research papers on the algorithms.

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+
+Top-level data flow for a normal production run:
+
+```
+worktodo.ini line
+    -> command-line/self-test dispatch                          [src/Mlucas.c: main(), ~line 3902]
```
tdulcet: I would avoid specifying specific line numbers in this document, as those are very likely to change in the future. The function and filename should be enough to find the definition. For example:
```suggestion
    -> command-line/self-test dispatch                          [src/Mlucas.c: main()]
```

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+page is a guided tour of the main subsystems and where to find them, so a new
+contributor can build a mental model in about 20 minutes instead of grepping
+blind. It intentionally stays conceptual + points-at-code; for the deep math,
+follow the cross-links to `docs/` and to the source comments themselves,
+which are frequently the *only* place a given algorithm is explained.
+
+Top-level data flow for a normal production run:
+
+```
+worktodo.ini line
```
tdulcet: Mlucas now uses the standard `worktodo.txt` file as of v21:
```suggestion
worktodo.txt line
```
This also occurs below.

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+  handled inline in `ernstMain()`'s same squaring loop, then a GCD checks
+  for a factor. If none is found and the assignment calls for it, **stage
+  2** hands off to a separate file, `src/pm1.c` (`pm1_stage2()`,
+  `pm1.c:958`, a big-step/baby-step extension over primes in `(B1, B2]`) —
+  which reuses the very same `func_mod_square` pointer (asserted explicitly
+  at `pm1.c:1046-1048`). See [`docs/pm1.txt`](pm1.txt) and
+  [`docs/brent-suyama.txt`](brent-suyama.txt) for the underlying number
+  theory and worked/known-factor examples (P-1: if `p-1` is `B`-smooth,
+  `a^(p-1) mod N` reveals a nontrivial factor via GCD; Brent-Suyama extends
+  stage 2 to catch factors with one additional large prime factor cheaply),
```
tdulcet: I do not believe the Brent-Suyama extension is actually implemented yet. Please check this.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,401 @@
...
+  which reuses the very same `func_mod_square` pointer (asserted explicitly
+  at `pm1.c:1046-1048`). See [`docs/pm1.txt`](pm1.txt) and
+  [`docs/brent-suyama.txt`](brent-suyama.txt) for the underlying number
+  theory and worked/known-factor examples (P-1: if `p-1` is `B`-smooth,
+  `a^(p-1) mod N` reveals a nontrivial factor via GCD; Brent-Suyama extends
+  stage 2 to catch factors with one additional large prime factor cheaply),
+  and `help.txt` §9 for the operational `-maxalloc`/`-pm1_s2_nbuf`
+  stage-2-memory flags and the `Pminus1=...,B2_start[,known_factors]`
+  stage-2-continuation assignment syntax. Big-integer GCD/modexp needed
+  here comes from `mi64.c` (§5).
```
tdulcet, line 267: Computing the GCD requires the optional GMP library.

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+rather than a fatal abort.
+
+**Res64 / Selfridge-Hurwitz residues.** `Res64` is the low 64 bits of the
+residue; `Res35m1`/`Res36m1` are the residue mod `2^35-1` and mod `2^36-1`
+(cheap extra checksums, computed via `mi64_div_by_scalar64`-style code around
+`Mlucas.c:5100-5130`). These three numbers together are what's compared
+against known-good reference residues during self-test (§4), printed in
+progress/checkpoint output, stored in savefiles, and reported in the final
+JSON result — they're the standard lightweight way GIMPS participants
+cross-check results against each other or PrimeNet without transmitting the
```
tdulcet: Only the Res64 (and Res2048 for PRP tests) are included in the JSON results. The Selfridge-Hurwitz residues are not typically used by other GIMPS software.

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+full (possibly gigabit-scale) residue. Mlucas does **not** implement a
+separate Jacobi-symbol check (see the feature-comparison table in the
+top-level README); Gerbicz is its only in-run correctness check.
+
+## 4. Self-test & the `mlucas.cfg` system
+
+Running with `-s <level>` (see `help.txt` for the `tiny`/`small`/`medium`/
+`large`/`all` presets) drives `main()` through a battery of known
```
tdulcet: Note that these are not all of the supported self-tests.

docs/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,401 @@
...
+that traces back to Richard Crandall's original 1998 F24 "Pepin residue
+file" definition. Fields, in order: a leading test-type byte and
+modulus-type byte; an 8-byte iteration/squaring count; the residue itself
+(`ceil(log2 N / 8)` bytes, endian-independent, renormalized from
+balanced-digit to standard non-negative form, and *un*-shifted — see
+`-shift`, `help.txt` §6, for the per-iteration residue-shift scheme and why
+the shift count is stored and re-applied separately rather than baked into
+the stored residue); the `Res64`/`Res35m1`/`Res36m1` checksums (re-verified
+on read); a variable-length ASCII tail for anything else needed to
+characterize the run (e.g. P-1 stage bounds); the FFT length in use at
```
tdulcet: Contrary to the code comment, that variable-length ASCII tail does not actually exist in the savefiles.

(Note that the [GIMPS status script](https://github.com/tdulcet/AutoPrimeNet/blob/main/gimps_status.py) can be used to easily parse, validate and extract the available information from Mlucas savefiles, which can be useful for debugging.)

tdulcet (changes requested):
Thanks for your work on this new documentation. It should be very helpful for new contributors!

I am not familiar with all of the low level details of Mlucas, but I noted several things that I believe are incorrect. What would you think about naming the document `CONTRIBUTORS.md` or `ARCHITECTURE.md` instead.

Ernst's [README webpage](https://mersenneforum.org/mayer/README.html) (and the subpages) include a lot of information about Mlucas, so if you are up for it, I believe it might be helpful to move that information between this new document and the existing `README.md`. The README webpage has been frozen for posterity, so it is becoming outdated (e.g. it was never updated for v21), which makes it difficult for users to determine what is still current. I believe it would be better to move that information to this repository, where we could keep it updated, and maybe Claude could help with that. Regardless, we should confirm that this new document is at least consistent with the information on that existing website.

MarkRose:
Thanks — all addressed in the latest push:

- Renamed to **`docs/ARCHITECTURE.md`** (via `git mv`, history preserved; README link updated).
- Rewrote the intro to credit Ernst's work and point to his README webpage, the forum, `help.txt`, and the research papers as the authoritative sources (dropped the "old"/"only" framing), with a note to keep this doc consistent with that (frozen, pre-v21) webpage.
- **Stripped all ~45 hardcoded source line numbers** — file + function name only now.
- `worktodo.ini` → `worktodo.txt` throughout.
- Factual corrections you flagged: **Brent-Suyama is not (yet) implemented** (I confirmed there's no such code in `src/`; kept it distinct from the implemented `Suyama_CF_PRP()` cofactor test); **GCD requires the optional GMP library** (noted for both PRP-CF and P-1); JSON result carries only **Res64 (+Res2048 for PRP)**, not the Selfridge-Hurwitz residues; the self-test level list is now marked illustrative; and the savefile **"ASCII tail" was a stale header comment** — corrected, with a pointer to your `gimps_status.py`.

On your larger suggestion: yes — I'm now migrating the content from Ernst's README webpage into the repo (updated for v21) so it stops drifting, as a separate PR. This ARCHITECTURE.md will be the home for the algorithmic material; the user-guide material (install/build/PrimeNet/savefiles) is going into `README.md`.

MarkRose:
Followed up on your suggestion to bring Ernst's README webpage into the repo. The **user-guide** content (install/build/PrimeNet/savefiles/tuning) is now migrated and updated for v21 in #113 (plus docs/PERFORMANCE.md and docs/ANDROID.md). The webpage's **"Algorithmic Q&A"** belonged here, so I've folded its non-duplicative material into this doc (latest push, 98b196c) rather than repeating §1:

- §1 — max FFT length (512M) and largest testable exponent (4,294,967,231 with a residue shift; larger, up to ~8.9B, needs `-shift 0`) verified against `MAX_FFT_LENGTH_IN_K` and the `-shift` check; the informal ~2x-FFTW speed note; and the SSE2/AVX/AVX2/AVX-512 + Armv8-NEON SIMD throughput scaling.
- §3 — the *why* of PRP-over-LL (it enables the Gerbicz check → non-ECC hardware runs reliably), and the `long/medium/short/hiacc` accuracy modes with the ROE-driven escalation + restart-at-next-larger-FFT behavior.

I dropped the version-specific KNL console dump and the claims I couldn't confirm against current source (the old "1M-iteration cap" and "since 1997"). Between this and #113, Ernst's webpage content is now in the repo where it can be kept current.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,474 @@
+# How Mlucas works: a contributor's map
+
+Mlucas is Ernst Mayer's life's work: dense, highly-optimized C code refined
+over more than two decades, with comments as the primary documentation. The
+authoritative references remain Ernst's own detailed
+[README webpage](https://mersenneforum.org/mayer/README.html), the hundreds
+of posts he and others have written on
+[mersenneforum.org](https://mersenneforum.org/), the `help.txt` file shipped
+with the source, and the peer-reviewed research papers describing the
+underlying algorithms. This page is a *supplementary* guided tour of the main
```
tdulcet: Maybe list those research papers somewhere. There are several linked on his README webpage. Here are some examples of ones authored by him (from [his resume](https://mersenneforum.org/mayer/resume.html)):
* https://mersenneforum.org/mayer/F24.pdf
* https://arxiv.org/pdf/1303.0328
* https://magazine.odroid.com/article/prime-number-discovery-use-odroid-c2-make-mathematical-history/

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,474 @@
...
+(`mers_thread_data_t`, `ferm_thread_data_t`) carrying the roots-of-unity
+tables, block indices, etc. each worker needs, and dispatch chunks via
+`mers_process_chunk()`/`fermat_process_chunk()`.
+
+CPU affinity is handled in the same file: if built with `hwloc`, threads are
+pinned via `hwloc_set_cpubind()` using logical-CPU
+objects looked up from the runtime topology; without hwloc it falls back to
+raw `sched_setaffinity()`/`CPU_SET` on Linux, `cpuset_setaffinity()` on
+FreeBSD, or is left to the OS on OpenBSD (all in `threadpool.c`). hwloc is
+optional at build time per the top-level README.
```
tdulcet: This sections needs updating after #110.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,474 @@
...
+**`mi64.c`/`mi64.h`** (~8800 lines) is the general multi-precision integer
+library underlying all of this: arrays of `uint64` limbs with add/sub
+(`mi64_add`), multiply, division (`mi64_div`,
+`mi64_div_by_scalar64`), modular exponentiation
+(`mi64_modpow_lr`, `mi64_twopmodq`/`mi64_twopmodq_qmmp`), and the
+bit/scalar utilities the rest of the codebase
+leans on for anything that doesn't fit in a native 64/128-bit integer.
+Beyond trial factoring, this same library backs P-1 stage 2 arithmetic and
+the GCD step used by cofactor-PRP (which, as noted in §2, requires GMP at
+build time).
```
tdulcet: Maybe briefly clarify how the GCD requires both mi64 and the GMP library.

tdulcet (changes requested):

MarkRose:
All three inline items done.

**Papers.** Added, each tied to the section it explains rather than as a bare list. One correction worth flagging: I first wrote the F24 link up from its filename and got the title wrong. Pulling the text out of the PDF gives Crandall, Mayer & Papadopoulos, *The twenty-fourth Fermat number is composite*, Math. Comp. **72**(243), 1555–1572 (2002) — so it is co-authored, peer-reviewed, and it is the source for the Pépin test and the floating-point DWT the code still uses. I pointed at the resume page for the wider list rather than mirroring it here, since that list will keep growing.

**Affinity (§ line 186).** You are right that #110 made this stale — it claimed everything outside Linux was "left to the OS". It now names the four real paths in `threadpool.c`: `sched_setaffinity()` on Linux, `cpuset_setaffinity()` on FreeBSD, `SetThreadGroupAffinity()` on Windows/MinGW (the group form, so machines with more than 64 logical CPUs stay addressable), and `thread_policy_set()` with `THREAD_AFFINITY_POLICY` on macOS.

**GCD (§ line 417).** Expanded to say which library does what, because "requires GMP" on its own does not explain the division of labour. Checked against `gcd()` in `Mlucas.c` rather than written from memory — and just as well, since my first draft invented `mi64_to_mpz`/`mi64_from_mpz` helpers that do not exist. What actually happens: the residue arrives as a plain `mi64` limb array (`uint64 *vec1`, `nlimb`), the bridge is GMPs own `mpz_import()`, and from there the modulus (`mpz_mul_2exp` + `mpz_sub`/`mpz_add`), the `mpz_gcd()` and the known-factor division (`mpz_tdiv_qr`) are all GMP. The answer comes back as a decimal string in `gcd_str`, not as limbs. That is why `INCLUDE_GMP == 0` degrades to `#warning ... No GCDs will be done on p-1 outputs` rather than to a slower path.

## On the bigger ask

> it might be helpful to move that information between this new document and the existing README.md ... Regardless, we should confirm that this new document is at least consistent with the information on that existing website.

Agreed on both, and I think they are separable. The consistency check is in scope for this PR and I will do it next — going through the README webpage section by section against what this document asserts, and fixing anything that disagrees. The migration is a much larger piece of work, and given the webpage is frozen and already missing v21, it deserves its own PR (or several) rather than being folded in here where it would swamp the review.

MarkRose:
Consistency check against the README webpage done, as promised above.

Method: worked through the page section by section against what this document asserts, and against the code where the two could differ. The page has been frozen since before v21 and its own News section is incomplete, so where they disagree I treated the code as authoritative rather than the page.

**Most of it already agrees.** §6's savefile material lines up with the page's "Savefile format and creation" — the redundant `p<p>`/`q<p>` pair, the rename to `p<p>.s1` at the end of a P-1 run, and the reason for it (so a later LL/PRP run does not clobber the stage-1 state, and so a deeper stage-2 continuation can reuse it). `worktodo.txt` is already the v21 name throughout, not `worktodo.ini`. §4's self-test → `mlucas.cfg` flow matches the tuning section.

Two things came out of it:

**1. Checkpoint interval — made concrete.** The document said only "default depends on thread count". The page gives numbers and the code agrees, so it now states them: 10000 for ≤ 4 threads, 100000 for more, overridable via `CheckInterval=` in `mlucas.ini`, which rejects values below 1000.

**2. PRP proofs — the page advertises a feature that does not exist.** Its v21 entry lists "PRP-proof support", but that bullet was never finished; it reads verbatim:

```
o PRP-proof support *** add details ***
```

and I cannot find the feature in the tree. Nothing writes a proof file, no `Cert=` worktype is parsed (`PRPDC=` is handled, but that is a double-check, not a certificate), `help.txt` documents no proof option, and the only trace in the source is a `cert_squarings` field declared in a struct at `Mlucas.c:3950` and never read anywhere. Since someone arriving from the webpage would go looking for it, the document now says plainly that it is absent.

If I have missed it and proof support does exist somewhere, please say — that is the one claim here I would most like a second pair of eyes on, since it is an absence rather than a presence and those are the easy ones to get wrong.

Separately: the manpage import and the v21 feature sweep you raised on #111 are being written up as their own PR rather than folded into either of these, and I will link it here when it is open.

PR #114, opened 2026-07-16
Mfactor: fix checkpoint-savefile corruption and tiny-exponent abort

MarkRose:
While wiring the Mfactor `test_fac` self-test into CI, I found that running an actual Mfactor factoring job aborts. Two independent, pre-existing bugs in the Mfactor savefile/factoring paths, each fixed in its own commit:

## 1. Checkpoint savefile corruption (`write_savefile`/`read_savefile`)

Single-threaded Mfactor runs write a `t{exponent}` checkpoint savefile every 1024 passes and at end-of-pass, and read it back on restart. Both directions were broken, so every run that reached a checkpoint aborted:

- **`write_savefile`** opened the file `"r+"` and interleaved `fgets` reads with `fprintf` writes with **no intervening `fseek`/`fflush`** — undefined behavior (C11 §7.21.5.3p7: input may not be directly followed by output on an update stream without a repositioning call). In practice it wrote fields at the wrong offsets, producing a garbled savefile (duplicated/truncated field names). It also tried to overwrite **variable-length** decimal fields in place (`know` grows from `"0"` to a many-digit value between checkpoints), which cannot work even with correct positioning. Rewritten to read back the run-invariant fields and re-emit the whole file via `init_savefile()`.
- **`read_savefile`** parsed the `bmin`/`bmax` lines with `sscanf(char_addr,"%lf",…)` after locating the `=`, but never advanced past the `=`, so `sscanf` saw `"= 0.000000"` and failed — every restart aborted with *"There were errors reading the savefile"*. Advance past the `=` (as the integer-field parsers already do), and guard a missing `=` instead of scanning through `NULL`.

Verified end-to-end: a from-scratch `-m 127 -bmax 20` run completes and writes a well-formed savefile (`know`/`passnow`/`#Q tried` correctly updated), and a subsequent restart reads it back cleanly and recognizes the completed run.

## 2. Trivial `k=0` candidate aborts tiny-exponent runs

With the default `-bmin 0`, the sieve can present `k = 0`, i.e. the trivial "factor" `q = 2·k·p+1 = 1`. Since 1 divides every M(p), it is flagged as a factor and PRP-tested via `mi64_pprimeF(q,3,…)`, whose Montgomery powering asserts `base < modulus` (`3 < 1` is false) → abort. Trivial to hit, e.g. `./Mfactor -m 23 -bmax 8`. 1 is never a factor, so the trivial `k=0` candidate is skipped. Real runs (`kmin ≥ 1`) never sieve `k=0`, so this is a no-op for them; verified `-m 67 -bmax 20` still finds `q = 193707721` and formerly-aborting tiny cases now exit cleanly.

Both bugs predate the current campaign; the Mfactor checkpoint code was evidently never exercised (note the standing `#error Multithreaded checkpointing stuff needs debug & test!`). This makes single-threaded Mfactor checkpoint/restart actually work.

### DIFF
--- a/src/factor.c
+++ b/src/factor.c
@@ -3770,6 +3770,11 @@ MFACTOR_HELP:
 							{
 								if((res >> l) & 1)	/* If Lth bit = 1, Lth candidate of the inputs is a factor */
 								{
+									// k = 0 yields the trivial "factor" q = 2.k.p+1 = 1, which divides everything. It can
+									// be sieved when the lower factor-bound admits it (e.g. -bmin 0 => kmin 0); skip it, since
+									// 1 is not a factor and PRP-testing it would violate mi64_pprimeF's base-<-modulus
+									// precondition and abort the run (seen with tiny exponents such as -m 23 -bmin 0):
+									if(k_to_try[l] == 0) continue;
 								#ifdef MULTITHREAD
 									pthread_mutex_lock(&mutex_mi64);
 								//	printf("Found Factor: Thread %u locked mutex_mi64 ... ",tid);
--- a/src/factor.c
+++ b/src/factor.c
@@ -3168,6 +3168,12 @@ MFACTOR_HELP:
 			// [ = 272272 or 226304, resp., depending on whether TF_CLASSES = 60 or 4620].
 			// We vectorize the 2nd loop, since each prime therein will hit at most one bit of the sievelet,
 			// i.e. we require no while-loop, only an if(curr_p's startval < bit_len or not) conditional.
+			// The vectorized bit-clearing below cannot handle the 0xFFFFFFFF 'p == curr_p' sentinel start-value
+			// get_startval() sets when the exponent p is itself within the sieving-prime range - only reachable
+			// by tiny exponents (e.g. the '-m 127 -bmax 20' self-test). Fall back to the scalar loop for that
+			// case; the AVX-512 path handles all real (large) exponents, where no sieving prime equals p.
+			const uint32 small_p = (lenP == 1) && (p[0] <= MAX_SIEVING_PRIME);
+			if(!small_p) {
 		// Loop #1:
 			curr_p = p_last_small;
 			for(m = nclear; m < nprime; m++)
@@ -3229,7 +3235,9 @@ MFACTOR_HELP:
 			);
 		/*	}	*/
 
-		#else	/******** Non-SIMD (pre-AVX512) **********/
+			} else	// small_p: the vectorized sieve can't handle the p==curr_p sentinel; use the scalar loop below
+		#endif
+			{	/******** Non-SIMD (pre-AVX512) - also the small-exponent fallback from the AVX-512 path above **********/
 
 		  #ifdef USE_NCQ
 			#warning Using 4-way bit-clear in PerPass_tfSieve.
@@ -3275,7 +3283,7 @@ MFACTOR_HELP:
 				}
 			}
 
-		#endif	// USE_AVX512 ?
+			}	// end of the AVX-512-vs-scalar (incl. small_p fallback) bit-clearing block
 
 //	if(pass==4)printf("\nPass %u: word0 after deep-prime clearing = %16" PRIX64 "\n",pass,bit_map2[0]);
 
--- a/src/factor.c
+++ b/src/factor.c
@@ -4496,11 +4496,13 @@ uint64*kmin, uint64*know, uint64*kmax, uint32*passmin, uint32*passnow, uint32*pa
 			char_addr = strstr(g_in_line, "=");
 			if(!char_addr) {
 				++nerr; fprintf(stderr,"ERROR: Line %d of factoring restart file %s lacks the required = sign!\n",curr_line,fname);
+			} else {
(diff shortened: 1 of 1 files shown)


MarkRose:
Verified the tiny-exponent sieve fix under emulation, since no CI runner reliably has AVX-512 hardware to hit this on.

On current `main`, an AVX-512 `Mfactor` running the very case `ec371d7c`'s comment names — `-m 127 -bmax 20` — dereferences out of bounds in the vector sieve:

```
$ sde64 -ptr-check -skx -- ./Mfactor -m 127 -bmax 20
Pass 0: SDE ERROR: DEREFERENCING BAD MEMORY POINTER
  PC=0x56525e2f3d25  MEMEA=0x5652965db8d4
  vpgatherdd zmm3, k2, dword ptr [rcx+zmm...]
  Function: PerPass_tfSieve
```

That's the `0xFFFFFFFF` sentinel from `get_startval()` going straight into the gather's index vector, exactly as described in the commit.

It is not compiler-specific — gcc 11.4 (ubuntu-22.04's) and gcc 13.3 (ubuntu-24.04's) both hit it, so it isn't a register-allocation artefact:

| tree | compiler | result |
|---|---|---|
| `main` | gcc 11.4 | bad-pointer dereference in `PerPass_tfSieve` |
| `main` | gcc 13.3 | same |
| `main` + this PR's 3 commits | gcc 13.3 | **clean, exit 0** |

With the fix applied the whole `test_fac()` sequence completes, reaching `twopmodq96_q4` and reporting no failed factorizations. Same compiler, same flags, same SDE invocation in every row — the tree is the only variable.

## This needs a rebase

The branch is based on `418f365b`, which predates #193. At that base `makemake.sh mfac avx512` does not link at all:

```
```

because the never-implemented USE_FMADD 100-bit modpow routines are still referenced there. So the PR as it stands cannot be built in the mode where the bug lives. The three commits cherry-pick onto current `main` with no conflicts, and that is what I tested above.

## Why it went unnoticed

An AVX-512 `Mfactor` only reaches this code when it actually executes on AVX-512, and runner CPUs are a lottery: within a single CI run I see Intel Xeon Platinum 8370/8573 (AVX-512 present) alongside AMD EPYC 7763 (absent) and EPYC 9V74 (present in silicon, masked by the hypervisor). A job lands on whichever it draws, so a defect like this surfaces intermittently and looks like flakiness. Running under `sde64 -skx` removes the lottery.

MarkRose:
Rebased onto current `main` — it now builds in AVX-512 mode, which it could not do at the old base. The three commits applied without conflict and the resulting tree is byte-identical to the one I tested above, so the verification stands unchanged.

MarkRose:
Rebased onto current `main` (8728dc5) — this had stopped merging on its own, independently of anything stacked on it.

## The conflict

One line, in `factor.c`. This PR turns the AVX-512 bit-clearing block from a compile-time `#if`/`#else` into a runtime `if(!small_p)`/`else`, so the closing `#endif` becomes a closing `}`. Meanwhile `main` reworded the comment on that same `#endif` when the guard became `#if defined(USE_AVX512) && !defined(USE_IMCI512)`:

```
main:  #endif  // AVX-512 (non-IMCI) sieve ?
here:  }       // end of the AVX-512-vs-scalar (incl. small_p fallback) bit-clearing block
```

Resolved in favour of this branch's `}` — the `#endif` it replaced has already been emitted higher up, right after the `} else`. Nothing else conflicted.

## Verification after the rebase

Builds clean in all three SIMD modes (`avx2`, `avx512`, `nosimd`).

The behaviour this PR is for, single-threaded, which is where `main` dies:

| build | `-m 127 -bmax 20` |
| --- | --- |
| `main` @ 8728dc5 | `ERROR: 'passnow' not found in Line 10 of factoring restart file t127!` — rc=134 |
| this branch, avx2 | `M(127) has 0 factors in range k = [0, 16336320], passes 0-15` — rc=0 |
| this branch, nosimd | same, rc=0 |

Because the conflict sat inside the AVX-512 block, I exercised both sides of the new runtime branch rather than just rebuilding:

- **small_p fallback** — the AVX-512 build under `sde64 -skx`, `-m 127`: rc=0, correct result. This is the path the resolved line closes.
- **vector fast path** — `-m 2976221`, which is above the 1299721 max sieving prime so `small_p` is false: rc=0, `M(2976221) has 0 factors in range k = [0, 49008960]`, 1907383 trial divides. (Worth noting for anyone testing this: `-m 1257787` is *not* good enough, it falls below the sieving bound and still takes the scalar fallback.)

## Stacking

#269 now merges cleanly on top, and the combination fixes the one case #269 could not verify on its own — the `nword` build, which asserts `NTHREADS == 1` and so cannot dodge this bug with `-nthread 4`:

```
#114 + #269, Mfactor_nword -m 521 -bmax 20, single-threaded
M(521) has 0 factors in range k = [0, 16336320], passes 0-15    rc=0
```

That is the whole Mfactor CI loop unblocked: on `main` today every word variant of that step aborts in the savefile path.

MarkRose:
@tdulcet This one should fix most of the failing build jobs.

PR #116, opened 2026-07-16
Fix two multi-length self-test heap-buffer-overflows (residue-array sizing; radix-12 scalar index)

MarkRose:
Two independent heap-buffer-overflows that surface in multi-FFT-length self-tests, each in its own commit. The first is the ASan crash reported on mersenneforum; the second I found while reproducing it under a NOSIMD+ASan build.

## Commit 1 — residue byte-array sized for the first exponent instead of the largest

Reported (arm64):
```
M320851: using FFT length 15K = 15360 8-byte floats ...
Using complex FFT radices 60 8 16
==ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 8
0x... is located 1448 bytes after 38656-byte region  (allocated by realloc in ernstMain)
```

`ernstMain` allocates the packed-byte residue/scratch array `arrtmp` **once** and reuses it for every exponent in the run. In a multi-FFT-length self-test the residue arrays are sized for the largest FFT length used (`maxFFT`), and the largest exponent processed can be as high as `pmax(maxFFT)` — but `arrtmp` was sized from the **first (smallest)** exponent's `p` plus a *"2.5 bytes per float-double word"* floor:
```c
arrtmp_alloc = MAX((p+63)>>2, (uint64)((maxFFT<<10)*2.5)) >> 3;
```
At ~15K and above the true bits-per-word is ~21 (~2.6 bytes/word), so `2.5·words` underestimates. For the **teensy** self-test (`maxFFT = 15K`, first exponent 22679) the floor gives `2.5·15360 = 38400` bytes (≈ the reported 38656-byte region incl. padding), but the range's largest member **M320851** needs `⌈320851/8⌉ = 40107` bytes — overrunning `arrtmp` by ~1.7 KB.

**Fix:** size `arrtmp` for 2× the limb-count of the largest exponent that can run against these arrays — `pmax` at the allocated (max) FFT length via `given_N_get_maxP()`, or the current `p` if the FFT-too-small warn-path admitted a `p` slightly above that `pmax`:
```c
uint32 alloc_n = (maxFFT > kblocks) ? (maxFFT << 10) : n;
arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(alloc_n)) + 63) >> 5;
```
No-op-or-larger for single-length runs (`maxFFT = 0`).

**Verified:** `./Mlucas -s tt -iters 100` now runs the exact reported sequence — `14K passed → M320851 at 15K, radices 60 8 16, 100 iters → 15K passed` — to completion (`arrtmp` 80608 vs 38400 bytes). 8K single-length still reproduces reference residue `85301536E4CA9B11`.

## Commit 2 — radix-12 scalar pass missing `#else j1 = j;`

While reproducing Commit 1 under a NOSIMD+ASan build I hit a **second, unrelated** heap-buffer-overflow in `radix12_di{f,t}_pass1` at NOSIMD radix-12 FFT lengths (e.g. 48K). Root cause: the radix-12 pass loops compute the padded fetch index `j1` as
```c
#ifdef USE_AVX
    j1 = (j & mask02) + br8[j&7];
#elif defined(USE_SSE2)
    j1 = (j & mask01) + br4[j&3];
#endif        // <-- no #else
```
`radix12_ditN_cy_dif1.c` is the **only one of the 42 `radix*_ditN_cy_dif1.c` files** lacking the scalar `#else j1 = j;` (present in both its `dif`/`dit` pass1 loops in every other radix). In a NOSIMD build `j1` was never set from `j`, so it retained stale state and `a[j1 + p*]` ran off the residue array.

**Fix:** add the scalar `#else j1 = j;` to both loops, matching the other 41 radix files. **Verified:** NOSIMD+ASan `-fft 48 -iters 100` completes with reference residue `6B1C76AB5431FDA4`; `-s t` runs past 48K/60K/72K (previously it overran at 48K). SIMD builds are unaffected.

The two commits are independent — happy to split into separate PRs if preferred.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1436,8 +1436,8 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 		// the limb-count [ #limbs for x bits = (x+63)>>6, doubled => (x+63)>>5 ] of the largest exponent that can
 		// run against these arrays: pmax at the allocated (max) FFT length, or the current p if the warn-path
 		// above admitted a p slightly above that pmax:
-		uint32 alloc_n = (maxFFT > kblocks) ? (maxFFT << 10) : n;	// max unpadded FFT length these arrays serve
-		arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(alloc_n)) + 63) >> 5;
+		// Size for the largest exponent supported at the largest FFT length these arrays will serve:
+		arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(maxFFT > kblocks ? maxFFT << 10 : n)) + 63) >> 5;
 		arrtmp = ALLOC_UINT64(arrtmp, arrtmp_alloc);if(!arrtmp ){ sprintf(cbuf, "ERROR: unable to allocate array ARRTMP with %" PRIu64 " limbs in main.\n",arrtmp_alloc); fprintf(stderr,"%s", cbuf);	ASSERT(0,cbuf); }
 
 		// For an n-word main-array, BIGWORD_BITMAP and BIGWORD_NBITS have (n/64) elts each, thus need 1/64 + 1/32 the total
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1427,11 +1427,18 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 			b_uint64_ptr = (uint64*)b; c_uint64_ptr = (uint64*)c; d_uint64_ptr = (uint64*)d; e_uint64_ptr = (uint64*)e;
 		}
 
-		// For this residue (and scratch) byte-array, conservatively figure at least 4 bits per float-double residue word.
-		// For multi-FFT-length self-tests, conservatively figure as many as 20 bits (2.5 bytes) per float-double residue word:
-		// v20: for largest currently supported FFT of 512Mdoubles, i still -barely - fits in a uint32, but 2.5*i does not:
-		arrtmp_alloc = i; arrtmp_alloc = MAX((p+63)>>2, (uint64)(arrtmp_alloc*2.5)) >> 3;	// #limb needed to store p bits = (p+63)>>6, so alloc at least 2x this
-		arrtmp = ALLOC_UINT64(arrtmp, arrtmp_alloc);if(!arrtmp ){ sprintf(cbuf, "ERROR: unable to allocate array ARRTMP with %u bytes in main.\n",i); fprintf(stderr,"%s", cbuf);	ASSERT(0,cbuf); }
+		// This residue/scratch byte-array is allocated once and reused for every exponent in the run. In a
+		// multi-FFT-length self-test the arrays are sized for the largest FFT length used (maxFFT), and the
+		// largest exponent processed can be as high as pmax(maxFFT) - not the current (first, smallest) p. The
+		// old "2.5 bytes per float-double word" heuristic underestimated the storage: max bits/word is ~21
+		// (~2.6 bytes) at these lengths, so a large exponent near pmax overran arrtmp (heap-buffer-overflow
+		// writing e.g. M320851's 40107-byte residue into a 38400-byte arrtmp sized for a 15K FFT). Size for 2x
+		// the limb-count [ #limbs for x bits = (x+63)>>6, doubled => (x+63)>>5 ] of the largest exponent that can
+		// run against these arrays: pmax at the allocated (max) FFT length, or the current p if the warn-path
+		// above admitted a p slightly above that pmax:
+		uint32 alloc_n = (maxFFT > kblocks) ? (maxFFT << 10) : n;	// max unpadded FFT length these arrays serve
+		arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(alloc_n)) + 63) >> 5;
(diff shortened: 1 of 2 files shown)


MarkRose:
Added a second commit for the radix-12 overflow I flagged above — and the root cause turned out to be **not** the array-padding guess in my note. It is a plain missing scalar case: `radix12_ditN_cy_dif1.c` computes the padded index `j1` under `#ifdef USE_AVX / #elif USE_SSE2` but, uniquely among all 42 `radix*_ditN_cy_dif1.c` files, omitted the `#else j1 = j;` in **both** `radix12_dif_pass1` and `radix12_dit_pass1`. So NOSIMD builds ran with a stale `j1` and `a[j1+p*]` overran the residue array (ASan heap-buffer-overflow at any NOSIMD radix-12 FFT length, e.g. 48K in `-s t`).

Fix is the one-line-each scalar `#else j1 = j;` in both loops, matching the other 41 radix files. Verified on the NOSIMD+ASan build: `-fft 48 -iters 100` now completes with reference residue `6B1C76AB5431FDA4`, and `-s t` runs past 48K/60K/72K. SIMD builds are unaffected. (The two commits here are independent — the `arrtmp` sizing fix and this radix-12 scalar fix — happy to split into two PRs if you prefer.)

src/Mlucas.c
```diff
@@ -1427,11 +1427,18 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+		// multi-FFT-length self-test the arrays are sized for the largest FFT length used (maxFFT), and the
+		// largest exponent processed can be as high as pmax(maxFFT) - not the current (first, smallest) p. The
+		// old "2.5 bytes per float-double word" heuristic underestimated the storage: max bits/word is ~21
+		// (~2.6 bytes) at these lengths, so a large exponent near pmax overran arrtmp (heap-buffer-overflow
+		// writing e.g. M320851's 40107-byte residue into a 38400-byte arrtmp sized for a 15K FFT). Size for 2x
+		// the limb-count [ #limbs for x bits = (x+63)>>6, doubled => (x+63)>>5 ] of the largest exponent that can
+		// run against these arrays: pmax at the allocated (max) FFT length, or the current p if the warn-path
+		// above admitted a p slightly above that pmax:
+		uint32 alloc_n = (maxFFT > kblocks) ? (maxFFT << 10) : n;	// max unpadded FFT length these arrays serve
+		arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(alloc_n)) + 63) >> 5;
```
tdulcet: The `alloc_n` variable is only used once:

```suggestion
		arrtmp_alloc = (MAX((uint64)p, given_N_get_maxP(maxFFT > kblocks ? maxFFT << 10 : n)) + 63) >> 5;
```

tdulcet (changes requested):
Thanks for these quick fixes. Were you able to confirm that the new heuristic works for the full range of supported FFT lengths?

MarkRose:
Thanks for the thorough multi-toolchain testing! Good news: **none of those radix-set "failures" are caused by this PR** — they are pre-existing, benign self-test skips, and this PR is in fact what lets the teensy test run to completion without crashing.

Breaking down the log categories (matches on both your macOS gcc/clang runs and my x86 reproduction):

1. **"CY routines with radix < 16 do not support shifted residues!"** (the large majority — 29/33 on nosimd, 4/? on asimd). The self-test starts each length from a *random* nonzero residue-shift, and the small-leading-radix carry routines (radix 5/6/7/8/9/10/14/15) dont implement shifted residues, so those sets are skipped. By design, and unrelated to anything here. (nosimd shows more than asimd simply because it offers more small-leading-radix sets per FFT length.)
2. **A few `ERR_INCORRECT_RES64` / `ERR_CARRY` / "Excessive roundoff"** at leading radices 18/22/26/30 (9K/11K/13K/15K). Those are the teensy self-test exponents sitting at ~99.5% of pmax for the length, so a few numerically-marginal radix sets exceed the roundoff threshold and are correctly rejected. Also pre-existing FFT-accuracy behavior, not touched by this PR.

To confirm its not a regression, I built **`main`** the same way (nosimd, 1 core) and ran `-s tt`: it shows the *identical* radix<16 skips at 1K–7K and then **segfaults at the radix-12 set** (`Using complex FFT radices 12 32 16`) — i.e. the exact heap-buffer-overflow this PR fixes. With this PR the run instead completes cleanly. No radix-12-leading set fails, and none of the wrong-`Res64` cases are at a size this PR touches.

So the fixes here are doing their job (no crash, correct residues where the radix set is accurate enough); the remaining skips are the same ones youd see on `main` at these tiny FFT lengths. If the *volume* of small-radix skips is bothersome, reducing it (e.g. running those sets at shift 0) would be a separate self-test-ergonomics change, not part of this overflow fix. Really appreciate you also confirming the medium test completes segfault-free on the asimd builds.

MarkRose:
Done — inlined `alloc_n` in 6eba6ff.

On the FFT-range question: this isn't a per-length heuristic — it sizes `arrtmp` from `given_N_get_maxP()` at the **largest** FFT length the run will allocate (`maxFFT`), so by construction it covers the biggest residue any length in a multi-length self-test can produce. I verified the teensy run (`-s tt`) now completes with the 14K→M320851@15K set passing, and that single-length runs (e.g. 8K) are byte-for-byte unchanged. The full CI self-test matrix (tt/t/s/m/l) will exercise the remaining ranges.

tdulcet (approved):
Thanks, this looks good to me. I will leave the PR open for a few days in case anyone else wants to review.

Issue #99, opened 2026-07-15
Allow using a random shift with FFT lengths larger than 256M and when testing Fermat numbers


tdulcet:
Currently using FFT lengths larger than 256M requires providing the `-shift 0` option to force a zero shift. For more information, see: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L1250-L1255 https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/mi64.c#L416-L418 It looks like a datatype just needs to be increased from a 32 to 64 bit integer: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/help.txt#L262-L263

Testing all Fermat numbers also requires providing the `-shift 0` option. For more information, see: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/docs/Fermat-testing.md?plain=1#L99-L107

We should remove these limitations, which could increase confidence in the final result, especially without support for generating proof files (#95).

PR #122, opened 2026-07-19
Fix Pépin+Gerbicz-check runs being unable to resume from their savefile

MarkRose:
## Summary

A Pépin test (Fermat primality) run with the Gerbicz check enabled cannot resume from its own savefile — any restart aborts:

```
ERROR: Function ernstMain, at line 1666 of file ../src/Mlucas.c
Assertion 'RES_SHIFT == 0ull' failed: Shifted residues unsupported for Pépin test with Gerbicz check!
```

## Cause

At run-from-scratch, a primality test seeds a random residue shift (the v18 shifted-residue error-detection scheme). But a **non-interactive Pépin test with the Gerbicz check** does not support shifted residues — the restart path asserts `RES_SHIFT == 0`. The from-scratch path randomizes the shift anyway, so the run:

1. picks a random shift (e.g. `RES_SHIFT = 745425` by iter 1000),
2. evolves and checkpoints it into the savefile,
3. and then aborts on the assertion above the moment it is restarted.

This affects **every** Pépin+G-check run — e.g. a `Fermat,Test=<n>` worktodo assignment. It reproduces on the first checkpoint-and-restart and is independent of how the run is stopped (SIGINT/SIGTERM or a hard `kill -9`), so it is not specific to the interrupt-savefile path. Mersenne PRP is unaffected (the assert is Fermat-Pépin-only), and the interactive `-f` timing-test mode is deliberately excluded (it legitimately allows a shift).

## Fix

For the non-interactive Pépin+G-check case, force `RES_SHIFT = 0` at run-from-scratch rather than randomizing it — matching what the restart path already requires:

```c
if(MODULUS_TYPE == MODULUS_TYPE_FERMAT && TEST_TYPE == TEST_TYPE_PRIMALITY && DO_GCHECK && !INTERACT) {
    RES_SHIFT = 0ull;
} else if(RES_SHIFT == -1ull || RES_SHIFT >= p) {
    ...
```

## Verification

On an avx2 build, a from-scratch `Fermat,Test=20` run:

- **Before:** initial shift randomized to a nonzero value, checkpoints it, and every resume aborts on the `RES_SHIFT == 0ull` assertion.
- **After:** `initial residue shift count = 0`, and it resumes correctly from its checkpoint with clean forward progress:

```
F20: using FFT length 64K ... initial residue shift count = 0
F20 Iter# = 10000 ... Residue shift count = 0.
Restarting F20 at iteration = 10000, residue shift count = 0.
F20 Iter# = 20000 ... Residue shift count = 0.
... 30000 ... 40000 ... 50000 ... 60000 ...
```

Found while validating #100/#108 (savefile-on-interrupt) across worktypes.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1640,6 +1640,15 @@ READ_RESTART_FILE:
 		if(TEST_TYPE == TEST_TYPE_PM1) {
 			ASSERT(RES_SHIFT == 0ull, "Shifted residues unsupported for p-1!\n");
 			RES_SHIFT = 0ull; a[0] = iseed;
+		} else if(MODULUS_TYPE == MODULUS_TYPE_FERMAT && TEST_TYPE == TEST_TYPE_PRIMALITY && DO_GCHECK && !INTERACT) {
+			// A Pépin test with Gerbicz check does not support shifted residues (the random-shift
+			// error-detection scheme is incompatible with the G-check residue-product scheme), and the
+			// resume path below asserts RES_SHIFT==0. Randomizing the shift here (as done for the other
+			// primality tests) would leave a nonzero shift in the savefile and make the run un-resumable
+			// from its own checkpoint (see #99). So force shift 0 for the non-interactive Pépin+G-check
+			// case; as in the p-1 branch above, a zero shift just means the seed goes straight into a[0]
+			// (no need for the random-shift selection or shift_word() carry-in of the else-branch below):
+			RES_SHIFT = 0ull; a[0] = iseed;
 		} else {
 			// Apply initial-residue shift - if user has not set one via cmd-line or current value >= p, randomly choose a value in [0,p).
 			// [Note that the RNG is inited as part of the standard program-start-sequence, via function host_init().]


src/Mlucas.c
```diff
@@ -1643,7 +1643,14 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
 			// Apply initial-residue shift - if user has not set one via cmd-line or current value >= p, randomly choose a value in [0,p).
 			// [Note that the RNG is inited as part of the standard program-start-sequence, via function host_init().]
-			if(RES_SHIFT == -1ull || RES_SHIFT >= p) {	// Since 0 a valid value of RES_SHIFT, treat -1 as "uninited"
+			if(MODULUS_TYPE == MODULUS_TYPE_FERMAT && TEST_TYPE == TEST_TYPE_PRIMALITY && DO_GCHECK && !INTERACT) {
+				// A Pépin test with Gerbicz check does not support shifted residues (the random-shift
+				// error-detection scheme is incompatible with the G-check residue-product scheme), and the
+				// resume path below asserts RES_SHIFT==0. Randomizing the shift here (as done for the other
+				// primality tests) would leave a nonzero shift in the savefile and make the run un-resumable
+				// from its own checkpoint - so force shift 0 for the non-interactive Pépin+G-check case:
+				RES_SHIFT = 0ull;
```
tdulcet: It looks like we would want to move this one layer up (after the `TEST_TYPE == TEST_TYPE_PM1` check on line 1640-1643), as we would not need to run the other code in this else block, if we are not shifting the residue. You may also want to mention #99.

tdulcet (changes requested):
Thanks for the fix. Just a small placement change is needed.

src/Mlucas.c
MarkRose: Moved up as suggested — the Pépin+G-check case is now an `else if` at the top level of the shift-init block (sibling of the `TEST_TYPE_PM1` check), so it skips the random-shift selection and the `shift_word()` carry-in of the general branch. For a zero shift that carry-in just seeds `a[0] = iseed` and touches nothing else (I traced `shift_word()` at `shift==0`: `wt_fwd`, `cos` = 1, `sin` = 0, shift-factor = 1, so `a[0]=iseed`, `a[j2]=0` which the preceding memset already set) — same as the p-1 branch does directly, so the branch now just does `RES_SHIFT = 0; a[0] = iseed;`. Also added a `#99` reference in the comment. (Full end-to-end re-verify on a small Fermat is currently blocked locally by the pre-existing small-Fermat crash that #115 fixes; the F20 path from the original testing is unaffected.)

MarkRose:
Closing in favour of #218, which fixes the same failure by making the combination work rather than by disallowing it.

This PR forces `RES_SHIFT = 0` for Pépin + Gerbicz-check runs so that the resume path stops aborting on

```c
ASSERT(RES_SHIFT == 0ull, "Shifted residues unsupported for Pépin test with Gerbicz check!")
```

That does remove the abort, but at the cost of the shifted-residue protection for every Pépin run, and it leaves the underlying reason the check was failing unaddressed. Worth recording what that reason turned out to be, since it was not what this PR assumed:

* The assertion sat in the **resume** branch of `if(ilo == 0)`, so it never actually prevented the combination. A from-scratch Pépin run picked a random nonzero shift and proceeded — then failed its first G-check, dropped the assignment from the workfile with no result, and only aborted on the *next* invocation. So the visible symptom was "cannot resume", but the real defect was "the check is wrong".
* Three defects underneath it: the `prp_mult` gate applied the Pépin random-bit residue doubling to the G-check auxiliary arrays as well as the main chain; the `2^{s_L}` correction factor is the *Mersenne* answer, which does not carry over because the Fermat-mod shift update ("mod-double **and add a random bit**") does not telescope; and `convert_res_*` applied the main chain's shift and sign to the G-check product, which for Fermat is not self-inverse (`×2^p == −1`) and so corrupted `b[]` on every resume.

With those fixed, nonzero shifts work: F16 passes 6/6 G-checks at shifts 0/1/12345/33333/65432/65535 against an independent GMP oracle, and F20 at the real production intervals (L = 1000, check at 1000000) passes with shift 654321. #218 has the details.

The two branches also conflict textually, so they cannot both land.

PR #124, opened 2026-07-21
types.h aliasing helper + strict-aliasing fixes (from #34)

MarkRose:

This PR covers the **strict-aliasing** group: the `types.h` helper plus every commit that depends on it.

**1. Foundation** — adds an `i64_f64` union and `u64_to_f64` / `f64_to_u64` / `f64_to_i64` inline helpers to `types.h`, giving a well-defined (union type-punning, valid in C99+) way to reinterpret `uint64 <-> double` instead of the `*(uint64*)&x` pointer casts that violate strict aliasing.

**2. Aliasing fixes (33 commits)** — replace the offending pointer-cast reinterprets with the new helpers across `rng_isaac.c`, `mi64.c`, `fermat_mod_square.c`, `mers_mod_square.c`, `qfloat.c`, `util.c`, `dft_macro.c`, `twopmodq80.c`, `radix8_dif_dit_pass.c`, and the `radix{16,28,32,40,48,56,60,64,128,144,160,176,208,224,240,256,352,384,768,960,992,1008,1024,4032}_ditN_cy_dif1.c` DFT files.

Notes:
- All 34 commits are @ldesnogu's, cherry-picked from #34 with original authorship preserved; no content changes.
- Purely warning/UB cleanup — no behavior change. Verified: nosimd build clean, LL self-test Res64 `71E61322CCFB396C` unchanged.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -99,6 +99,36 @@ EWM: nvcc gives 'error: invalid redeclaration of type name "int64_t"' for the ty
 	typedef uint64		uint64_t;
 #endif
 */
+
+/* Type to safely cast uint64 <-> double.
+ * This is used to get rid of gcc strict aliasing violation warnings.
+ */
+typedef union {
+  uint64 u64;
+  double f64;
+} i64_f64;
+
+static inline double u64_to_f64(uint64 u)
+{
+  i64_f64 cvt;
+  cvt.u64 = u;
+  return cvt.f64;
+}
+
+static inline uint64 f64_to_u64(double u)
+{
+  i64_f64 cvt;
+  cvt.f64 = u;
+  return cvt.u64;
+}
+
+static inline int64 f64_to_i64(double u)
+{
+  i64_f64 cvt;
+  cvt.f64 = u;
+  return cvt.u64;
+}
+
 /*******************************************************************************
    Some useful utility macros:
 *******************************************************************************/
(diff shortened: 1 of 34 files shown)


src/types.h
```diff
@@ -99,6 +99,36 @@ EWM: nvcc gives 'error: invalid redeclaration of type name "int64_t"' for the ty
...
+  i64_f64 cvt;
+  cvt.f64 = u;
+  return cvt.u64;
+}
+
+static inline int64 f64_to_i64(double u)
+{
+  i64_f64 cvt;
+  cvt.f64 = u;
+  return cvt.u64;
```
tdulcet: Is it safe to implicitly convert an unsigned to signed int here.

tdulcet (dismissed):
Thanks, this looks great! I just had one question.

Note that this also fixes part of #90, without needing the `memcpy()`.

src/types.h
MarkRose: Good catch. I added a follow-up commit (01852ad) that gives the union an explicit `int64 i64` member and has `f64_to_i64` read that directly, so it is a pure union type-pun to `int64` with no unsigned→signed conversion — well-defined in every C standard. The result bits are identical on all two's-complement platforms, so no behavior change (the only caller, the `qfloat.c` self-test's ULP-distance check, still passes).

tdulcet (approved):
Thanks, this looks great!

PR #125, opened 2026-07-21
radix12: initialize j1 in scalar radix12_dif/dit_pass1 (from #34)

MarkRose:
From @ldesnogu's #34, pulled out on its own because it's a genuine correctness fix rather than a cosmetic warning cleanup, and shouldn't get lost in the larger warnings batch.

**The bug:** in scalar (non-AVX/non-SSE2) builds, `radix12_dif_pass1` and `radix12_dit_pass1` fall through their `j1` selection block with `j1` unset, then use it in `j1 = j1 + ((j1 >> DAT_BITS) << PAD_BITS)`. gcc flags it at `-O1`/`-O3`:
```
radix12_ditN_cy_dif1.c: warning: 'j1' may be used uninitialized [-Wmaybe-uninitialized]
```

**The fix:** add the scalar `#else j1 = j;` — the same initialization every sibling `*_pass1` uses in its scalar branch (verified in radix8/16/24/32/36; radix12 was the lone outlier).

**Scope / severity:** these two `_pass1` routines are currently reachable only from the `test_fft_radix.c` harness (not part of the build) and a dispatch table with no callers; production radix-12 FFT work goes through `radix12_ditN_cy_dif1`, which is unaffected. So this is a correctness/hygiene fix for the standalone pass routines (and silences the warning), not a live-run bug.

Authored by @ldesnogu, cherry-picked from #34 with original authorship preserved. Verified: nosimd build clean, LL self-test Res64 `71E61322CCFB396C` unchanged.

### DIFF
--- a/src/radix12_ditN_cy_dif1.c
+++ b/src/radix12_ditN_cy_dif1.c
@@ -1399,6 +1399,8 @@ void radix12_dif_pass1(double a[], int n)
 		j1 = (j & mask02) + br8[j&7];
 	#elif defined(USE_SSE2)
 		j1 = (j & mask01) + br4[j&3];
+	#else
+		j1 = j; // LD: double-check this additional init is what is expected (comes from radix24)
 	#endif
 		j1 = j1 + ( (j1 >> DAT_BITS) << PAD_BITS );	/* padded-array fetch index is here */
 		j2 = j1+RE_IM_STRIDE;
@@ -1662,6 +1664,8 @@ void radix12_dit_pass1(double a[], int n)
 		j1 = (j & mask02) + br8[j&7];
 	#elif defined(USE_SSE2)
 		j1 = (j & mask01) + br4[j&3];
+	#else
+		j1 = j; // LD: double-check this additional init is what is expected (comes from radix24)
 	#endif
 		j1 = j1 + ( (j1 >> DAT_BITS) << PAD_BITS );	/* padded-array fetch index is here */
 		j2 = j1+RE_IM_STRIDE;
--- a/src/radix12_ditN_cy_dif1.c
+++ b/src/radix12_ditN_cy_dif1.c
@@ -1400,7 +1400,7 @@ void radix12_dif_pass1(double a[], int n)
 	#elif defined(USE_SSE2)
 		j1 = (j & mask01) + br4[j&3];
 	#else
-		j1 = j; // LD: double-check this additional init is what is expected (comes from radix24)
+		j1 = j;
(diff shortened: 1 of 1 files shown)


MarkRose:
Heads-up that this is more than a hygiene/warning fix — the uninitialized `j1` in the scalar `radix12_dif_pass1`/`dit_pass1` is a **live-run crash**, not confined to the (unbuilt) `test_fft_radix.c` harness.

While bringing up the first nosimd/scalar-double build on RISC-V I hit a hard SIGSEGV in a plain self-test. Building `makemake.sh nosimd` and running `./Mlucas -s t` crashes at the leading-radix-12 self-test:

```
Program received signal SIGSEGV, Segmentation fault.
#0  radix12_dif_pass1 (a=..., n=<optimized out>) at radix12_ditN_cy_dif1.c
#2  ernstMain (mod_type=1, test_type=1, exponent=1001467, fft_length=48, radix_set=3, ...) at Mlucas.c:1980
#3  main (...) at Mlucas.c:4673
```

i.e. `ernstMain` → `radix12_dif_pass1` for the 48K `{12,8,16,16}` decomposition, exercising the standalone `_pass1` routine on the production path. Because the fall-through leaves `j1` holding stack garbage, the resulting `a[j1 + …]` index is out of bounds nondeterministically — it SIGSEGVs when the garbage is large and merely produces a bad (roundoff-rejected) result when it's small, which is why an isolated single-radset run can look benign while the sequential `-s` tier reliably dies.

Applying this PR's `#else j1 = j;` fix resolves the crash: `-s t` then completes cleanly on x86-64, arm64 and riscv64 (nosimd). So this affects any scalar build that lands on a leading-radix-12 FFT — worth flagging as a correctness/crash fix rather than warning-only.

src/radix12_ditN_cy_dif1.c
```diff
@@ -1399,6 +1399,8 @@ void radix12_dif_pass1(double a[], int n)
 		j1 = (j & mask02) + br8[j&7];
 	#elif defined(USE_SSE2)
 		j1 = (j & mask01) + br4[j&3];
+	#else
+		j1 = j; // LD: double-check this additional init is what is expected (comes from radix24)
```
tdulcet: It seems we have now double-checked this, so maybe remove the comment. This also occurs below.

tdulcet (dismissed):

It is the same bug that was also fixed in #116.

tdulcet (approved):
Thanks!

PR #126, opened 2026-07-21
Remove warnings from radix DFT files (from #34)

MarkRose:

This is the **plain-warning cleanup for the radix DFT files** — 71 commits, each removing compiler warnings from one `radix*` source file (unused-variable/parameter, format, signedness, etc.), plus the paired `*_main_carry_loop.h` where applicable. Files: the `radix{5..1024,4032}_ditN_cy_dif1.c` set, and `radix{8,16,32}_{dif_dit_pass,dyadic_square,wrapper_square}.c`.

- All 71 commits are @ldesnogu's, cherry-picked from #34 with original authorship preserved; no content changes.
- Touches only `radix*` files (plus one small doc file, below) — independent of the aliasing group (#124) and of the rename/factor cluster still in #34.
- Verified: nosimd build clean, LL self-test Res64 `71E61322CCFB396C` unchanged. (SIMD configs left to CI; both local compilers are too new to build main's inline asm — a pre-existing, unrelated issue.)

One commit (radix16_dif_dit_pass.c warnings) also adds `docs/ConvertInlineAsm.txt`, @ldesnogu's reference notes on MSVC→GCC inline-asm conversion; it rode along in that commit and is carried here as-is.

Excluded / handled separately: `705e70f` (radix12 `j1` init) — a real correctness fix, in #125.


docs/ConvertInlineAsm.txt
tdulcet, line 1: I do not believe this file is needed. See https://github.com/primesearch/Mlucas/pull/34#discussion_r2278710833.

Edit: I see this came from a comment in the `src/radix16_dif_dit_pass.c` file. I would just keep it as a inline comment.

src/radix1008_ditN_cy_dif1.c
```diff
@@ -296,32 +370,60 @@ int radix1008_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
...
 
 #ifdef MULTITHREAD
 
 	static struct cy_thread_data_t *tdat = 0x0;
 	// Threadpool-based dispatch stuff:
-	static int main_work_units = 0, pool_work_units = 0;
+  #if 0//def OS_TYPE_MACOSX
+	static int main_work_units = 0;
+  #endif
+	static int pool_work_units = 0;
```
tdulcet, line 426: Why would this be used for macOS? It also occurs in many of the other files.

src/radix1008_ditN_cy_dif1.c
```diff
@@ -551,18 +655,26 @@ int radix1008_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 		ss0       = tmp + 0x05;
 		tmp += 0x6;	// sc_ptr += 0xfc6
 	  #ifdef USE_AVX512
-		cy_r = tmp;	cy_i = tmp+0x07e;	tmp += 2*0x07e;	// RADIX/8 vec_dbl slots for each of cy_r and cy_i carry sub-arrays
+		cy_r = tmp;										// RADIX/8 vec_dbl slots for each of cy_r and cy_i carry sub-arrays
+	   #ifndef MULTITHREAD
+					cy_i = tmp+0x07e;					// RADIX/8 vec_dbl slots for each of cy_r and cy_i carry sub-arrays
+	   #endif
+										tmp += 2*0x07e;	// RADIX/8 vec_dbl slots for each of cy_r and cy_i carry sub-arrays
```
tdulcet: I do not believe we would need to duplicate comments when splitting up definitions or assignments across multiple lines. It can make the code harder to read. This occurs many other times in this PR.

src/radix15_ditN_cy_dif1.c
```diff
@@ -53,15 +53,20 @@ int radix15_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
+	  #endif
+		,t1,t2
 		,aj1p0r,aj1p1r,aj1p2r,aj1p3r,aj1p4r,aj1p5r,aj1p6r,aj1p7r,aj1p8r,aj1p9r,aj1p10r,aj1p11r,aj1p12r,aj1p13r,aj1p14r
 		,aj1p0i,aj1p1i,aj1p2i,aj1p3i,aj1p4i,aj1p5i,aj1p6i,aj1p7i,aj1p8i,aj1p9i,aj1p10i,aj1p11i,aj1p12i,aj1p13i,aj1p14i
 		,cy_r0,cy_r1,cy_r2,cy_r3,cy_r4,cy_r5,cy_r6,cy_r7,cy_r8,cy_r9,cy_r10,cy_r11,cy_r12,cy_r13,cy_r14
 		,cy_i0,cy_i1,cy_i2,cy_i3,cy_i4,cy_i5,cy_i6,cy_i7,cy_i8,cy_i9,cy_i10,cy_i11,cy_i12,cy_i13,cy_i14
 		,temp,frac,scale;
 	double maxerr = 0.0;
+#if 0
 #if PFETCH
```
tdulcet: Maybe combine these two `#if` lines.

src/radix160_ditN_cy_dif1.c
```diff
@@ -189,11 +210,23 @@ int radix160_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
 	static int poff[RADIX>>2];	// Store mults of p4 offset for loop control
 #ifndef MULTITHREAD
 // Shared DIF+DIT:
+  #ifndef USE_SSE2
 	double rt,it;
+  #endif
+  #ifndef USE_SSE2
```
tdulcet: These two lines could be removed, as there is already a `#ifndef USE_SSE2` line above.

src/radix160_ditN_cy_dif1.c
```diff
@@ -189,11 +210,23 @@ int radix160_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
...
 	static int t_offsets[32];
+  #endif
 	// Need storage for 2 circular-shifts perms of a basic 5-vector, with shift count in [0,4] that means 2*9 elts:
-	static int dif_offsets[RADIX], dif_p20_cperms[18], dif_p20_lo_offset[32], dif_phi[ODD_RADIX];
-	static int dit_offsets[RADIX], dit_p20_cperms[18], dit_p20_lo_offset[32], dit_phi[ODD_RADIX];
+	static int dif_offsets[RADIX];
+	static int dif_p20_cperms[18], dif_p20_lo_offset[32];
```
tdulcet: These variables could be defined on the same line, as they were before. This also occurs below with the `dit_*` variables.

src/radix16_dif_dit_pass.c
```diff
@@ -171,17 +154,26 @@ void radix16_dif_pass	(double a[],             int n, struct complex rt0[], stru
...
+		,*ss0
+	  #endif
+	  #if !defined(MULTITHREAD) || (!(defined(USE_AVX2) && !defined(REFACTOR_4DFT_3TWIDDLE)))
+		,*c0thru15
+	  #endif
+		;
+	// This variable might have to be reenabled if commented out Chebyshev code below is ever enabled.
+	// In this case the variable must be 'static' if !MULTITHREAD.
+  #if defined(MULTITHREAD) && !(defined(USE_AVX2) && !defined(REFACTOR_4DFT_3TWIDDLE))
 	uint64 *sign_mask;
```
tdulcet, line 210: Maybe add a `#if 0` to handle the "'static' if !MULTITHREAD" case, to make it clearer what would need to change if this code is enabled.

src/radix16_dyadic_square.c
```diff
@@ -106,16 +120,23 @@ void radix16_dyadic_square(
   #endif
 	static uint32 *sm_arr = 0x0,*sm_ptr;	// Base-ptr to arrays of k1,k2-index-vectors used in SIMD roots-computation.
 	static vec_dbl *sc_arr = 0x0, *sc_ptr;
-	double *add0,*add1, *bdd0,*bdd1;	/* Addresses into array sections */
+	double *add0,*add1;	/* Addresses into array sections */
+  #ifndef USE_ARM_V8_SIMD
+	double *bdd0;		/* Addresses into array sections */
+  #endif
+  #ifndef USE_ARM_V8_SIMD
```
tdulcet: These two lines could be remove, as there is already a `#ifndef USE_ARM_V8_SIMD` line above.

src/radix16_wrapper_square.c
```diff
@@ -227,16 +231,16 @@ The scratch array (2nd input argument) is only needed for data table initializat
 		#endif
 
 		#ifdef USE_SSE2
-		//	fprintf(stderr, "%s: pfetch_dist = %d\n",func,pfetch_dist);
+		//	fprintf(stderr, "%s: pfetch_dist = %d\n",__FUNCTION__,pfetch_dist);
```
tdulcet: Please use C99 `__func__` instead of the non-standard `__FUNCTION__`. This also occurs below.

src/radix192_ditN_cy_dif1.c
```diff
@@ -185,9 +201,15 @@ int radix192_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
 	static uint64 psave = 0;
 	static uint32 bw,sw,bjmodnini,p1,p2,p3,p4,p5,p6,p7,p8,p9,pa,pb,pc,pd,pe,pf
 		,p10,p20,p30,p40,p50,p60,p70,p80,p90,pa0,pb0, nsave = 0;
+  #ifndef MULTITHREAD
 // Shared DIF+DIT:
-	int ilo,ihi,idx,jdx,lo_neg,hi_neg,k0,k1,k2,nshift, *iptr;
-	static int plo[16],phi[12], t_offsets[64];
+	int ilo,ihi,idx,jdx,hi_neg,k0,k1,k2, *iptr;
+	int nshift;
```
tdulcet: These two lines could be combined, as they were originally.

src/radix20_ditN_cy_dif1.c
```diff
@@ -217,23 +247,37 @@ int radix20_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
+		,*r10,*r20,*r30
+		,*s1p00r,*s1p01r,*s1p02r,*s1p03r,*s1p04r,*s1p05r,*s1p06r,*s1p07r,*s1p08r,*s1p09r,*s1p10r,*s1p11r,*s1p12r,*s1p13r,*s1p14r,*s1p15r,*s1p16r,*s1p17r,*s1p18r,*s1p19r
+	#endif
+		;
 
+  #ifndef MULTITHREAD
 	static int *bjmodn00,*bjmodn01,*bjmodn02,*bjmodn03,*bjmodn04,*bjmodn05,*bjmodn06,*bjmodn07,*bjmodn08,*bjmodn09,*bjmodn10,*bjmodn11,*bjmodn12,*bjmodn13,*bjmodn14,*bjmodn15,*bjmodn16,*bjmodn17,*bjmodn18,*bjmodn19;
 	static vec_dbl *cy00,*cy04,*cy08,*cy12,*cy16;
-  #ifndef USE_AVX
+   #if !defined(USE_AVX)
```
tdulcet: We should keep the `#ifndef USE_AVX` here, just indent it.

src/radix224_ditN_cy_dif1.c
```diff
@@ -259,32 +277,50 @@ int radix224_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
 		,p10,p20,p30,p40,p50,p60,p70,p80,p90,pa0,pb0,pc0,pd0, nsave = 0;
 	static int poff[RADIX>>2];	// Store mults of p04 offset for loop control
 #ifndef MULTITHREAD
+  #ifndef USE_SSE2
 	double re,im;
-	static int t_offsets[32], dif_offsets[RADIX], dit_offsets[RADIX];
+  #endif
+  #ifndef USE_SSE2
```
tdulcet: These lines could be removed, as there is already a `#ifndef USE_SSE2` line above.

src/radix24_ditN_cy_dif1.c
```diff
@@ -179,44 +179,77 @@ int radix24_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
   #elif defined(USE_AVX)
+   #ifndef MULTITHREAD
 	static struct uint32x4 *n_minus_sil,*n_minus_silp1,*sinwt,*sinwtm1;
+   #endif
   #else
+   #ifndef MULTITHREAD
 	int n_minus_sil,n_minus_silp1,sinwt,sinwtm1;
 	double wtl,wtlp1,wtn,wtnm1;	/* Mersenne-mod weights stuff */
+   #endif
   #endif
```
tdulcet, line 253: Maybe just gate this entire section in `#ifndef MULTITHREAD`, to avoid repeating that in every inner block. This also occurs below.

src/radix288_ditN_cy_dif1.c
```diff
@@ -184,8 +200,12 @@ int radix288_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
 	static int poff[RADIX>>2];	// Store mults of p4 offset for loop control
 #ifndef MULTITHREAD
 // Shared DIF+DIT:
+  #ifndef USE_SSE2
 	double rt,it,re;
+  #endif
+  #ifndef USE_SSE2
```
tdulcet: These lines could be removed, as there is already a `#ifndef USE_SSE2` line above.

src/radix320_ditN_cy_dif1.c
```diff
@@ -2122,7 +2195,10 @@ void radix320_dit_pass1(double a[], int n)
 	// In order to preserve length-2 numeric-index-offset property here, use hex for digit:
 	static int NDIVR,p1,p2,p3,p4,p5,p6,p7,p8,p9,pa,pb,pc,pd,pe,pf
 		,p10,p20,p30,p40,p50,p60,p70,p80,p90,pa0,pb0,pc0,pd0,pe0,pf0,pg0,ph0,pi0,pj0, first_entry=TRUE;
-	static int plo[16],phi[20];
+	static int plo[16];
+  #ifndef USE_SSE2
+	//static int phi[20];
+  #endif
```
tdulcet: Maybe comment out this `#ifndef` as well, since it is an unused variable.

src/radix32_ditN_cy_dif1.c
```diff
@@ -452,14 +514,14 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
-		r0C = tmp + 0x0c;	r1C = tmp + 0x1c;	r2C = tmp + 0x2c;	r3C = tmp + 0x3c;
-		r0E = tmp + 0x0e;	r1E = tmp + 0x1e;	r2E = tmp + 0x2e;	r3E = tmp + 0x3e;
+		r00 = tmp + 0x00;	//r10 = tmp + 0x10;	r20 = tmp + 0x20;	r30 = tmp + 0x30;
+		//rr02 = tmp + 0x02;	r12 = tmp + 0x12;	r22 = tmp + 0x22;	r32 = tmp + 0x32;
+		//rr04 = tmp + 0x04;	r14 = tmp + 0x14;	r24 = tmp + 0x24;	r34 = tmp + 0x34;
+		//rr06 = tmp + 0x06;	r16 = tmp + 0x16;	r26 = tmp + 0x26;	r36 = tmp + 0x36;
+		//rr08 = tmp + 0x08;	r18 = tmp + 0x18;	r28 = tmp + 0x28;	r38 = tmp + 0x38;
+		//rr0A = tmp + 0x0a;	r1A = tmp + 0x1a;	r2A = tmp + 0x2a;	r3A = tmp + 0x3a;
+		//rr0C = tmp + 0x0c;	r1C = tmp + 0x1c;	r2C = tmp + 0x2c;	r3C = tmp + 0x3c;
+		//rr0E = tmp + 0x0e;	r1E = tmp + 0x1e;	r2E = tmp + 0x2e;	r3E = tmp + 0x3e;
```
tdulcet: It looks like an extra `r` was accidentally added to the beginning of these lines, after the comment.

src/radix32_main_carry_loop.h
```diff
@@ -59,14 +59,15 @@ for(k=1; k <= khi; k++)	/* Do n/(radix(1)*nwt) outer loop executions...	*/
 
 	if(MODULUS_TYPE == MODULUS_TYPE_GENFFTMUL)
 	{
-		l = 0; addr = (double *)cy_r; addi = (double *)cy_i;
+		// TODO: 'addi' use is commented out since 'genfftmul_carry_norm_pow2_errcheck' doesn't use its cy parameter in its current implementation
```
tdulcet: Should we open an issue to track this?

src/radix36_ditN_cy_dif1.c
```diff
@@ -186,43 +205,56 @@ int radix36_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 					s   =  0.64278760968653932631,	/* sin(2*pi/9) */
 					c2  =  0.17364817766693034887,	/* cos(2*u) */
 					s2  =  0.98480775301220805936,	/* sin(2*u) */
-					c3  = -0.50000000000000000000,	/* cos(3*u) */
```
tdulcet, line 189: Maybe we should comment out this variable instead, for consistency with the other similar changes where we kept constants that could potenchally be useful in the future? This also occurs below.

src/radix40_ditN_cy_dif1.c
```diff
@@ -140,24 +140,44 @@ int radix40_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
 	const char func[] = "radix40_ditN_cy_dif1";
+#if !defined(MULTITHREAD) && defined(USE_SSE2)
 	const int pfetch_dist = PFETCH_DIST;
+#endif
 	const int stride = (int)RE_IM_STRIDE << 1;	// main-array loop stride = 2*RE_IM_STRIDE
 #ifdef USE_SSE2
-  #if COMPACT_OBJ
+  #if COMPACT_OBJ && !defined(MULTITHREAD)
 	static uint32 pp07[8];
+   #ifndef MULTITHREAD
```
tdulcet: This is redundant with the above `!defined(MULTITHREAD)`.

src/radix48_ditN_cy_dif1.c
```diff
@@ -143,38 +143,59 @@ int radix48_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
 	const char func[] = "radix48_ditN_cy_dif1";
+#if !defined(MULTITHREAD) && defined(USE_SSE2)
 	const int pfetch_dist = PFETCH_DIST;
+#endif
 	const int stride = (int)RE_IM_STRIDE << 1;	// main-array loop stride = 2*RE_IM_STRIDE
 #ifdef USE_SSE2
-  #if COMPACT_OBJ
+  #if COMPACT_OBJ && !defined(MULTITHREAD)
 	static uint32 plo[16];
+   #ifndef MULTITHREAD
```
tdulcet: This is redundant with the above `!defined(MULTITHREAD)`.

tdulcet (changes requested):
Wow, this is a amazing! Thank you @ldesnogu for all your work on this.

I have reviewed through the `src/radix52_main_carry_loop.h` file and I will try to review the remainder tomorrow. So far, I just have a couple questions and a few opportunities for minor simplification. I typically just flagged the first occurrence of each potential issue that I found.

This PR makes several changes to the `USE_IMCI512` define, which currently has no test coverage, so we may want to finish and merge #155 first, so we could at least check that this  still builds for K1OM.

ldesnogu:
Thanks a lot @MarkRose and @tdulcet, I really appreciate that this didn't get lost :)

src/radix56_ditN_cy_dif1.c
```diff
@@ -172,37 +172,61 @@ int radix56_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
 	const char func[] = "radix56_ditN_cy_dif1";
+#if !defined(MULTITHREAD) && defined(USE_SSE2)
 	const int pfetch_dist = PFETCH_DIST;
+#endif
 	const int stride = (int)RE_IM_STRIDE << 1;	// main-array loop stride = 2*RE_IM_STRIDE
 #ifdef USE_SSE2
-  #if COMPACT_OBJ
+  #if COMPACT_OBJ && !defined(MULTITHREAD)
 	static uint32 pp07[8];
+   #ifndef MULTITHREAD
```
tdulcet: This line is redundant, as there is already `!defined(MULTITHREAD)` above.

src/radix56_ditN_cy_dif1.c
```diff
@@ -262,48 +291,87 @@ int radix56_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
+  #ifndef MULTITHREAD
+   #ifndef USE_SSE2
+	struct complex *tptr;
+   #endif
+	double *addr;
+   #ifndef USE_SSE2
+	double *addi;
+   #endif
+  #endif
+  #ifndef MULTITHREAD
```
tdulcet: These two lines could be removed, as there is already a `#ifndef MULTITHREAD` line above.

src/radix60_ditN_cy_dif1.c
```diff
@@ -2480,17 +2607,27 @@ void radix60_dit_pass1(double a[], int n)
 	{
 		struct cy_thread_data_t* thread_arg = targ;	// Move to top because scalar-mode carry pointers taken directly from it
 		double *addr,*addi;
+	#ifndef USE_SSE2
 		struct complex *tptr;
+	#endif
+	#ifdef USE_SSE2
```
tdulcet: These two lines could be replaced with an `#else`. This occurs several other times.

src/radix60_ditN_cy_dif1.c
```diff
@@ -2541,39 +2684,78 @@ void radix60_dit_pass1(double a[], int n)
...
-		vec_dbl *max_err, *sse2_rnd, *half_arr, *two, *sse2_c3m1, *sse2_s, *sse2_cn1, *sse2_cn2, *sse2_ss3, *sse2_sn1, *sse2_sn2,
-			*s1p00,*s1p01,*s1p02,*s1p03,*s1p04,*s1p05,*s1p06,*s1p07,*s1p08,*s1p09,*s1p0a,*s1p0b,*s1p0c,*s1p0d,*s1p0e,
+	  #endif
+		vec_dbl *max_err,
+		  #ifndef USE_AVX512
+			*sse2_rnd,
+		  #endif
+			*half_arr,
+		  #ifndef USE_ARM_V8_SIMD // TODO that looks strange that this is needed to get rid of unused 'two'
+			*two,
```
tdulcet: We should look into this TODO comment. It also occurs below.

src/radix64_ditN_cy_dif1.c
```diff
@@ -251,43 +301,73 @@ int radix64_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
...
-		*r60,*r62,*r64,*r66,*r68,*r6A,*r6C,*r6E,*r70,*r72,*r74,*r76,*r78,*r7A,*r7C,*r7E,
+		*r00,
+	  #ifndef MULTITHREAD
+		*r02,*r04,*r06,*r08,*r0A,*r0C,*r0E,*r10,*r20,*r30,*r40,*r50,*r60,*r70,
+		//*r12,*r14,*r16,*r18,*r1A,*r1C,*r1E,
+		//*r22,*r24,*r26,*r28,*r2A,*r2C,*r2E,*r32,*r34,*r36,*r38,*r3A,*r3C,*r3E,
+		//*r42,*r44,*r46,*r48,*r4A,*r4C,*r4E,*r52,*r54,*r56,*r58,*r5A,*r5C,*r5E,
+		//*r62,*r64,*r66,*r68,*r6A,*r6C,*r6E,*r72,*r74,*r76,*r78,*r7A,*r7C,*r7E,
+	  #endif
+	  #ifndef MULTITHREAD
```
tdulcet: These two lines could be removed, as there is already a `#ifndef MULTITHREAD` line above.

src/radix960_ditN_cy_dif1.c
```diff
@@ -173,29 +179,54 @@ int radix960_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
 	const int jhi_wrap_mers =  7;
 	const int jhi_wrap_ferm = 15;	// For right-angle transform need *complex* elements for wraparound, so jhi needs to be twice as large
   #endif
-	int NDIVR,i,incr = 0,j,j1,j2,jt,jp,jstart,jhi,full_pass,k,khi,l,ntmp,outer,nbytes;
+	int NDIVR,i,j,j1,jt,jstart,jhi,full_pass,khi,l,outer;
+  #if !defined(MULTITHREAD) && defined(USE_SSE2)
+	int incr = 0;
+  #endif
```
tdulcet: This definition should be moved down into the below block, as you did in the other files. This also occurs below.

src/radix960_ditN_cy_dif1.c
```diff
@@ -315,31 +375,53 @@ int radix960_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
...
   #ifdef USE_AVX512
-	static int mcycle[ODD_RADIX],mc_idx;
-	static int ncycle[ODD_RADIX],nc_idx;
-	static int ocycle[ODD_RADIX],oc_idx;
-	static int pcycle[ODD_RADIX],pc_idx;
+	static int mcycle[ODD_RADIX]/* ,mc_idx */;
+	static int ncycle[ODD_RADIX]/* ,nc_idx */;
+	static int ocycle[ODD_RADIX]/* ,oc_idx */;
+	static int pcycle[ODD_RADIX]/* ,pc_idx */;
+  #ifndef MULTITHREAD
```
tdulcet, line 418: These commented out variables should now be removed, as their definitions were moved below.

src/radix960_ditN_cy_dif1.c
```diff
@@ -286,26 +323,49 @@ int radix960_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
...
 /*...stuff for the reduced-length DWT weights array is here:	*/
 	static int n_div_nwt;
-	int col,co2,co3,m,m2;
+  #ifndef MULTITHREAD
+	int col,co2,co3;
+	#ifndef USE_SSE2
+	int m,m2;
+	#endif
+  #endif
+  #ifndef USE_PTHREAD
```
tdulcet: Was `USE_PTHREAD` intended here, as usage of `USE_PTHREAD` was replaced in the other files with `USE_SSE2`? It also occurs below in this file.

src/radix992_ditN_cy_dif1.c
```diff
@@ -160,28 +160,37 @@ int radix992_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[]
...
 	static double ndivr_inv;
 	uint64 itmp64;
 	static uint64 psave = 0;
 	static uint32 bw,sw,bjmodnini,p1,p2,p3, nsave = 0;
 	static int poff[RADIX>>2];
 #ifndef MULTITHREAD
 	// Need storage for circular-shifts perms of a basic 31-vector, with shift count in [0,31] that means 2*31 elts:
-	static int dif_p20_cperms[62], plo[32],phi[62],jj[32], *iptr;
-	int kk,idx,pidx,mask,lshift, is_even,is_odd, k0,k1,k2,k3,k4,k5,k6,k7,k8,k9,ka,kb,kc,kd,ke,kf, o[32];	// o[] stores o-address offsets for current radix-32 DFT in the 31x-loop
+	static int dif_p20_cperms[62], plo[32],phi[62],jj[32], *iptr, o[32];	// o[] stores o-address offsets for current radix-32 DFT in the 31x-loop
```
tdulcet: Was converting the `o` array to static intended?

tdulcet (commented):
Thanks again for all your work on this!

I finally finished my review. I just have a couple of additional questions and a few more opportunities for minor simplification. Overall, it looks great.

This is such a massive change that it would be helpful if @MarkRose could use his AI to review it as well. It took me three days to review, so I may have missed some things.

This would not need to be done in this PR, but note that there are some additional opportunities for simplification of these added preprocessor defines that could improve readability. For example, there are many parallel blocks that all include `!defined(MULTITHREAD)` (or `#ifndef MULTITHREAD`), so that could potenchally be factored out to avoid repeating it in every block.

ldesnogu:
Thanks @tdulcet
As you wrote, your comments could be taken care of once this is pushed. IMHO the massive amount of work that has already been done shouldn't be extended too far and only fixes to regressions should be made, so that further cleanups can be done on a solid basis.

PS - by regressions, I mean regressions introduced by all the current patch sets.

tdulcet:
> As you wrote, your comments could be taken care of once this is pushed. IMHO the massive amount of work that has already been done shouldn't be extended too far and only fixes to regressions should be made, so that further cleanups can be done on a solid basis.

Yes, my comments were only about that code that was changed in this PR. Many of them are just nits that could be addressed in follow ups, but I did have a couple of questions that we may want to resolve before merging this, to prevent potential regressions.

tdulcet (commented):
Thanks for the fixes. Before approving this, it might be good to look into https://github.com/primesearch/Mlucas/pull/126#discussion_r3650216206 above (a TODO comment that @ldesnogu added), which could potentially be masking a bug. Otherwise, this looks good to me.

MarkRose:
Thanks for the thorough review — three days on this is a lot. Everything below is one consolidated reply covering all of it; I'd posted it in pieces earlier and have deleted those to keep the thread readable.

Ten commits on top of @ldesnogu's 71, all mine. His are untouched.

## Two defects found while working through the review

**`o[]` silently became `static` in radix992** ([r3650275227](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650275227)) — `fb4297f`. While the declarations were being split up, `o[32]` was merged into the adjacent `static int` line. On `main` it is automatic, and the three other `o[32]` declarations in that same file (`radix992_dif_pass1`, `radix992_dit_pass1`, `cy992_process_chunk`) all remain automatic. Not a live bug — the block is inside `#ifndef MULTITHREAD` so it only ever compiles single-threaded, and all 32 entries are written before use in each radix-32 DFT batch — but it is a semantic change in a PR that makes none, and would become a data race if that block ever gained a threaded path. `.bss` shrinks 160 bytes and `.text` grows 192, confirming it moved back off static storage.

**Two `#error Unhandled combination of preprocessor flags!` guards were deleted** (not raised in review) — `75d7206`. `main` has three, in radix28/radix56/radix224; only radix224's survived here. Enumerating the `#if` chain around the radix-7 trig constants over MULTITHREAD × USE_SSE2 × USE_AVX2 × USE_ARM_V8_SIMD × LO_ADD gives exactly one combination where this branch takes neither arm but `main` took one: **MULTITHREAD + scalar-double (no SSE2) with the default LO_ADD=1**, a threaded non-SIMD build. So the `#error` began firing and was deleted rather than accommodated. Harmless in itself — threaded builds keep those constants in `cy_thread_data_t`, and both files compile clean in that configuration on `main` and here — but it means a genuinely unhandled combination would now silently declare no constants instead of failing at compile time. Both restored with an explicit `#elif defined(MULTITHREAD)` no-op branch documenting the intended gap; verified the `#error` fires in none of 12 configurations. Guard count is back to `main`'s 3 in 3 files.

## The `two` TODO in radix60 — not masking a bug

[r3650216206](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650216206) — `fdae047`. Both TODOs replaced with the explanation:

- `two` is passed to `SSE2_RADIX_15_{DIF,DIT}_X2` from inside a plain `#ifdef USE_SSE2` block, so it is referenced on every SIMD arch;
- those macros forward it to `SSE2_RADIX_05_DFT_0TWIDDLE_X2`;
- on ARM V8 that `_X2` macro is a two-line wrapper calling the non-`_X2` `SSE2_RADIX_05_DFT_0TWIDDLE` twice — and the non-`_X2` macro takes no `two` argument on *any* arch. The argument is discarded, so the variable really is unused there.

Verified by expanding the macro with a marker token in place of `two`: it appears **0** times under `USE_ARM_V8_SIMD` and **exactly once** under each of `USE_SSE2`, `USE_AVX`, `USE_AVX2`, `USE_AVX512`. The guard is layout-safe — the assignment is an absolute offset (`two = tmp + 0x07`) with an unconditional `tmp += 0x08` after it — and `VEC_DBL_INIT(two, 2.0)` is unguarded, so the slot still holds 2.0. The same guard appears without a TODO in radix40, radix160 and radix960, for the same reason.

## The `addi` TODO in radix32 — a dangling pointer

[r3646940419](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646940419) — `572c613` and `2bf9930`. The TODO was accurate: the active V2 `genfftmul_carry_norm_pow2_errcheck()` does ignore its `cy` parameter. But commenting out `addi = (double *)cy_i;` while still passing `*addi` left `addi` declared and never assigned. That is inert today, since an unused macro parameter is never expanded and so never evaluated — but `carry.h` selects between two implementations with `#if 0`/`#else`, and the V1 variant **does** use `cy`. Flipping that switch dereferences an uninitialised pointer, and neither gcc nor clang diagnoses it at `-O2 -Wall -Wextra`.

- `572c613` (`src/carry.h`) documents the asymmetry and marks `cy` used inside V2, so a caller can keep a correct `cy_i` setup without the warning. `(void)sizeof(cy)` rather than `(void)(cy)`, because `sizeof`'s operand is unevaluated: it marks the parameter used without dereferencing it, which matters since callers pass an lvalue like `*addi`. `.text` is byte-identical for both callers, so the shipped V2 build is unchanged.
- `2bf9930` restores the initialisation and the four `++addi` increments and drops the TODO. That block is byte-identical to `main` again.

For the record, `main` does *not* emit an `addi` warning in the default configuration, because `addi` is shared with the Fermat-mod carry block below, which reads it. So this pair is robustness and documentation, not a bug fix.

## `#if 0//def OS_TYPE_MACOSX` — no change needed

[r3644305903](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644305903). It is not used for macOS; that path is already disabled, and this PR makes it consistent. The idiom is pre-existing — 68 occurrences across 34 files on `main` — and means "a formerly-macOS-only path, switched off, with the original condition left visible". What warned is that on `main` the *declaration* was unconditional (`radix1008_ditN_cy_dif1.c:324`) while all its *uses* already sat inside those disabled blocks. This PR moves the declaration under the same guard. All four sites in radix1008 (declaration 424, assignments 567/568/572, loop 1941) are inside disabled blocks, so nothing references an undeclared variable.

## Guard and style tidy-up

`367c07e` (scripted) and `4b20d0f` (hand edits). The scripted pass applies only to directive lines this PR introduced, so pre-existing structure is untouched:

| transformation | sites |
|---|---|
| merge adjacent blocks with an identical condition | 10 |
| drop an inner guard already implied by an enclosing `… && !defined(X)` | 3 |
| collapse an inverse guard pair into `#else` | 26 |
| spell simple conditions `#ifndef X` / `#ifdef X` rather than `#if !defined(X)` | 10 |
| drop a trailing comment repeating the identical comment on the line above | 205 lines |

Covers [r3644586341](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644586341), [r3644841399](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644841399), [r3646464243](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646464243), [r3646739075](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646739075), [r3647086949](https://github.com/primesearch/Mlucas/pull/126#discussion_r3647086949), [r3647165082](https://github.com/primesearch/Mlucas/pull/126#discussion_r3647165082), [r3650181686](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650181686), [r3650187358](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650187358), [r3650212580](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650212580), [r3650228586](https://github.com/primesearch/Mlucas/pull/126#discussion_r3650228586), [r3646421812](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646421812), [r3644322675](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644322675).

On duplicated comments, radix64 is the clearest case: `main` had `static int poff[RADIX>>2],po_lin[8],po_br[8];	// Store mults of p-offsets for loop-controlled DFT macro calls` on one line. Split into three guarded lines, that comment went from 2 occurrences in the file to 6; it is back to 2.

Hand edits in `4b20d0f`:

- **radix15** ([r3644569054](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644569054)) — nested `#if 0` / `#if PFETCH` combined into `#if 0 && PFETCH`.
- **radix160** ([r3644600663](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644600663)) — `dif_*`/`dit_*` rejoined onto single lines. `dif_phi[]`/`dit_phi[]` cannot join them: those two are under `#ifdef USE_SSE2` while the rest are unguarded.
- **radix192** ([r3646210645](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646210645)) — recombined onto the one line `main` had, minus `lo_neg`, which is genuinely unused.
- **radix320** ([r3646853251](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646853251)) — dropped the `#ifndef USE_SSE2` around the already commented-out `static int phi[20];`.
- **radix36** ([r3646988673](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646988673)) — `c3 = -0.5 /* cos(3*u) */` kept as a commented-out constant at the three sites where it had been deleted.
- **radix16_dif_dit_pass** ([r3644754600](https://github.com/primesearch/Mlucas/pull/126#discussion_r3644754600)) — added an inert `#elif 0` branch showing the `static` form a !MULTITHREAD build would need if the Chebyshev code is re-enabled.

## How the review round was verified

For every touched file the preprocessed token stream is compared against the previous commit across 10 configurations — nosimd, SSE2, ARM_V8_SIMD, AVX, AVX2, AVX512 and the threaded variant of each — with numeric and string literals normalised, since `__LINE__` and `__FILE__` legitimately move. 34 of 36 files are token-identical. The two that are not are radix160 and radix192, where rejoining declarations necessarily removes lines; those were checked by object code instead, and every differing instruction is an `ASSERT` `__LINE__` immediate shifted by exactly the number of lines removed. Everything compiles clean under nosimd and `-DUSE_THREADS` with zero net new `-Wall -Wextra` warnings.

That check earned its keep: it rejected two earlier versions of the scripted pass — one merged pre-existing `#ifdef USE_PTHREAD` section boundaries instead of the guards under review, the other collapsed a run of alternating `#ifdef`/`#ifndef` blocks into two `#else` directives in a single conditional.

## Still open: the radix24 section gating

[r3646599780](https://github.com/primesearch/Mlucas/pull/126#discussion_r3646599780) — deferred rather than done, and I want to be explicit about why, since it is the one review item without a commit.

Gating the whole block in a single `#ifndef MULTITHREAD` cannot be a straight wrap. That declaration region has 14 MULTITHREAD-related guards, but they are interleaved with declarations that are not MULTITHREAD-gated:

- `jhi_wrap` (`#ifdef USE_AVX512` / `#else`) and `nbytes` (`#ifdef USE_SSE2`) carry no MT guard at all;
- `dtmp` is `#if defined(MULTITHREAD) && defined(USE_SSE2) && !defined(USE_AVX512)` — the opposite polarity;
- one block is `#if !defined(MULTITHREAD) || defined(USE_SSE2)`, a disjunction that is true in some MULTITHREAD builds;
- the trailing arch switch (`#ifdef USE_AVX512` / `#elif defined(USE_AVX)` / `#else`) has its own `#ifndef MULTITHREAD` inside each of the three arms.

So factoring it means reordering declarations so the MT-only ones become contiguous, plus special-casing the two opposite-polarity blocks.

The reason I would rather not do that here is verification, not taste. Every other change in this round was shown byte-identical at the preprocessed-token level across all 10 configurations. Reordering declarations changes the token stream, so that check stops applying, and the fallback — object-code comparison — only runs locally for nosimd and threaded-nosimd, because this machine cannot build the inline asm for the six SIMD configurations. Those are exactly the configurations where a mis-scoped declaration would bite, since most of the interleaving here is SIMD-conditional. I would rather not land a restructure whose riskiest configurations I could not check.

If you do want it, the clean way is a follow-up PR where it is the only change: CI's SIMD matrix becomes the verification, and it can be reviewed in isolation rather than as commit 82 of this one. Happy to open that whenever you like. And if you would prefer it verified to the same standard as the rest of this round, it can be: comparing the *set* of declarations reaching each configuration as an order-insensitive multiset, rather than comparing preprocessed output for equality, tests exactly the property that matters and works for all 10 configurations without needing to assemble the asm.

tdulcet (approved):
Thank you for addressing all of my feedback. This looks great!

> Gating the whole block in a single `#ifndef MULTITHREAD` cannot be a straight wrap. That declaration region has 14 MULTITHREAD-related guards, but they are interleaved with declarations that are not MULTITHREAD-gated:

Lines 236-253 could replaced with a single outer `#ifndef MULTITHREAD` instead of repeating that in each inner block, but yes, this could be addressed in a follow up:
```c
  #ifndef MULTITHREAD
	int col,co2,co3;
   #ifdef USE_AVX512
	double t0,t1,t2,t3;
	static struct uint32x8 *n_minus_sil,*n_minus_silp1,*sinwt,*sinwtm1;
   #elif defined(USE_AVX)
	static struct uint32x4 *n_minus_sil,*n_minus_silp1,*sinwt,*sinwtm1;
   #else
	int n_minus_sil,n_minus_silp1,sinwt,sinwtm1;
	double wtl,wtlp1,wtn,wtnm1;	/* Mersenne-mod weights stuff */
   #endif
  #endif
```

PR #129, opened 2026-07-21
Remove warnings from qfloat / pm1 / factor_test (from #34)

MarkRose:

Removes compiler warnings from `qfloat.c`, `pm1.c` (two commits), and `factor_test.h` (5 commits total).

- All @ldesnogu's, cherry-picked from #34 with original authorship preserved.

**Reviewer note:** one commit here — `ac28c59` "Define qfldbl and ldbl_to_q only for x87 code" — is a *functional* change (it `#if`-gates those `qfloat` helpers to x87 builds), not a pure warning removal. Worth a look to confirm the x87 gating is correct. The other four are mechanical.

### DIFF
--- a/src/pm1.c
+++ b/src/pm1.c
@@ -1025,7 +1025,9 @@ based on iteration count versus PM1_S1_PROD_BITS as computed from the B1 bound, 
 	// "working map". Alas, since we alloc map[] at runtime, we can't actually declare hi|lo as const-ptrs. File under
 	// "stupid C tricks" - lack of a "set once at runtime but compiler/OS treat as const subsequently" declaration option:
 	static uint8 *map = 0x0,*lo = 0x0,*hi = 0x0; uint8 *map_lo,*map_hi, *rmap;
+#if !defined(PM1_STANDALONE) && (defined(PM1_DEBUG) || USE_PP1_MULTS)
 	static uint64 *vec1 = 0ull, *vec2 = 0ull;
+#endif
 	static uint32 nlimb, *b = 0x0;
 	static double *a_ptmp = 0x0, *a = 0x0;	// Storage for stage 2 in form of ptrs to m*[24,40 or 48] residue-length subarrays, depending
 	static double **buf = 0x0;	// on whether bigstep = [210,330,420 or 840]. a[] is simply a tmp-ptr used to init buf[]
@@ -1220,7 +1222,7 @@ based on iteration count versus PM1_S1_PROD_BITS as computed from the B1 bound, 
 	// 2 extra word-slots at high end of map used for these temps - can't declare as const pointers,0x but treat as such below:
 	lo = map + m*wsize; hi = lo + wsize;
 	rmap = hi + wsize;
-#ifndef PM1_STANDALONE
+#if !defined(PM1_STANDALONE) && (defined(PM1_DEBUG) || USE_PP1_MULTS)
 	// We know that the arrtmp vector passed in by caller is large enough to hold
 	// at least 2 nlimb-sized arrays,0x so simply point vec1 and vec2 at successive nlimb-sections of it:
 	vec1 = arrtmp; vec2 = arrtmp + nlimb;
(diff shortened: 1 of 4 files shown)


src/pm1.c
```diff
@@ -425,27 +425,40 @@ uint32 pm1_s1_ppow_prod(const uint64 iseed, const uint32 b1, uint64 accum[], uin
 	fprintf(stderr,"Pre-loop accumulator = %" PRIu64 " + 2^64*%" PRIu64,accum[0],accum[1]);
   }
 #endif
+// Undef to get debug output.
+//#define DEBUG_PM1_S1_PPOW_PROD
```
tdulcet: Do you mean uncomment the define?

Maybe we should just comment out the unused code, rather than adding the complexity of an additional define here.

src/pm1.c
```diff
@@ -2122,6 +2139,7 @@ MME = 0;
 				++nmodmul;
 			#endif
 			}
+			(void)nq; // currently unused
```
tdulcet: Should we comment out the definition and initialization instead?

tdulcet (commented):
Thanks, I just have a couple of questions.

tdulcet (changes requested):
(Just requesting changes per https://github.com/primesearch/Mlucas/pull/129#pullrequestreview-4765586926 above.)

MarkRose:
Thanks — both good catches, and chasing the first one turned up a build regression in the branch as well. Pushed three commits on top; @ldesnogu's five are untouched.

**1. `DEBUG_PM1_S1_PPOW_PROD` (pm1.c:429)** — you're right, the comment was backwards: you *uncomment* (define) it to get output, not undef it. And rather than add a second debug switch, I've reused `PM1_DEBUG`, which pm1.c already uses in ~35 places and which this file's own header comment documents at pm1.c:25. That drops the misleading comment, the commented-out `#define` and the trailing `#undef`, while keeping `j` declared only where it's used so the warning stays fixed. Also `uint32_t j` → `uint32` for consistency, and gated the "Stage 1 exponent" print that pairs with the trailing newline so the output is coherent.

**2. `(void)nq` (pm1.c:2142)** — agreed, removed rather than voided. Worth recording *why* that's safe: `nq` is documented as "# of stage 2 q's processed", but it is only incremented in the M-odd 0-interval loop and never in the extended-window loop that handles the bulk of the q's, so it could not hold the total it claims even if it were printed. Abandoned scaffolding rather than a diagnostic someone forgot to report. `np`/`ns` *are* maintained in both loops and are still reported in the summary line. I deleted it with a comment recording the above rather than commenting the declaration out — happy to switch if you'd prefer to keep the shape.

**3. A regression this branch introduced: `-DPM1_DEBUG` stopped compiling.** `9998395` moved `uint32 word,bit;` behind `#if USE_PP1_MULTS`, but `bit` is also used by the prime-pairing cross-checks under `#ifdef PM1_DEBUG` (pm1.c:2093/2095), so a PM1_DEBUG build without `USE_PP1_MULTS` failed:

```
pm1.c:2093:33: error: 'bit' undeclared (first use in this function)
```

That config compiles cleanly on `main`, and it is the one documented at pm1.c:25. Fixed by declaring just `bit` in the PM1_DEBUG-without-PP1 case — declaring `word` as well would re-introduce an unused-variable warning in exactly the build this PR is cleaning up.

**Verification**

* `pm1.c` compiles with 0 errors in all six combinations of {`PM1_STANDALONE`, full} × {–, `PM1_DEBUG`, `USE_PP1_MULTS`, both}. Under `-Wall -Wextra` the default build goes 32 → 23 warnings with **no new** ones (`j`, `nq`, `bit`, `word`, `vec1`, `vec2`, `thr_id`, `jhi`, `b1` all cleared); `-DPM1_DEBUG` goes 29 → 23.
* The default build is behaviourally unchanged: `.text` is byte-identical to the pre-fix branch, differing only in `ASSERT`'s `__LINE__` immediates.
* The `PM1_STANDALONE` simulator gives byte-identical output to the merge-base for `-bigstep 420 -b1 100000 -b2 3000000 -m 8` (`M = 8: #buf = 384, #pairs: 85077, #single: 37586 (81.91% paired), #blocks: 6502, #modmul: 135667`).

tdulcet (approved):
Thanks for the fixes! This looks good to me.

PR #130, opened 2026-07-21
Fix ALLOC_UINT/ALIGN_UINT type and prefetch macro warnings (from #34)

MarkRose:

Two small macro fixes:
- **`align.h`** (`7136a86`) — `ALLOC_UINT`/`ALIGN_UINT` used the bare `uint` type, which isn't guaranteed to be defined; changed to `uint32`. A real portability fix, not just a warning.
- **`prefetch.h`** (`2ca2100`) — give `prefetch_p_doubles` an empty-macro definition that references its parameter, silencing unused-parameter warnings at call sites.

- Both @ldesnogu's, cherry-picked from #34 with original authorship preserved.

### DIFF
--- a/src/prefetch.h
+++ b/src/prefetch.h
@@ -82,7 +82,7 @@ to point to a simple pointer variable:
 		__asm	prefetcht0 [edx]\
 	}
   #else
-	# define prefetch_p_doubles(_pd)	/* */
+	# define prefetch_p_doubles(_pd)	/* NO-OP! */ do { (void)(_pd); } while (0)
   #endif
 
 /* AMD64/Solaris, Sun C compiler */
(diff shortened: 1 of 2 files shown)


src/prefetch.h
```diff
@@ -82,7 +82,7 @@ to point to a simple pointer variable:
 		__asm	prefetcht0 [edx]\
 	}
   #else
-	# define prefetch_p_doubles(_pd)	/* */
+	# define prefetch_p_doubles(_pd)	/* NO-OP! */ do { (void)(_pd); } while (0)
```
tdulcet: Is the do/while loop necessary here? I know it is a C idiom, but could we simplify this to just `(void)(_pd)` in this case to silence the warning. This also occurs below.

tdulcet (dismissed):
Thanks. I have a minor question, but it does not block approval.

tdulcet (approved):
Thanks, this looks good to me!

PR #132, opened 2026-07-21
Rename cstr/in_line globals to g_cstr/g_in_line (from #34)

MarkRose:
Part of splitting @ldesnogu's warnings-cleanup PR #34. **Stacked on #131** — please review/merge that first.

⚠️ Until #131 merges, this PR's diff also shows #131's 11 commits. **The 3 commits that are new here** are the renames:
- `ba701b7` Rename global `cstr` → `g_cstr`
- `5c6d69f` Rename parameter `cstr` → `p_cstr`
- `2da2cca` Rename `in_line` → `g_in_line` (global) / `p_in_line` (param)

These resolve name collisions between the global buffers and same-named local parameters (a warning source). Purely mechanical 1:1 identifier substitutions across `Mlucas.[ch]`, `factor.c`, `Mdata.h`, `util.[ch]`, `get_preferred_fft_radix.c`, `mi64.c`, `twopmodq128/192/256.c` — 290 insertions / 290 deletions, no logic change.

- All @ldesnogu's, cherry-picked from #34 with original authorship preserved.
- Verified: nosimd build clean, LL self-test Res64 `71E61322CCFB396C` unchanged.

Note: these commits touch `generate_JSON_report`, where the `cstr` handling still uses `strlen`; per your earlier comment on #34 that would ideally pass the buffer size instead — a small follow-up to do on top of this rather than part of the rename.

PR #131, opened 2026-07-21
Remove warnings from Mlucas.c / mi64.c / util.c / twopmodq128-256 (from #34)

MarkRose:
Part of splitting @ldesnogu's warnings-cleanup PR #34 into smaller, independently-reviewable pieces.

Mechanical compiler-warning cleanup for the core non-radix sources — 11 commits touching `Mlucas.c`, `mi64.c`, `util.c` (incl. removing the unused `VAR_WARN` macro), and `twopmodq128/192/256.c` (unused vars/params, format specifiers, `sizeof` vs literal buffer length, etc.).

- All @ldesnogu's, cherry-picked from #34 with original authorship preserved; no content changes.
- Verified: nosimd build clean, LL self-test Res64 `71E61322CCFB396C` unchanged.

This is the base of a 3-PR stack from #34's remaining "core files" cluster:
1. **this PR** — pre-rename warning cleanups
2. the `cstr`/`in_line` global renames (stacked on this)
3. post-rename cleanups incl. the factor.c AVX-512 sieve change

It's independent and can merge first; the other two build on it.

PR #133, opened 2026-07-21
factor.c/Mlucas.c post-rename warning cleanups (from #34)

MarkRose:
Part of splitting @ldesnogu's warnings-cleanup PR #34. **Stacked on #132 (which is stacked on #131)** — merge order: #131 → #132 → this.

⚠️ Until #131/#132 merge, this PR's diff also shows their commits. **The 7 commits new here** (they depend on the renames in #132):
- `3d46403` Mlucas.c: fix strncpy warning
- `4fd24cd` factor.c: fix singlethread compilation  ← **see note**
- `2f6536d` factor.c: remove warnings
- `f28e202` Mlucas.c/factor.c/Mdata.h: format warnings
- `3a965da` mi64.c: make mi64_div_y32 source array const
- `708a8ef` factor.c: remove copy warnings
- `24f651a` util.c: fix warnings found on Windows

**Reviewer note:** `4fd24cd` is not a pure warning fix — it adds a `psmall` parameter to the sieve function's signature under `#if defined(USE_AVX512)` to fix single-threaded compilation. It's a functional/interface change on the AVX-512 factoring path; worth verifying independently.

- All @ldesnogu's, cherry-picked from #34 with original authorship preserved.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -995,7 +995,8 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 
 						// Copy all but the final (pm1_done) char of the assignment into g_cstr and append pm1_done = 1. If g_in_line ends with newline, first --j:
 						j = strlen(g_in_line) - 1;	j -= (g_in_line[j] == '\n');
-						strncpy(g_cstr,g_in_line,j);	g_cstr[j] = '\0';	strcat(g_cstr,"1\n");
+						// Note we over copy with strcpy, but since the end of string is explictly set, that's no issue.
+						strcpy(g_cstr, g_in_line);	g_cstr[j] = '\0';	strcat(g_cstr,"1\n");
 						split_curr_assignment = TRUE;	// This will trigger the corresponding code following the goto:
 						goto GET_NEXT_ASSIGNMENT;
 					}
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -86,9 +86,9 @@ uint32 NERR_GCHECK = 0;	// v20: Add counter for Gerbicz-check errors encountered
 int ITERS_BETWEEN_GCHECK_UPDATES = 1000;	// iterations between Gerbicz-checkproduct updates
 int ITERS_BETWEEN_GCHECKS     = 1000000;	// #iterations between Gerbicz-checksum residue-integrity checks
 
-char ESTRING[STR_MAX_LEN];	// Mersenne exponent or Fermat-number index in string form - for M(p) this == p, for F(m) this == m
-char BIN_EXP[STR_MAX_LEN];	// Binary exponent in string form - for M(p) this == p, for F(m) this == 2^m
-char PSTRING[STR_MAX_LEN];	// Modulus being used in string form, e.g. "M110916929" and "F33".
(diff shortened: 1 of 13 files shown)


src/Mlucas.c
```diff
@@ -995,7 +995,8 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 
 						// Copy all but the final (pm1_done) char of the assignment into g_cstr and append pm1_done = 1. If g_in_line ends with newline, first --j:
 						j = strlen(g_in_line) - 1;	j -= (g_in_line[j] == '\n');
-						strncpy(g_cstr,g_in_line,j);	g_cstr[j] = '\0';	strcat(g_cstr,"1\n");
+						// Note we over copy with strcpy, but since the end of string is explictly set, that's no issue.
+						strcpy(g_cstr, g_in_line);	g_cstr[j] = '\0';	strcat(g_cstr,"1\n");
```
tdulcet: I believe the correct fix is to use `strncpy` with the buffer length, as otherwise we could theoretically risk a buffer overflow.

src/factor.c
```diff
@@ -1832,7 +1833,7 @@ Fermat Fn (n > 0): 0,Acceptable km-values for the ? possible pm (= p%60) values:
 	pm = 16: 0, 6, 8,14,18,20,24,26,30,36,38,44,48,50,54,56	<*** F36 factor has k = 20 ... why do I miss? ***
 	pm = 32: 0, 4,10,12,18,22,24,28,30,34,40,42,48,52,54,58
 */
-	i = CHECK_PKMOD60  (&itmp64,1, k, incr);
+	i = CHECK_PKMOD60  (&itmp64,1, 0/* k */, incr);
```
tdulcet, line 1840: Why are we changing this value to zero?

src/factor.c
```diff
@@ -189,14 +189,14 @@ int restart;
...
-	char PSTRING[STR_MAX_LEN];	/* [exponent | index] of [Mersenne | Fermat ] Number being tested in string form, typically
+	char ESTRING[STR_MAX_LEN];		/* Exponent in string form */
+	char PSTRING[STR_MAX_LEN+1];	/* [exponent | index] of [Mersenne | Fermat ] Number being tested in string form, typically
 								 estring concatenated with other descriptors, e.g. strcat("M",estring) for Mersenne case
 								*/
 	uint64 PMIN;	/* minimum #bits allowed for FFT-based mul */
 	uint64 PMAX;	/* maximum #bits allowed depends on max. FFT length allowed
 					  and will be determined at runtime, via call to given_N_get_maxP(). */
-	char cbuf[STR_MAX_LEN*2], g_cstr[STR_MAX_LEN];
+	char cbuf[STR_MAX_LEN*2+200], g_cstr[STR_MAX_LEN];
```
tdulcet: This is already a large buffer, but maybe `*3` instead of the more arbitrary `*2+200`. This also occurs below.

tdulcet (commented):
Thanks, this looks good, but it will need to wait for #131 and #132 in order to be merged.

Note that the `psmall` changes were also made in #83.

tdulcet (dismissed):

tdulcet (dismissed):

tdulcet (dismissed):
Thanks, hopefully the second time is the charm...

MarkRose:
Hmm. Very odd. Since you approved I'll bypass the rules.

Issue #98, opened 2026-07-15
Allow Fermat number testing on ARM


tdulcet:
We should allow Fermat number testing on ARM when using the ASIMD build mode. For more information, see: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/docs/Fermat-testing.md#L12-L19

This is currently disabled with an assert: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L1202-L1204 However, after removing that assert, it seems to at least partially work. Running the `config-fermat.sh` script does not produce any errors and it generates a `fermat.cfg` file:
```
21.0.2
         4  msec/iter =    0.29  ROE[avg,max] = [0.000000000, 0.000000000]  radices =   8 16 16  0  0  0  0  0  0  0    p = 16: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000003,           3,           3
         7  msec/iter =    0.50  ROE[avg,max] = [0.000000000, 0.000000000]  radices =  28  8 16  0  0  0  0  0  0  0    p = 17: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000000,           0,           0
         8  msec/iter =    0.38  ROE[avg,max] = [0.000000000, 0.000000000]  radices =   8 32 16  0  0  0  0  0  0  0    p = 17: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000003,           3,           3
        14  msec/iter =    0.55  ROE[avg,max] = [0.000000000, 0.000000000]  radices =  28 16 16  0  0  0  0  0  0  0    p = 18: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000001,           1,           1
        15  msec/iter =    0.60  ROE[avg,max] = [0.000000000, 0.000000000]  radices =  60  8 16  0  0  0  0  0  0  0    p = 18: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000001,           1,           1
        16  msec/iter =    0.60  ROE[avg,max] = [0.000000000, 0.000000000]  radices =  32 16 16  0  0  0  0  0  0  0    p = 18: 100-iter Res mod 2^64, 2^35-1, 2^36-1 = 0000000000000001,           1,           1
[...]

PR #135, opened 2026-07-22
Verify self-test residues for command-line exponents / Fermat indices

### LINKED ISSUE #123: Self-test of specific exponent does not verify residues
For Mersenne numbers, when doing a self-test of a specific exponent that is included in the built-in self-test residue matrix, it does not verify that the resulting residues are correct. This was reported by Ken on [the forum](https://www.mersenneforum.org/node/1082613?p=1120145#post1120145):
> If one specifies an exponent for a selftest run such as
> `./Mlucas -fft 3K -iters 10000 -m 66431 -cpu 0 -prp 3 &>3K-10kiters-selftestcheck.txt`
> the result is correct, except the program incorrectly warns as follows, even when the specified exponent is exactly the default exponent, for selftest residue data already built into the program.
> Calling the default exponent non-default and warning about it seems like a bug to me. It's warning A != A, in effect.
> Presumably, due to the false alarm, it then skips checking the residues obtained in the run against the tabulated values.
> ```
> ********** Non-default exponent - you will need to manually verify that the residue **********
> ********** triplets output for this self-test match for all FFT radix combinations! **********
> ```

For Fermat numbers, the only way to run the self-test is by specifying a specific exponent (e.g. via the `config-fermat.sh` script), which seems to imply that the resulting residues are never verified. This was shown in #98, where every self-test produced bad residues, but they were still written to the `fermat.cfg` file.

https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/src/Mlucas.c#L4502-L4512

MarkRose:
Resolves #123.

## Problem
A self-test exponent given via `-m` (Mersenne) or `-f` (Fermat) is stored in the empty **0-pad slot** of the `MvecPtr[]` / `FermVec[]` reference-residue arrays, whose tabulated residues are `0`. The verification path decides "is this brand-new data we can't check?" from that zero residue (`new_data = TRUE`), so a command-line self-test was **always** treated as new data:

- Its computed residues were **never compared** against the built-in reference table.
- It printed the spurious `***** Non-default exponent - you will need to manually verify ... *****` warning even when the exponent was *exactly* the tabulated default for that FFT length (as the reporter put it, "warning A != A").
- For **Fermat** numbers, `-f` is the *only* way to run the self-test (there is no default-exponent path), so Fermat self-test residues were **never verified at all** — which is why every run in #98 happily wrote bad residues to `fermat.cfg`, and why `config-fermat.sh` cannot currently catch a wrong result.

## Fix
Before dispatching a command-line self-test, look up the specified exponent / Fermat-index **and** FFT length in the reference table. On a match, redirect the test (`start`/`finish`) to that populated entry so the existing residue-verification path runs, and clear the bogus "non-default" flag. Non-tabulated exponents/lengths are unaffected (still treated as new data). `config-fermat.sh` invokes `Mlucas -f N -fft F -iters …`, so its tabulated (index, length) combos now verify.

The PRP reference table (`MvecPRP[]`) is mostly unpopulated (zero residues), so `-prp` self-tests of a tabulated exponent still can't be verified — but the false warning is now correctly suppressed. Filling in PRP reference residues is a separate data task.

## Verification (nosimd)
- `-m 66431 -fft 3K -iters 100`: no longer prints the "non-default" warning, and verifies against `MersVec[3K]`.
- `-f 16 -fft 4K -iters 100`: now verifies against `FermVec` (previously new-data / unchecked).
- **Proof verification is genuinely active:** corrupting the `FermVec` F16 reference residue makes the self-test correctly report `*** Res64 Error ***`, where before the fix it passed silently (the #98 failure mode).
- Normal `-fftlen 192` self-test (no `-m`) unchanged: `Res64: 71E61322CCFB396C`.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -4380,6 +4380,30 @@ just below the upper limit for each FFT lengh in some subrange of the self-tests
 	// In fact it eases the logic to explicitly set selfTest = TRUE whenever iters is set (but not nec. the converse), so do that here:
 	if(iters) selfTest = TRUE;
 
+	// A user-specified self-test exponent (-m) or Fermat index (-f) is stored in the empty 0-pad slot of
+	// MvecPtr[]/FermVec[], whose reference residues are 0. The residue-verification path (below) then treats
+	// it as brand-new data (new_data = TRUE) and never checks the computed residues against the built-in
+	// reference table - even when the exponent/index + FFT length exactly match a tabulated entry (the bogus
+	// "Non-default exponent" warning is the symptom). For Fermat numbers, -f is the only way to run the
+	// self-test, so their residues were never verified at all (see #98/#123). If the user-specified
+	// exponent/index and FFT length match a reference-table entry, redirect the self-test to that entry so
+	// the residues are actually verified against the tabulated values.
+	if(selfTest && userSetExponent) {
+		if(modType == MODULUS_TYPE_FERMAT) {
+			for(i = 0; i < numFerm; i++) {
+				if(FermVec[i].Fidx == FermVec[numFerm].Fidx && FermVec[i].fftLength == FermVec[numFerm].fftLength) {
+					start = i; finish = i+1; userSetExponent = 0; break;
+				}
+			}
+		} else {
+			for(i = 0; i < numTest; i++) {
+				if(MvecPtr[i].exponent == MvecPtr[numTest].exponent && MvecPtr[i].fftLength == MvecPtr[numTest].fftLength) {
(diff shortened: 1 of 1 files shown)


tdulcet (approved):
Thank for fixing this. It looks good to me.

> The PRP reference table (`MvecPRP[]`) is mostly unpopulated (zero residues), so `-prp` self-tests of a tabulated exponent still can't be verified — but the false warning is now correctly suppressed.

The missing self-test residues will be added in #253.

PR #138, opened 2026-07-22
factor: size k_to_try for the widest SIMD modpow init-warmup (fix OOB read)

MarkRose:
## Problem

The one-time *"init-calls to inline-asm-using modpow functions"* in `main()` (`factor.c`, in the `MULTITHREAD` branch) pass `k_to_try` to `twopmodq78_3WORD_DOUBLE_q8/q16/q32/q64`.

`q32` and `q64` read a full 32- and 64-wide batch of k-values from that array **at function entry** — ahead of the `if(init_sse2) return 0;` early return that the init-mode call takes:

```c
for(j = 0; j < 32; j++) {
    ASSERT((k[j] >> 52) == 0ull, "Ks must be < 2^52!");
    kdbl[j] = (double)k[j];
}
```

`q8` and `q16` are not like that: their `init_sse2` early return comes before their first `k[]` read, so in init mode they never touch the array. (The `k[]` reads near the top of `q16` are inside `#if FAC_DEBUG`.)

But `k_to_try` is allocated with only `TRYQ * NTHREADS` entries, and `TRYQ` is 8 or less in every configuration (4 in the common one), so the q32/q64 warmup calls read **past the end of the buffer** — 28 and 60 elements OOB respectively at `TRYQ = 4, NTHREADS = 1`.

The values don't matter to the warmup (it runs with `p = 0`), but the read itself is out of bounds. On a host whose heap past the allocation isn't zeroed, a stray `k[j] >= 2^52` trips the AVX-512 precondition

```c
ASSERT((k[j] >> 52) == 0ull, "Ks must be < 2^52!")
```

and aborts. This reproduces on Windows/MinGW **AVX-512** CI runners; it's masked elsewhere only because the adjacent heap usually reads back as zero.

## Fix

Size the zero-filled allocation to `MAX(TRYQ * NTHREADS, 64)` (widest warmup = q64 = 64) at both alloc sites (single- and multi-threaded). The extra slots are zero, so the warmup reads stay valid.

## Testing

Reproduced deterministically on Linux under Intel SDE with a dirtied heap:

```
$ MALLOC_PERTURB_=255 sde64 -skx -- ./Mfactor -m 127 -bmax 20
  ...twopmodq78_3WORD_DOUBLE_q32: Setting up...
  Assertion '(k[j] >> 52) == 0ull' failed: Ks must be < 2^52!     # before
  Factoring self-tests completed successfully.                    # after fix
```

nosimd / AVX2 unaffected — q32/q64 are only called under `#ifdef USE_AVX512`, and the q8/q16 calls that the SSE2/AVX builds make return before reading `k[]`. The change only enlarges a zeroed allocation in any case.

### DIFF
--- a/src/factor.c
+++ b/src/factor.c
@@ -1005,7 +1005,7 @@ Others are optional and in some cases mutually exclusive:
 		// TRYQ*NTHREADS-sized (TRYQ can be 4) buffer is over-read. The extra slots are zero (calloc), so
 		// the warmup reads stay valid. Undersized here => OOB read of uninitialized heap, which trips the
 		// 'Ks must be < 2^52' assertion in q32/q64 on hosts whose heap past the alloc isn't zeroed.
-		k_to_try = (uint64 *)calloc(MAX(TRYQ * NTHREADS, 64), sizeof(uint64));
+		k_to_try = (uint64 *)CALLOC(MAX(TRYQ * NTHREADS, 64), sizeof(uint64));
 	#else
 		MAX_THREADS = get_num_cores();
 		ASSERT(MAX_THREADS > 0, "Illegal #Cores value stored in MAX_THREADS");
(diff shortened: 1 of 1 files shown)


tdulcet (dismissed):

Note that `calloc` will need to be converted to `CALLOC` as noted in #88.

MarkRose:
Done — both `k_to_try` sites now use `CALLOC`.

Rebased onto current main first: this branch forked 301 commits back, before #88, so `CALLOC` didn't exist on it and the change wouldn't have compiled standalone. The `free()` is unaffected, since `CALLOC` still hands back a plain `malloc`'d pointer.

Verified: `makemake.sh mfac avx2` rc=0, `Mfactor -h` rc=0, and the built-in factoring self-tests complete ("Factoring self-tests completed successfully") — that's the path exercising the `k_to_try` init-warmup this PR sizes.

One thing you'll see if you run it further: `-m 127 -bmax 16` then aborts with `There were errors writing the savefile`. That's #114, not this change — plain main built the same way aborts identically on the same command.

Sorry, the push dismissed your approval — re-requesting.

tdulcet (approved):

PR #139, opened 2026-07-23
mi64: fix out-of-bounds writes and latent memory-safety traps

MarkRose:
Findings from a systematic memory-safety audit of the Mfactor-linked sources. Two commits:

**Commit 1 — confirmed OOB writes + latent traps** (each OOB item reproduces as an ASan heap-buffer-overflow with exactly-sized buffers):
1. `mi64_div_by_scalar64_u2/_u4`: in-place zero-padding wrote `x[lenu]`(`.._u2`, on **every** call) / `x[lenu..lenu+2]`(`_u4`), and wrote the quotient `y[]` with the padded length — up to 3 words past the caller's exactly-sized arrays (all in-tree callers pass exactly-`lenu`-word arrays). Fixed by dispatching non-multiple-of-fold lengths to the narrower variants and deleting the pad machinery — the folded routines now never touch memory outside `[0, lenu)`.
2. `mi64_div_mont`: single-word-remainder restore writes `rem_save[nws]` with `nws == lenD` possible (e.g. divisor `3*2^65`) — one word past the scratch section; +1 pad word. Also guarded an `(uint64)-1 >> 64` UB for `nshift % 64 == 0`.
3. `mi64_twopmodq`: static q-buffer reallocs sized by `lenQ` but gated on `len` (skip-then-overrun trap); a `pshift` re-realloc that could silently shrink it; lead-bits extraction over the nominal rather than effective length (stale-word read after realloc).
4. `mi64_shrl`: the documented "fudge factor" (`output_len` one word short) was never implemented — every path wrote one word past the buffer; the 0-shift path passed a *negative count* to `mi64_clear` (≈2^32-word wipe). All writes now clamped.
5. `mi64_is_div_by_scalar64_x4`: inverted ASSERT (`len == 0`) aborted every legitimate call.
6. `mi64_add`, ASM variant: added a `len != 0` guard, because the ASM loop counts down with `decq` before its first termination test and would therefore treat `len == 0` as 2^64 iterations. Note that this does **not** restore parity with the C variant, contrary to what an earlier revision of this description said: the live non-`YES_ASM` `mi64_add` — the `#elif 1` 2-folded loop — has no such assert and silently returns 0 for `len == 0`. Its matching assert is dead code inside the preceding `#if 0` arm. So after this change the ASM variant aborts where the C variant returns 0; making the C side match is a separate call for the maintainer.
7. `mi64_vcvtuqq2pd`: `vmovaps` → `vmovups` (caller arrays carry no 64-byte-alignment contract).

**Commit 2 — asm-operand hardening:** converted every asm-written operand declared as a plain input-`"m"` to a `"+m"` output, following the file's own Jan-2025 precedent where this exact pattern caused a real miscompile. 17 asm blocks across 7 functions: `radix_power64` (1), `mi64_is_div_by_scalar64` (2), `mi64_is_div_by_scalar64_x4` (1), `mi64_is_div_by_scalar64_u2` (1), `mi64_is_div_by_scalar64_u4` (2), `mi64_div_by_scalar64_u2` (4), `mi64_div_by_scalar64_u4` (6). All blocks use named operands; none positional.

## Testing
- ASan harness (exactly-sized buffers; 1-word-divisor `mi64_div` at lengths 6/7/70, `mi64_div` by `3*2^65` with no remainder array, `mi64_shrl` fudge cases): red on pristine main at the first case, all green after.
- Semantic check `quotient*q + rem == dividend` across nosimd/AVX2 × `-O1`+ASan/`-O3`: byte-identical remainders in every config.
- Warning-neutral at `-O2/-O3 -Wall`; AVX-512 compile clean; Mfactor self-test passes on the integration branch carrying these commits.

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -614,11 +614,19 @@ uint64	mi64_shrl(const uint64 x[], uint64 y[], uint32 nshift, uint32 len, uint32
 	So allow output_len to be 1 limb smaller than 17 as a fudge factor to handle arbitrary in-word copy-bit boundaries:
 	*/
 	ASSERT(output_len >= (len-nwshift)-1, "mi64_shrl: output_len must be large enough to hold result!");
+	/* v21: the "fudge factor" case output_len == (len-nwshift)-1 permitted by the above assert was
+	documented ("only copy that many and exit") but never implemented: every write path below wrote
+	the full (len-nwshift) result words, i.e. one word past the caller's buffer. All writes are now
+	clamped to output_len. */
 	// Special-casing for 0 shift count:
 	if(!nshift) {
 		if(x != y) {
-			mi64_set_eq(y, x, len);				// Set y = x...
-			mi64_clear(y+len, output_len - len);// ...and 0 any excess high limbs in y.
+			i = MIN(len, output_len);
+			mi64_set_eq(y, x, i);				// Set y = x...
+			if(output_len > len)
+				mi64_clear(y+len, output_len - len);// ...and 0 any excess high limbs in y.
+				// (Note the guard: the old unguarded form passed a huge unsigned count if
+				// output_len < len, wiping (2^32 - small) words beyond the buffer.)
 		}
 		return 0ull;
 	}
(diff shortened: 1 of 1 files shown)


MarkRose:
Cleared the unused-variable warnings this branch introduces (#244). Declaring the asm-written operands as `+m` outputs left eleven variables unreferenced in `mi64_div_by_scalar64_u2/u4`.

They are not all the same case, and deleting all eleven would have broken every non-x86 build. `mi64.h` has a `NO_ASM` control that selects the portable-C arm, so compiling the file both ways separates them exactly:

| | with asm | without asm | |
| --- | --- | --- | --- |
| u2 `i`, `j` | unused | unused | removed |
| u4 `j`, `itmp64`, `ptr_incr`, `ptr_inc2` | unused | unused | removed |
| u4 `i`, `len2` | unused | **used** | kept, moved into the `#ifndef YES_ASM` block |
| u4 `iptr1`, `iptr2`, `iptr3` | unused | **used** | kept, moved |

So six were dead on both paths and are gone; the five the portable-C arm still needs are now declared inside the block that uses them, which silences the warning without removing anything a non-asm target depends on.

Verified both arms — `gcc -Wall` with and without `-DNO_ASM` compiles with zero errors and none of the eleven warnings. What is left in either case is pre-existing on `main` (`mi64.c`'s `i` at 4008, and the `BLOCKLENM1`/`BASEADDRMASK`/`minlen` set that only the NO_ASM arm reports); those belong to the separate `main` patch. Full avx2 build clean, `-s tt` rc=0 with 23 residues.

MarkRose:
Pushed a follow-up: `mi64.c` did not compile on this branch for any x86-64 build below AVX2.

The declaration-block restructuring narrowed `iptr1`/`iptr2` and `len2` to `#ifndef YES_ASM` and dropped `ptr_incr`/`ptr_inc2` entirely, but the `YES_ASM && !USE_AVX2` arm of `mi64_div_by_scalar64_u4()` uses all five:

```
src/mi64.c:7654: error: 'iptr1' undeclared (first use in this function)
src/mi64.c:7654: error: 'iptr2' undeclared
src/mi64.c:7657: error: 'ptr_incr' undeclared
src/mi64.c:7658: error: 'ptr_inc2' undeclared
src/mi64.c:7661: error: 'len2' undeclared
```

That is this branch's own head, not a merge artifact - I reproduced it by compiling `src/mi64.c` straight off it.

`len2` is now declared under "portable-C arm or pre-AVX2 asm arm", and the four pointers/increments in a `YES_ASM && !USE_AVX2` arm of their own. Full builds after the change:

| mode | result |
|---|---|
| nosimd | rc=0 |
| sse2 | rc=0 |
| avx | rc=0 |
| avx2 | rc=0 (control - this mode was never affected) |

Found while building the integration branch, which carries this PR; the fix was applied there first and is now here at source.

PR #140, opened 2026-07-23
util: fix CORE_SET bitmap bound check; declare missing asm clobbers in TEST_SIMD code

MarkRose:
1. **`parseAffinityTriplet` OOB write (real bug):** `CORE_SET` is `uint64[MAX_CORES>>6]` (16 words), but the bound assert checked `word < MAX_CORES` — permitting `-cpu` indices ≥ 1024 to write up to ~8KB past the global array (silent corruption whose victim depends on link layout). Now checks `word < (MAX_CORES>>6)`.
2. **Missing asm clobbers in TEST_SIMD-only code:** `test_mul50x50` writes `rdx` undeclared; `test_simd_transpose_8x8`'s AVX-512 block writes `zmm1/zmm2` undeclared. Both test-only (gated/commented call sites), fixed for correctness.

## Testing
Warning-count-neutral vs main at `-O2 -Wall`; the AVX-512+TEST_SIMD config's pre-existing unrelated errors are byte-identical before/after (no new diagnostics); the fixed transpose block compiles standalone. Integration branch carrying this passes the Mlucas/Mfactor self-tests.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -3206,7 +3206,7 @@ I = 981 Needed extra sub: a = 916753724; p = 11581569; pinv = 370 [a/p = 79.1562
 					 ,[__binv] "m" (dptr2)\
 					 ,[__crnd50] "m" (dptr3)\
 					 ,[__prod1_adj] "m" (dptr4)\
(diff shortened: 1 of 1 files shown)


tdulcet (approved):
Thanks.

PR #141, opened 2026-07-23
twopmodq80: add missing asm memory clobbers; fix FAC_DEBUG and dead-AVX512_I landmines

MarkRose:
1. **Missing `"cc","memory"` clobbers (live code):** the `SSE2_twopmodq78_modmul_q8` (AVX/AVX2) and both `SSE2_twopmodq78_modmul_q4` macro variants store into the `Xfq0` data region through `%%rdx` (e.g. `vmovaps %%ymm3,0x1c0(%%rdx)`) but declared no `"memory"` clobber — formally invalid asm that currently survives only because neighbouring blocks' clobbers bracket it. Every sibling macro (q16/q32/q64) already declares `"cc","memory"`; these three now match.
2. **FAC_DEBUG compile error (twopmodq78_3WORD_DOUBLE_q16):** `k[0] = 67885719840ull` inside a condition — `=` should be `==` (invalid-lvalue hard error in any FAC_DEBUG build); also guarded with `init_sse2 == FALSE` so init-mode calls don't read `k[0]`.
3. **Dead `USE_AVX512_I` q64:** `mask_lo52` was written at `sc_ptr+0x310` but read at `+0x308`; unified to `0x308` (consts pack one 8-word register apart, matching the q32 counterpart's consistent `0x180→0x188` spacing; `0x310` left a phantom gap).

## Testing
Compile-clean and warning-count-neutral vs main in nosimd/SSE2/AVX2/AVX-512 modes. FAC_DEBUG AVX2 build: pristine main fails at the `=` typo, fixed tree compiles with 0 errors. Integration branch carrying this passes Mlucas/Mfactor self-tests.

### DIFF
--- a/src/twopmodq80.c
+++ b/src/twopmodq80.c
@@ -4840,7 +4840,7 @@ if(~pshift != p+78) {
 		uint64 twopmodq78_3WORD_DOUBLE_q16(uint64 p, uint64 k[], int init_sse2, int thr_id)
 		{
 		#if FAC_DEBUG
-			int dbg  = (p==17715697 && k[0] = 67885719840ull);
+			int dbg  = (init_sse2 == FALSE) && (p==17715697 && k[0] == 67885719840ull);	// == not =; guard k[] deref against init-mode calls
 		#endif
 			const char func[] = "twopmodq78_3WORD_DOUBLE_q16";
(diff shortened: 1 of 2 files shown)


tdulcet (approved):
Thanks.

Note that the `USE_AVX512_I` build mode is not finished or usable, so it cannot be tested.

PR #142, opened 2026-07-23
twopmodq128-256: guard lead-bits extraction against shift-by-64 UB when j == 0

MarkRose:
Common fix across the scalar twopmodq files: the powering-loop lead-bits extraction computes `((TOP << j) + (NEXT >> (64-j))) >> 56` after `j = leadz64(TOP)`. When TOP's MSB is set, `j == 0` and `NEXT >> 64` is undefined behaviour — x86 masks the count to `>> 0`, silently injecting the full lower word into the lead bits (wrong `start_index`/`zshift` → wrong residue), and other targets/compilers may do worse.

Guarded the cross-word term with `(j ? (NEXT >> (64-j)) : 0ull)` at every affected site:
- twopmodq128.c — 5 sites; twopmodq192.c — 8 sites (incl. `_q4/_q4_qmmp/_q8`); twopmodq256.c — 3 sites (d3/d2/d1 cascade); twopmodq160.c — 1 site (its `pshift.d2` site is provably `j >= 32` per its own ASSERT and was correctly left alone).
- twopmodq128_96.c / twopmodq96.c use single-word pshift (no cross-word term) — untouched.

Not triggered by current self-test data (exponent magnitudes keep `j ≥ 10`), but reachable for moduli sized right at a word boundary; found while chasing a case where garbage in a struct's top word set the MSB.

## Testing
All four files compile warning-count-neutral vs main; the integration branch carrying this passes the full Mfactor `test_fac` (64/128/192/256-bit Fermat + 63-96-bit factor sections) and Mlucas self-tests.

### DIFF
--- a/src/twopmodq128.c
+++ b/src/twopmodq128.c
@@ -248,7 +248,8 @@ if(dbg) printf("twopmodq128:\n");
 			j = leadz64(pshift.d1);
 			start_index = 128-j-7;
 			/* Extract leftmost 7 bits of pshift and subtract from 127: */
-			zshift = 127 - (((pshift.d1<<j) + (pshift.d0>>(64-j))) >> 57);
+			/* j==0 => shift-by-64 UB; guard the cross-word term: */
+			zshift = 127 - (((pshift.d1<<j) + (j ? (pshift.d0>>(64-j)) : 0ull)) >> 57);
 		} else {
 			j = leadz64(pshift.d0);
 			start_index =  64-j-7;
@@ -462,7 +463,7 @@ if(dbg) printf("twopmodq128x2:\n");
 			j = leadz64(pshift.d1);
 			start_index = 128-j-7;
 			/* Extract leftmost 7 bits of pshift and subtract from 127: */
-			zshift = 127 - (((pshift.d1<<j) + (pshift.d0>>(64-j))) >> 57);
+			zshift = 127 - (((pshift.d1<<j) + (j ? (pshift.d0>>(64-j)) : 0ull)) >> 57);
 			pshift.d1 = ~pshift.d1;
 		}
 		else
@@ -647,7 +648,7 @@ uint64 twopmodq128x2B(uint64 *p_in, uint128 q)
 			j = leadz64(pshift.d1);
 			start_index = 128-j-7;
 			/* Extract leftmost 7 bits of pshift and subtract from 127: */
-			zshift = 127 - (((pshift.d1<<j) + (pshift.d0>>(64-j))) >> 57);
+			zshift = 127 - (((pshift.d1<<j) + (j ? (pshift.d0>>(64-j)) : 0ull)) >> 57);
 			pshift.d1 = ~pshift.d1;
 		} else {
 			j = leadz64(pshift.d0);
@@ -766,7 +767,7 @@ uint64 twopmodq128_q4(uint64* p_in, uint64 k0, uint64 k1, uint64 k2, uint64 k3)
 		if(pshift.d1)
 		{
 			j = leadz64(pshift.d1);
-			lead7 = (((pshift.d1<<j) + (pshift.d0>>(64-j))) >> 57);
+			lead7 = (((pshift.d1<<j) + (j ? (pshift.d0>>(64-j)) : 0ull)) >> 57);
 			start_index = 128-j-7;
 			pshift.d1 = ~pshift.d1;
 		}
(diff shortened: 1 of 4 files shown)


MarkRose:
**Here is an observable consequence, and a regression test for it.**

This PR's description says the UB is "not triggered by current self-test data... but reachable for moduli sized right at a word boundary", and argues the case on language grounds. That is accurate but undersells it — the reachable region is large and ordinary, not a corner:

For `twopmodq192`, `j = leadz64(pshift.d2)`, so `j == 0` for **every p with bit 127 set** — i.e. 99.6% of exponents in `[2^127, 2^128)` hit the undefined shift. On x86 the count is masked to `>> 0`, which injects the entire next word into the lead bits, giving a wrong `lead8` and hence a wrong `start_index`/`zshift`.

The two 127-bit values anyone would reach for as test cases — `MM127` and `F127` — both escape by luck, which is a fair explanation for why this never showed up in the existing self-tests.

**It now has coverage.** While building constructed `twopmodq192_q4` vectors above 128 bits (there was previously *no* `test_fac()` coverage of that routine above 128 bits at all — every catalogued factor lands in the safe regime), 24 entries / 96 lanes were added spanning p = 118–191 bits. Against a stock `makemake.sh mfac 3word` build:

| tree | entries wrong |
|---|---|
| stock `main` | 9 / 24 |
| + this PR | 7 / 24 |
| + this PR + #167 | **0 / 24** |

So this PR is directly responsible for 2 of the 9, and the new loop is a regression test for it. That work is on a separate branch and will be filed once its two prerequisites land; it must merge after this one and after #167.

No changes requested — posting because "UB with no named consequence" and "wrong `start_index` for 99.6% of 127-bit exponents, now under test" are likely to be prioritised differently.

src/twopmodq128.c
```diff
@@ -248,7 +248,8 @@ if(dbg) printf("twopmodq128:\n");
 			j = leadz64(pshift.d1);
 			start_index = 128-j-7;
 			/* Extract leftmost 7 bits of pshift and subtract from 127: */
-			zshift = 127 - (((pshift.d1<<j) + (pshift.d0>>(64-j))) >> 57);
+			/* j==0 => shift-by-64 UB; guard the cross-word term: */
+			zshift = 127 - (((pshift.d1<<j) + (j ? (pshift.d0>>(64-j)) : 0ull)) >> 57);
```
ldesnogu, line 252: This could also be written this way:

`zshift = 127 - ((j ? ((pshift.d1<<j) + (pshift.d0>>(64-j))) : pshift.d1) >> 57);`

Matter of preference. Your fix is fine.

All the uses of this snipped below cry for code sharing :)

ldesnogu (commented):
The fix is fine (with a comment just in case).

src/twopmodq128.c
MarkRose: Thanks @ldesnogu — agreed on both counts.

On the form: yours is equivalent and arguably tidier. I've kept the current one only because it puts the guard on the single term that is actually undefined, so the reason for the ternary is local to it; with yours the reader has to notice that the sum collapses to `pshift.d1` when `j == 0`. Pure preference, and I don't feel strongly.

On code sharing: you're right, and it's more than cosmetic. The idiom occurs 18 times across `twopmodq128/160/192/256` — 11 on `d0`, 6 on `d1`, 1 on `d2` — and right now the UB guard has to be repeated at every one of them. A single helper would make it impossible to add a site that forgets it. Something like:

```c
// Leftmost nbits of the (hi:lo) window, left-justified by j = leadz64(hi).
// j == 0 must not reach the cross-word shift: a shift by 64 is UB.
#define LEAD_BITS(hi,lo,j,nbits) ( (((hi)<<(j)) + ((j) ? ((lo)>>(64-(j))) : 0ull)) >> (64-(nbits)) )
```

`lead7`/`lead8` are just `nbits` 7 and 8, so every site collapses to one call. I'd rather do that as a separate PR on top of this one, so this stays a minimal correctness fix that's easy to review and land — filed as #241.

One thing worth flagging before you spot it: `twopmodq160.c:100` is the only one of the 18 I left unguarded. That's deliberate — `pshift` there is a 192-bit type holding a 160-bit value, so `d2` holds at most the top 32 bits and `leadz64(pshift.d2) >= 32`; `j == 0` can't happen, and there's already a `FAC_DEBUG` assert asserting exactly that. If the helper lands it'll get the guard anyway, harmlessly.

ldesnogu: No strong feeling either for the form.

Regarding the code sharing, making a separate change was IMHO the way to go, thanks!

ldesnogu (approved):

PR #143, opened 2026-07-23
twopmodq160: fix Fermat-mod final subtraction and FAC_DEBUG decl order

MarkRose:
The (currently uncalled — all call sites commented out) 160-bit scalar modpow's Fermat-mod path was doubly wrong:
1. Missing the `FERMAT <<= 1;` every sibling routine (twopmodq128/128_96/192/256) applies — Fermat case would subtract 1 instead of 2.
2. `q.d0 -= FERMAT;` had no borrow propagation — Fermat factors routinely have `q.d0 == 1`, so the subtraction wraps without borrowing from `q.d1`. Replaced with the general `SUB160` pattern used by twopmodq192/256 ("carry may propagate > 1 word").

Also moved the FAC_DEBUG `dbg` init below the declaration of `p` it references (use-before-declaration compile error in FAC_DEBUG builds).

Latent (routine has no live callers), but it now matches its 128/192/256 siblings so re-enabling it can't silently produce wrong Fermat-factor verdicts.

## Testing
Compiles warning-count-neutral vs main; pristine main fails to compile with `-DFAC_DEBUG=1` (`'p' undeclared`), fixed tree compiles clean.

### DIFF
--- a/src/twopmodq160.c
+++ b/src/twopmodq160.c
@@ -43,16 +43,16 @@ The key 3-operation sequence here is as follows:
 */
 uint64 twopmodq160(uint64 *p_in, uint64 k)
 {
-#if FAC_DEBUG
-	int dbg = STREQ(&char_buf[convert_uint192_base10_char(char_buf, p)], "0");
-	uint160 y,z;
-#endif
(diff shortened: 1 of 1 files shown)


tdulcet (approved):
Thanks.

PR #144, opened 2026-07-23
twopmodq96: repair dead #if-0 q8 variant's layout arithmetic (kept disabled)

MarkRose:
The `#if 0`-gated `twopmodq96_q8` ("Disable until regain GCC 4.6+ functionality") contains mechanical layout bugs that would be immediate OOB writes if the gate were ever flipped:
- Three mutually inconsistent per-thread stride/alloc conventions (alloc in uint64 units vs layout in 16-byte uint96 slots vs a third stride in the init loop) — reconciled to one convention (0x32 uint96 = 0x64 uint64 = 800 B/thread) across alloc, init and rebase.
- `perm_mask` init wrote the same slot 7 times (`*(ptr32+1) = 7; *(ptr32+1) = 1; ...`) and the matching ASSERT was unsatisfiable (same lvalue required to equal 7 and 1) — both now use consecutive indices.
- Slot-index typos aliasing `x[]` with `qhlf[]` and `lo[]` with `hi[]` — renumbered to the disjoint layout from the routine's own byte-offset map; missing `uint32 *ptr32` declared.
- Setup fprintf said "_q4".

The gate stays `#if 0` with a loud UNTESTED comment: with the gate flipped, the repaired code parses but still needs a `MULH96_PTR_q8` that was never written (the q8 ASM port is unfinished) — this PR only removes the memory-layout traps.

## Testing
Normal builds unaffected (all edits inside the disabled block; warning-count identical to main).

### DIFF
--- a/src/twopmodq96.c
+++ b/src/twopmodq96.c
@@ -1533,6 +1533,14 @@ if(dbg)printf("xout = %s\n", &char_buf[convert_uint96_base10_char(char_buf, x)])
 /*** 8-trial-factor version ***/
 /******************************/
 #if 0//YES_ASM	*** Disable until regain GCC 4.6+ functionality ***
+/******************************************************************************************
+ * WARNING: twopmodq96_q8 below remains DEAD and UNTESTED. Its local-store layout
+ * arithmetic was repaired (2026: alloc/stride units reconciled to 0x64 uint64 = 0x32
+ * uint96 slots/thread, x/lo pointer groups de-aliased from qhlf/hi, perm_mask init and
+ * ASSERT fixed) so that it would no longer perform OOB writes if re-enabled, but the
+ * routine has NOT been compiled into a working build or numerically validated. Do not
+ * trust it until it has been exercised end-to-end.
+ ******************************************************************************************/
 
 	// The'init_sse2' arg is ill-named here as it refers to 'init any local-mem needed for inline-asm routines' whether SSE/AVX used or
 	// not - here we use just x86_64 non-vector instructions - but use same name as in vector-radix* code to show analogous mem-handling:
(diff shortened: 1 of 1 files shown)


tdulcet:
We should probably wait to fix any bugs in dead/unfinished code like this until we decide if it would be worthwhile to finish and enable the code. Without being able to test the code, we may not be making it any better.

In general, Mlucas has a lot of dead code, which we should probably systematically audit and try to determine what Ernst's plans for it all was. Then, we would be in a better position to decide if we should enable or just remove it.

MarkRose:
Agreed — closing this.

Your reasoning holds and I should have weighed it before opening the PR. The code cannot be executed, so a fix to it is unverifiable by construction. Worse, in this particular case the change required *choosing* between three mutually inconsistent stride conventions, and which one Ernst intended is not recoverable from the source. I picked the one that made the surrounding arithmetic self-consistent, but that is a guess dressed up as a repair, and merging it would have replaced "obviously broken" with "plausibly broken", which is harder for the next person to spot.

The findings themselves are worth keeping, so recording them here rather than in a diff. In the `#if 0`-gated `twopmodq96_q8`:

- three mutually inconsistent per-thread stride/alloc conventions — allocation in uint64 units, layout in 16-byte uint96 slots, and a third stride again in the init loop;
- `perm_mask` init writes the same slot seven times (`*(ptr32+1) = 7; *(ptr32+1) = 1; ...`), and the matching ASSERT is unsatisfiable, requiring one lvalue to equal both 7 and 1;
- slot-index typos aliasing `x[]` with `qhlf[]`.

Any of those is an immediate out-of-bounds write the moment that gate is flipped, so whoever revives it should start by assuming the layout arithmetic is wrong rather than trusting it.

I have opened #281 for the systematic audit you suggested, seeded with this as the worked example plus a count of what is actually out there — 350 `#if 0` blocks across 80 files, and five whole source files parked with a `.txt` suffix. It also notes that the blocks are not one kind of thing: this one says "Disable until regain GCC 4.6+ functionality", which is a 2011 condition that may simply have expired, while others are unfinished work or superseded by later design changes. Those want different outcomes, so classification looks like the first useful step rather than a blanket decision.

PR #145, opened 2026-07-23
factor_test: initialize uint256 top words; fix error prints; harden debug configs

MarkRose:
Self-test (`test_fac()`) correctness fixes, found while chasing flaky Windows CI crashes:

**Commit 1:**
- The 256-bit Fermat-factor loop built its uint256 operands with only `.d0-.d2` assigned — `.d3` was whatever the stack held, and `LSHIFT256`/`twopmodq256` read it (twopmodq256 branches on `p.d3` and derives its powering start index from the top word). Benign only where that stack slot happens to be zero. Both operands now zero `.d3`.
- The MM31 known-factor error prints converted the residue into `cbuf0` but printed from the unrelated global `cbuf` (error path only; diagnostic printed garbage).

**Commit 2:**
- The P4WORD+FAC_DEBUG debug block uses `p63`/`cbuf3-6` which are declared nowhere (config could not compile) — declared locally inside the block (deliberately NOT in the top-of-function declaration block, to avoid colliding with #129's rework of it). Ditto `tmp64` in the dead `#if(0)` 192_B block.
- The >256-bit Fermat loop reseeds only word 0 of its 640-word p/q arrays each iteration; `mi64_shl` could pick up a stale high word from the previous iteration for non-word-aligned shifts (benign with the current table, fragile against future entries) — arrays now fully re-zeroed per iteration, plus a defensive ASSERT on the `k << (l-j-1)` shift count.

Known remaining (pre-existing, out of scope, noted for the record): the P4WORD+FAC_DEBUG block's `MUL_LOHI64(p63,p64,&p256.d0,&p256.d1)` doesn't branch on `MUL_LOHI64_SUBROUTINE` like every other call site, so that config still fails under the inline-asm macro variant.

## Testing
Default config compiles byte-for-byte warning-identical to main; `-DP4WORD` compiles clean; the d3 fix's effect was confirmed live on Windows CI (deterministic `q.d3` garbage observed in crash telemetry before the fix). Integration branch carrying this passes the full Mfactor self-test.

### DIFF
--- a/src/factor_test.h
+++ b/src/factor_test.h
@@ -766,12 +766,18 @@ ASSERT(0 == mi64_div_by_scalar64(p, 458072843161ull, i, p), "M7331/458072843161 
 		ASSERT(l >= (j+2), "Power of 2 appearing in factor of Fn must be >= [n+2]!");
 		k =         ffacBig[i].d0;	// Factor k; must be odd in this schema
 		ASSERT(1ull == (k & 1ull), "k must be odd!");
+		// p,q are calloc'd once above and only word 0 is reseeded below; fully re-zero them each pass so
+		// mi64_shl's in-place shift-in doesn't pick up stale nonzero high words from a previous iteration
+		// (harmless overhead here - 2*640 words over ~10 iterations of a self-test):
+		memset(p, 0, 640*sizeof(uint64));
+		memset(q, 0, 640*sizeof(uint64));
 		lenP = (j+63)>>6;	// Assume Fermat index increases as we traverse ffacBig array, thus this overwrites previous
 		p[0] = 1ull;	p[lenP] = mi64_shl(p,p,j,lenP);	lenP += (p[lenP] != 0ull);	// case's p = (1 << j) array elements.
 		lenQ = (l+63)>>6;
 		q[0] = k;		q[lenQ] = mi64_shl(q,q,l,lenQ);	lenQ += (q[lenQ] != 0ull);
 		q[0] += 1;	// q = 2.k.p + 1; No need to check for carry since 2.k.p even
 	//printf("Testing F%u, q = %" PRIu64 " * 2^%u + 1, lenQ = %u...\n",j,k,l,lenQ);
+		ASSERT(l-j-1 < 64, "k << (l-j-1) requires l-j-1 < 64!");	// Defensive: shift count must stay in-range for a uint64
 		uint32 res1 = mi64_twopmodq(p, lenP, k << (l-j-1), q, lenQ, q2);	// Fiddle k to put q in Mersenne-like form = 2.k'.2^j + 1
 			//	res1 = mi64_twopmodq_qferm(j, k << (l-j), q2);
 		if(res1 != 1) {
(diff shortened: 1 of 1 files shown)


src/factor_test.h
```diff
@@ -739,12 +741,18 @@ ASSERT(0 == mi64_div_by_scalar64(p, 458072843161ull, i, p), "M7331/458072843161
...
+		// (harmless overhead here - 2*640 words over ~10 iterations of a self-test):
+		memset(p, 0, 640*sizeof(uint64));
+		memset(q, 0, 640*sizeof(uint64));
 		lenP = (j+63)>>6;	// Assume Fermat index increases as we traverse ffacBig array, thus this overwrites previous
 		p[0] = 1ull;	p[lenP] = mi64_shl(p,p,j,lenP);	lenP += (p[lenP] != 0ull);	// case's p = (1 << j) array elements.
 		lenQ = (l+63)>>6;
 		q[0] = k;		q[lenQ] = mi64_shl(q,q,l,lenQ);	lenQ += (q[lenQ] != 0ull);
 		q[0] += 1;	// q = 2.k.p + 1; No need to check for carry since 2.k.p even
 	//printf("Testing F%u, q = %" PRIu64 " * 2^%u + 1, lenQ = %u...\n",j,k,l,lenQ);
+		ASSERT(l-j-1 < 64, "k << (l-j-1) requires l-j-1 < 64!");	// Defensive: shift count must stay in-range for a uint64
```
tdulcet: Could `l-j-1` ever realistically overflow, without there being a serious bug elsewhere.

MarkRose:
Good question — it can't, and I've reframed the guard to say so rather than leaving it looking like a defence against arithmetic that might go wrong.

Measured across `ffacBig`: of its 119 rows, **57 are actually processed** (the loop stops at the `{0,0,0}` sentinel and breaks at `n > 1000`). For those:

| | |
| --- | --- |
| `l-j-1` range | **1 .. 10** |
| widest `k << (l-j-1)` | **48 bits** |
| rows with `l <= j` | 0 |
| rows with `l-j-1 >= 64` | 0 |

So neither arm can fire on the table as it stands. The minimum of 1 isn't luck: by Lucas, any factor of F_n has the form k·2^pow2 + 1 with pow2 ≥ n+2, so `l-j-1 >= 1` for every well-formed row — and F230 with pow2 = 232 sits exactly on that bound.

Which means the only way to trip it is a **mistyped table row**, and that is worth catching, because `l` and `j` are `uint32`: a row with `pow2 <= n` wraps `l-j-1` to a huge value and shifts by ≥ 64 rather than failing visibly. So I've kept it but made it check the thing that would actually be wrong:

```c
ASSERT(l > j && l-j-1 < 64, "ffacBig row: need pow2 > n, and pow2-n-1 < 64 for the k << (pow2-n-1) below");
```

The `l > j` arm is new — that's the underflow the old form only caught indirectly, after the wrap. The comment now records the observed range so the next reader doesn't have to re-derive it.

Verified: nosimd Mfactor builds clean, `Testing > 256-bit Fermat factors` runs, and `Factoring self-tests completed successfully` — the assert doesn't fire. Also rebased onto current `main`.

tdulcet (approved):
Thanks.

PR #146, opened 2026-07-23
Allow leading FFT radices < 16 in production Fermat runs by forcing residue shift 0 (fixes #119)

### LINKED ISSUE #119: Radix sets with a small radix of 7, 8, or 15 cause errors
There are several known Fermat number testing bugs, which were previously reported by @xanthe-cat on [the forum](https://www.mersenneforum.org/node/22697#post22697). Many of these were fixed in #11 and #14, but this appears to be the most serious remaining one that does not already have a open issue:

> **Radix sets with a small radix of 7, 8, or 15 cause errors**: again mainly affecting the smaller Fermats, if Mlucas finds a radix set with a small first radix (7, 8, or 15) is the fastest, this will be written to the `fermat.cfg` config file, and in production mode errors such as `CY routines with radix < 16 do not support shifted residues!` even if `-shift 0` is specified; substituting a different radix set with 16 or 32 as the first radix removes the problem.
```
src/radix10_ditN_cy_dif1.c:75:          WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix11_ditN_cy_dif1.c:141:         WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix13_ditN_cy_dif1.c:68:          WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix14_ditN_cy_dif1.c:98:          WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix15_ditN_cy_dif1.c:80:          WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix5_ditN_cy_dif1.c:72:           WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix6_ditN_cy_dif1.c:67:           WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix7_ditN_cy_dif1.c:90:           WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix8_ditN_cy_dif1.c:84:           WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
src/radix9_ditN_cy_dif1.c:88:           WARN(HERE, "CY routines with radix < 16 do not support shifted residues!", "", 1);
```

Also see #21 and #101.

MarkRose:
## Problem (#119)

The carry routines for a leading FFT radix < 16 (`radix{5..15}_ditN_cy_dif1.c`) cannot process a nonzero circular residue shift and abort (`CY routines with radix < 16 do not support shifted residues!`). Since v20, Fermat-mod draws a **random** shift whenever the user gives none — so any production run whose `fermat.cfg` landed on a small-leading-radix set (e.g. the 2K `{8,8,16}` set for F13) aborts at iteration 0, and even `-shift 0` doesn't reliably help in paths where the shift is (re)randomized. The only workaround was hand-editing the cfg to a leading radix ≥ 16.

## Fix

In `ernstMain()`'s initial-residue setup, for `RADIX_VEC[0] < 16` force `RES_SHIFT = 0` (exactly as the existing p-1 branch does) and, when a nonzero shift was requested explicitly or would have been applied implicitly, emit an informational note instead of aborting. The defensive guard in the small-radix carry routines stays intact — it now only fires if a genuinely nonzero shift reaches them. The change is modulus-agnostic, so it also covers Mersenne runs landing on small-radix sets.

## Testing (nosimd)

- Before: production Pépin F13 on the 2K `{8,8,16}` set aborts (`mod_square returned with error code[12]`).
- After: same run completes — `Res64,Res35m1,Res36m1 = D79356EC3B040B5E, 3434508623, 52864871946`; identical residues with `-shift 0`; interactive `-f 13 -fft 2` with no shift / `-shift 0` / `-shift 5` all give identical `Res64 = C8FC67EA3A1AC788` (shift-invariant, as expected).
- Mersenne regression `-fftlen 192`: expected `Res64 = 71E61322CCFB396C`.

Resolves #119

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1661,7 +1661,7 @@ READ_RESTART_FILE:
 			// seed straight into a[0] - and if a nonzero shift was requested (explicitly via -shift, or
 			// implicitly via the Fermat-mod random default), emit an informational note rather than aborting:
 			if(parse_cmd_args_get_shift_value() != 0ull) {	// != 0 catches both an explicit nonzero -shift and -1 = "-shift unspecified" (=> would-be random)
-				snprintf(cbuf,STR_MAX_LEN*2, "INFO: Leading FFT radix %u is < 16, which does not support shifted residues; setting residue shift = 0.\n", RADIX_VEC[0]);
+				snprintf(cbuf,sizeof(cbuf), "INFO: Leading FFT radix %u is < 16, which does not support shifted residues; setting residue shift = 0.\n", RADIX_VEC[0]);
 				mlucas_fprint(cbuf,-INTERACT);
 			}
 			RES_SHIFT = 0ull; a[0] = iseed;
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -1650,6 +1650,21 @@ READ_RESTART_FILE:
 		if(TEST_TYPE == TEST_TYPE_PM1) {
 			ASSERT(RES_SHIFT == 0ull, "Shifted residues unsupported for p-1!\n");
 			RES_SHIFT = 0ull; a[0] = iseed;
+		} else if(RADIX_VEC[0] < 16) {
+			// v21 (#119): The CY (carry) routines for a leading FFT radix < 16 do not support nonzero
+			// residue shifts - they WARN + return ERR_ASSERT on RES_SHIFT != 0 (see e.g. the guard at the
+			// top of radix8_ditN_cy_dif1.c). Fermat-mod, however, applies a random default shift whenever
+			// the user specifies none, so a leading-radix-<16 radix set (e.g. the 2K {8,8,16} set) would
+			// abort mid-carry-step even for an otherwise-valid production invocation, and even an explicit
+			// '-shift 0' does not help in the cases where the shift is (re)randomized below. So for a leading
+			// radix < 16, force shift 0 - as in the p-1 branch above, a zero shift simply routes the initial
+			// seed straight into a[0] - and if a nonzero shift was requested (explicitly via -shift, or
+			// implicitly via the Fermat-mod random default), emit an informational note rather than aborting:
+			if(parse_cmd_args_get_shift_value() != 0ull) {	// != 0 catches both an explicit nonzero -shift and -1 = "-shift unspecified" (=> would-be random)
+				snprintf(cbuf,STR_MAX_LEN*2, "INFO: Leading FFT radix %u is < 16, which does not support shifted residues; setting residue shift = 0.\n", RADIX_VEC[0]);
+				mlucas_fprint(cbuf,-INTERACT);
+			}
+			RES_SHIFT = 0ull; a[0] = iseed;
 		} else {
 			// Apply initial-residue shift - if user has not set one via cmd-line or current value >= p, randomly choose a value in [0,p).
 			// [Note that the RNG is inited as part of the standard program-start-sequence, via function host_init().]
--- a/src/fermat_mod_square.c
+++ b/src/fermat_mod_square.c
@@ -410,33 +410,29 @@ int fermat_mod_square(double a[], int arr_scratch[], int n, int ilo, int ihi, ui
 			nradices_radix0 = 2;
 			radix_prim[l++] = 2; radix_prim[l++] = 2; break;
 		*/
-		case 5:
-//			nradices_radix0 = 1;
-			radix_prim[l++] = 5; break;
-		case 6:
-//			nradices_radix0 = 2;
-			radix_prim[l++] = 3; radix_prim[l++] = 2; break;
+		/* Leading radices 5,6,9,10,11,12,13,18,20,22,24,25,26 are deliberately absent below. Their
+		carry routines have no Fermat-mod branch - the dispatch switch further down calls them
+		*without* the rn0/rn1 Fermat trig tables, unlike every radix that is supported - yet this
+		switch used to accept them, so a run that reached one dereferenced the si[] array this
+		routine passes as 0x0 and died. On stock main,
+			./Mlucas -f 18 -fft 18 -radset 4 -iters 100      (leading radix 9)
+		segfaults. Three of the thirteen (12, 20, 24) happened to hit an assert first; the rest
+		crashed. Letting the default arm below reject them turns those crashes into the clean
+		"radix N not available for Fermat-mod transform" message it already prints.
+
+		This matters more with the < 16 shift-forcing this PR adds: those routines' own
+		'if(RES_SHIFT) { WARN; return ERR_ASSERT; }' guard used to catch such runs by accident,
+		the shift being nonzero. Forcing it to zero removes that accidental protection.
+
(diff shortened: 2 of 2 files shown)


MarkRose:
Two things here.

**Rebased onto current `main`.** The branch was 207 commits behind, and its stale `makemake.sh` no longer builds the current sources — `threadpool.c` fails on `sched_setaffinity`/`CPU_ZERO` because the old script doesn't pass `-D_GNU_SOURCE`. That's a build artifact of the staleness, not a code problem, but it made the branch untestable standalone.

**Added: stop `fermat_mod_square()` dispatching leading radices that have no Fermat carry routine.** Its `switch(radix0)` accepted 5, 6, 9, 10, 11, 12, 13, 18, 20, 22, 24, 25, 26, whose routines have no Fermat-mod branch — the dispatch switch further down says so plainly, calling them *without* the `rn0`/`rn1` Fermat trig tables, unlike every supported radix. A run that reached one dereferenced the `si[]` array passed as `0x0`:

```
./Mlucas -f 18 -fft 18 -radset 4 -iters 100     # leading radix 9  ->  SIGSEGV
```

on stock `main`. Three of the thirteen (12, 20, 24) happened to hit an assert first; the other nine crashed.

This belongs with *this* PR rather than its own, because the `< 16` shift-forcing added here is what removes the accidental protection: those routines' own `if(RES_SHIFT) { WARN; return ERR_ASSERT; }` guard used to catch such runs only because the shift was nonzero. Force it to zero and they run on into the crash.

Also fixed the `default:` arm to name the radix it actually rejected — it printed `RADIX_VEC[i]`, whose `i` is left over from an earlier loop, so a rejected `radix0` of 18 was reported as "radix 16". Only visible now that the arm is reachable in practice.

Verified: `-radset 2` and `-radset 4` now give the clean *"radix 18/9 not available for Fermat-mod transform"* instead of crashing, and F18 at 14K/16K/28K still completes every radix set (6 of 6, 5 of 5, 6 of 6). Reachability is `-radset`-only — `get_fft_radices()` offers Fermat lengths whose odd part is 1, 7, 15 or 63, so automatic selection never picks these.

One residual, unchanged by this commit: FFT lengths 8K/18K/30K remain *fatal* rather than skipping the unsupported radix set and carrying on, because that `default:` arm is an `ASSERT` rather than a soft `return ERR_ASSERT`. Same before and after — I left it alone since bailing mid-init would risk leaving the routine's statics half-set for a subsequent call.

src/Mlucas.c
```diff
@@ -1650,6 +1650,21 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+			// residue shifts - they WARN + return ERR_ASSERT on RES_SHIFT != 0 (see e.g. the guard at the
+			// top of radix8_ditN_cy_dif1.c). Fermat-mod, however, applies a random default shift whenever
+			// the user specifies none, so a leading-radix-<16 radix set (e.g. the 2K {8,8,16} set) would
+			// abort mid-carry-step even for an otherwise-valid production invocation, and even an explicit
+			// '-shift 0' does not help in the cases where the shift is (re)randomized below. So for a leading
+			// radix < 16, force shift 0 - as in the p-1 branch above, a zero shift simply routes the initial
+			// seed straight into a[0] - and if a nonzero shift was requested (explicitly via -shift, or
+			// implicitly via the Fermat-mod random default), emit an informational note rather than aborting:
+			if(parse_cmd_args_get_shift_value() != 0ull) {	// != 0 catches both an explicit nonzero -shift and -1 = "-shift unspecified" (=> would-be random)
+				snprintf(cbuf,STR_MAX_LEN*2, "INFO: Leading FFT radix %u is < 16, which does not support shifted residues; setting residue shift = 0.\n", RADIX_VEC[0]);
```
tdulcet: We should use `sizeof(cbuf)` here, as we did in the other PRs.

src/fermat_mod_square.c
```diff
@@ -410,33 +410,29 @@ int fermat_mod_square(double a[], int arr_scratch[], int n, int ilo, int ihi, ui
 			nradices_radix0 = 2;
 			radix_prim[l++] = 2; radix_prim[l++] = 2; break;
 		*/
-		case 5:
-//			nradices_radix0 = 1;
-			radix_prim[l++] = 5; break;
-		case 6:
-//			nradices_radix0 = 2;
-			radix_prim[l++] = 3; radix_prim[l++] = 2; break;
```
tdulcet, line 418: For consistency, maybe block comment out these radix values instead of removing them, as some of the other ones already are. That may make it easier to see which radix values are currently unsupported, as well as potenchally reenable them in the future.

src/fermat_mod_square.c
```diff
@@ -410,33 +410,29 @@ int fermat_mod_square(double a[], int arr_scratch[], int n, int ilo, int ihi, ui
...
+		crashed. Letting the default arm below reject them turns those crashes into the clean
+		"radix N not available for Fermat-mod transform" message it already prints.
+
+		This matters more with the < 16 shift-forcing this PR adds: those routines' own
+		'if(RES_SHIFT) { WARN; return ERR_ASSERT; }' guard used to catch such runs by accident,
+		the shift being nonzero. Forcing it to zero removes that accidental protection.
+
+		Automatic radix selection never picks these for Fermat-mod - get_fft_radices() only offers
+		lengths whose odd part is 1, 7, 15 or 63 - so this is reachable only via an explicit
+		-radset. */
```
tdulcet: This comment does not need to describe the bug in detail after it has already been fixed, so maybe reduce the verbosity.

tdulcet (changes requested):
Thanks for working on this. I just have a few nits.

MarkRose:
All three done.

**`sizeof(cbuf)`** — and the literal was wrong as well as brittle: it said `STR_MAX_LEN*2` where `cbuf` is `STR_MAX_LEN*3`.

**Unsupported radices commented back in** rather than deleted — 5, 6, 9–13 and 18, 20, 22, 24, 25, 26, each in its original position so the switch still reads in numeric order, matching the existing commented-out block for cases 2–4 just above it. Agreed that this is better: it keeps visible which radices are unsupported, and re-enabling one becomes a matter of uncommenting it.

**Comment trimmed** to six lines — what the default arm does now, and the fact these are reachable only via an explicit `-radset` — instead of the full account of how the crash was found.

Re-verified on nosimd after the changes:

- the #119 repro (F13, 2K `{8,8,16}`, leading radix 8) prints the INFO note and completes with `Res64: C8FC67EA3A1AC788`
- identical residue with no `-shift`, `-shift 0` and `-shift 5` — still shift-invariant
- `-f 18 -fft 18 -radset 4` (leading radix 9, now commented out) gives the clean `not available for Fermat-mod transform` message instead of segfaulting
- Mersenne `-fftlen 192` unchanged at `Res64: 71E61322CCFB396C`

Also rebased onto current `main`; the branch was 94 commits behind.

tdulcet (approved):
Thanks.

PR #147, opened 2026-07-23
Locate completed assignment in worktodo by content, not position (fixes #121)

### LINKED ISSUE #121: Crash after inserting assignment at top of work file
This was reported by Ken on [the forum](https://www.mersenneforum.org/node/22985?p=1114655#post1114655):
> After ~7 months (208.+ days elapsed, October 24 early to May 20 late) of running Mlucas 21.0.2 on Xeon Phi 7210, Ubuntu 22.04.2 LTS, on 3321928483 P-1, including TWO 5+hour GCD NF at the end of stage 2, ~4996. total hours run time logged, instead of a results file and record, I got this at the console:
> `obdp1.sh: line 1: 8373 Aborted (core dumped) numactl --preferred=1 ./Mlucas -shift 0 -cpu 0:63,64:127 -maxalloc 99 &>> obdp1.log`

> This might have occurred because I inserted other work into worktodo.txt in front of the OBD line while the run was in progress.

and:
> I think the rerun succeeding confirms the issue is related to the work in progress no longer being the first worktodo line when it completes.

Users are strongly encouraged not to modify the work file while their GIMPS software is running, but doing so should also not result in a crash.

MarkRose:
## Problem (#121)

Two sites assumed the just-completed assignment is the **first eligible line** of `worktodo.txt`: the `GET_NEXT` deletion loop, and — the crash the reporter actually hit — `generate_JSON_report()`, which ASSERTs the first line contains the completed exponent *while writing the result*. Insert other work above a months-long in-progress run and, at completion, Mlucas dumps core **without writing the result** (reproduced: SIGABRT, no results.txt).

## Fix

Both sites now scan the workfile for the line matching the completed exponent (ESTRING, or the binary exponent for Fermat KBNC lines). The deletion loop copies every other line — comments, blanks, and unrelated/user-inserted assignments — through verbatim and in order, deleting/modifying only the matching line. If no matching line exists (user edited it away mid-run), both sites warn and continue: the result is always written; a mangled workfile can never cost a run.

## Testing (nosimd, LL M40009 with mid-run workfile edits)

| scenario | before | after |
|---|---|---|
| line inserted above running assignment | SIGABRT, **result lost** | completes; result written; correct line removed; inserted line preserved and then processed |
| assignment is first line (normal) | ok | unchanged |
| assignment's line deleted mid-run | SIGABRT | completes; result written; two warnings; remaining entries intact |

Mersenne sanity `-fftlen 192`: expected `Res64 = 71E61322CCFB396C`.

Resolves #121

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -403,6 +403,7 @@ uint32	ernstMain
 #endif
 	double tests_saved = 0.0;	// v21: make this a dfloat to allow fractional-parts
 	uint32 pm1_done = FALSE, split_curr_assignment = FALSE, s2_continuation = FALSE, s2_partial = FALSE;
+	uint32 curr_assignment_found = FALSE;	// v21: TRUE once the just-completed assignment line is located in the workfile
 	uint32 pm1_bigstep = 0, pm1_stage2_mem_multiple = 0, psmall = 0;
 /*...allocatable data arrays and associated params: */
 	static uint64 nbytes = 0, nalloc = 0, arrtmp_alloc = 0, s1p_alloc = 0;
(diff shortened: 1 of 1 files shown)


MarkRose:
Dropped the `GET_NEXT` label this branch orphans (#244).

On `main` the end-of-assignment code reads line 1 of the workfile and, if it is not an assignment line, does `goto GET_NEXT;` to skip ahead. This branch replaces that assume-line-1 logic with a scan over the whole file, which removes the only jump to the label and leaves it defined but unreferenced — `-Wunused-label`.

The label is Ernst's and predates this work, so worth being explicit that removing it is safe rather than a behaviour change: there is no remaining `goto GET_NEXT;`, and no computed goto or label-address (`&&GET_NEXT`) reference either. The fourteen `goto GET_NEXT_ASSIGNMENT;` statements elsewhere target a *different* label and are untouched — a plain grep for `goto GET_NEXT` matches those by prefix, which made this look busier than it was and briefly had me chasing the wrong thing.

Build clean with no `-Wunused-label`, `-s tt` rc=0 with 23 residues. Also rebased onto current `main` (it was 159 behind).

PR #148, opened 2026-07-23
radix320: work around GCC 11 AVX-512 auto-vectorizer miscompile (SIGSEGV)

MarkRose:
## Problem

The `Mlucas Linux (ubuntu-22.04, gcc)` CI job segfaults in the `-s m` self-test at **M12628613 / 640K FFT / radices {320,32,32}**, multithreaded, on any AVX-512-capable runner. ubuntu-22.04 ships gcc 11.4; the same job's clang build and the gcc-13/15 jobs (ubuntu-24.04/26.04) pass.

## Two distinct failures, each needing its own workaround

Root-caused with an instrumented build under Intel SDE (ubuntu:22.04 container, gcc 11.4.0, exact CI flags: `-mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma -flto=auto`).

**1. Auto-vectorizer, `-O3` only.** The compiler-vectorized code in `cy320_process_chunk()` dereferences a NULL-based address — SIGSEGV on a compiler-generated `vmovaps zmm0,[rax]` with `rax = 0x7d80`, a struct offset whose base pointer has been lost. Addressed by the file-level `#pragma GCC optimize ("no-tree-vectorize","no-tree-slp-vectorize")`, guarded to GCC 11 + AVX-512.

**2. Pointer corruption, every level from `-O1` up.** `cy320_process_chunk()`'s local `double *a` (extracted from `thread_arg->arrdat`) is confirmed **valid** at the point of extraction — a real heap pointer, identical across all 4 worker threads. It is then **silently corrupted to a small integer** (losing its base address, keeping only an index/offset component) later in the same function, before being passed as `a+jt` into `SSE2_RADIX_64_DIT()`.

The second is not fixed by the first:

| build (all `-mavx512f` et al, gcc 11.4.0) | result |
|---|---|
| `-O3`, no workaround | SIGSEGV (matches CI) |
| `-O3` + `-fno-tree-vectorize -fno-tree-slp-vectorize` | **still crashes** — different addresses each build, same defect |
| `-O3` + those + `-fira-algorithm=priority` (#118's workaround) | **still crashes** |
| `-O1` | **still crashes** |
| `-O0` | correct (`Res64 F951B697F5C5032A`, matching a native AVX2 build exactly) |

## Fix

Disable optimization for just the one affected function via `__attribute__((optimize("O0")))` rather than the whole file, keeping the cost scoped to that function. The file-level pragma stays: it addresses failure 1, which the attribute does not.

## Caveat on the attribution

This is presented as a GCC 11 codegen bug because `-O0` is clean and GCC 12+/clang are unaffected. **That reasoning is weaker than it looks, and I have not re-tested it against the alternative.** Two "compiler bug" verdicts in this tree turned out to be undeclared inline-asm clobbers — UB that is latent until the allocator happens to pick the undeclared register, which makes it look exactly like an optimization-level-dependent, compiler-version-dependent miscompile. See #68, which fixes that class tree-wide and also touches `radix320_main_carry_loop.h` (for a separate unsigned pointer-wrap). Before this workaround is treated as permanent it is worth checking whether the asm reached from `cy320_process_chunk()` declares everything it writes.

## Testing

With the fix, under gcc-11.4.0 + Intel SDE (`-skx`), the CI's own repro (M12628613 @ 640K FFT, radices {320,32,32}, 4 threads): 100 iterations complete cleanly over 3 repeated runs, deterministic, `Res64 = F951B697F5C5032A` every time, matching a native AVX2 build exactly.

Spot-checked radix1024 (`M19800083 @ 1024K`, `-radset 0`) for the same class of bug since `cy1024_process_chunk()` has a similar structure — no early crash. A full sweep of the other large-radix `*_process_chunk()` functions (64/192/960/4032) has not been done.

### DIFF
--- a/src/radix320_ditN_cy_dif1.c
+++ b/src/radix320_ditN_cy_dif1.c
@@ -23,6 +23,23 @@
 #include "Mlucas.h"
 #include "radix64.h"
 
+/* Work around a GCC 11 auto-vectorizer bug: in AVX-512 builds at -O3 (where GCC 11 enables the
+tree vectorizers), GCC 11.x miscompiles this translation unit - the compiler-vectorized code in
+cy320_process_chunk() dereferences a NULL-based address (observed on the ubuntu-22.04/gcc-11.4 CI
+runners and reproduced under Intel SDE with the same toolchain: SIGSEGV on a compiler-generated
+`vmovaps zmm0,[rax]` with rax = 0x7d80, i.e. a struct offset with its base pointer lost, at the
+very first multithreaded carry step of a leading-radix-320 run, e.g. M12628613 @ 640K FFT with
+radices {320,32,32}). The same build at -O2 - or with the tree vectorizers disabled at -O3, as
+done here - produces correct code (verified: 100-iteration Res64 matches the good-compiler
+reference). GCC 12+ (13/15/16 verified) and Clang are unaffected, as is the AVX2 build; note this
+is a different GCC 11 bug than the register-allocator one worked around in twopmodq96.c - forcing
+-fira-algorithm=priority does NOT fix this one. Only compiler-autovectorized glue code is
+affected; the hand-written AVX-512 asm macros which do the heavy lifting are untouched, so the
+performance impact is negligible. */
+#if defined(USE_AVX512) && defined(__GNUC__) && !defined(__clang__) && (__GNUC__ == 11)
+	#pragma GCC optimize ("no-tree-vectorize","no-tree-slp-vectorize")
+#endif
+
 #define RADIX 320	// Use #define rather than const int to ensure it's really a compile-time const in the C sense
 #define ODD_RADIX 5	// ODD_RADIX = [radix >> trailz(radix)]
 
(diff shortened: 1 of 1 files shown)


MarkRose:
Closing this — **the GCC 11 codegen bug it works around does not exist.** The crash is an undeclared inline-asm clobber, and **#203 already fixes it.**

## The cause

`SSE2_RADIX_05_DFT_0TWIDDLE` in `src/sse2_macro_gcc64.h` writes `%zmm8` and `%zmm9` in its AVX-512 and IMCI512 bodies, but its clobber list stops at `"xmm7"` (lines 2931 and 3022 on `418f365`). That is the *same macro and the same two registers* as the radix-240 SIGSEGV, which is what #203 fixes — radix240 and radix320 are the only two leading radices whose AVX-512 build expands this macro.

## The measurement

gcc 11.4 in an `ubuntu:22.04` container under `sde64 -skx`, all on pristine `418f365`, with **no** `optimize("O0")` and **no** pragma. Case is `M12628613 @ 640K, radices {320,32,32}`, 4 threads:

| tree | result |
|---|---|
| pristine | SIGSEGV at `0x7d80` |
| + `"xmm8","xmm9"` on **one** clobber list (only `radix320_ditN_cy_dif1.o` rebuilt) | `Res64: F951B697F5C5032A` — correct |
| fix reverted | SIGSEGV again |
| pristine + `-ffixed-xmm8 -ffixed-xmm9`, no source change | correct |
| pristine + `-ffixed-xmm12 -ffixed-xmm13` (also heavily allocated in that function, never written by the asm) | SIGSEGV |

That last pair is the discriminating control: it is those two specific registers, not generic allocation perturbation.

The disassembly closes it. `cy320_process_chunk` has exactly three 64-bit reads of `xmm8`: `vmovdqa64 %xmm29,%xmm8` parks a pointer, `vfmadd132pd %zmm31,%zmm5,%zmm8` — the macro body, verbatim — destroys it, and `vmovq %xmm8,%r8` then hands it to `SSE2_RADIX_64_DIF` as a base address. Lane 0 of that FMA is `0.0`, so the pointer becomes NULL and only the callee's offset survives. That is exactly the "corrupted to a small integer, base lost, offset kept" symptom this PR describes, and why the faulting address is a clean, repeatable `0x7d80`.

Every observation this PR read as evidence of a compiler bug is explained by the missing clobber — including "`-O1` also fails" and "disabling the vectorizers doesn't help", both of which are what latent asm UB looks like when the allocator's choices shift.

## Two further reasons not to merge as-is

**The `optimize("O0")` is unguarded.** Unlike the pragma above it, the attribute at line 2556 has no `__GNUC__ == 11` and no `USE_AVX512` condition. It forces `-O0` on the whole per-thread carry worker for **every** multithreaded build — gcc 12/13/16, clang, AVX2, SSE2, scalar — across all ten leading-radix-320 FFT lengths.

**#68 does not cover this**, checked both ways: it never touches this macro, and `418f365` plus its full diff still SIGSEGVs at the same address. #68 and #203 are complementary — with both applied the radix320 TU audits clean for vector *and* mask gaps, and all four 640K radix sets give `F951B697F5C5032A`.

## For the record

This is the **third** "compiler bug" in this tree to dissolve into an undeclared clobber, after #181 (radix1024, `zmm18/19` + `k1`-`k4`) and #162 (radix32 wrapper, `r11/r12/r13`), all with the identical signature: `-O0` clean, one compiler version affected, others not. That hypothesis deserves to be the default first check.

The branch is kept, so this is trivially reopenable if I have this wrong.

PR #149, opened 2026-07-23
imul_macro: fix generic MUL64x32 truncating partial products to 32 bits (fixes macOS-ARM ASan CI)

MarkRose:
## Problem

The `AddressSanitizer macOS` CI job (Apple Silicon) aborts in `test_fac`'s 96-bit factor section: `twopmodq96_q4(2147483647, k=68745 x4)` returns `0` instead of `15` → assert at factor_test.h:885.

## Root cause

The **generic** `MUL64x32` (imul_macro0.h) — used on ARM and any target that falls through to the generic-C wide-multiply branch — computes its partial products as

```c
_lo = ((uint32)((_x) & 0xffffffff)) * (_y);   /* a*c */
_t  = ((uint32)((_x) >> 32))       * (_y);    /* b*c */
```

When a caller passes a **32-bit** `_y` — which `SQR_LOHI96` does (`uint32 __tt`), as do the many `MUL64x32(x.d0, y.d1, …)` sites where `dN` is a `uint32` field — both operands are 32-bit, so the product is evaluated in **32-bit arithmetic and truncated** before the widening assignment to the `uint64` result. The high half of the 96-bit product is lost, corrupting every 96-bit squaring: `SQR_LOHI96 → SQR_LOHI96_q4 → twopmodq96_q4` returns a wrong residue.

`x86_64`/PPC never hit this: their `MUL64x32` is `MUL_LOHI64(_x, (uint64)_y, …)`, a full 64×64 multiply with `_y` widened to 64 bits.

## Fix

Cast one factor of each partial product to `uint64` so the multiply is always 64-bit regardless of `_y`'s type (preserving the documented low-32-bits-of-`_y` semantics). Applied to both the macro and its `EWM_DEBUG` twin.

## Testing

Under an aarch64 build (podman `--arch arm64`, gcc-13, qemu):
- an `__int128`-reference unit test of `MUL_LOHI96`/`SQR_LOHI96`/`MULL96`/`MULH96` (200k random inputs) — **`SQR_LOHI96` failed before, all pass after**;
- `twopmodq96_q4(M31, k=68745 x4)` returns **15** (was 0).

x86 is unaffected — the patched branch is `#elif`-excluded there — and its identical unit test still passes both before and after.

### DIFF
--- a/src/imul_macro0.h
+++ b/src/imul_macro0.h
@@ -1369,31 +1369,6 @@ or the with functions using them (if we declare no _-prepended variables local t
 	Even though the high output for 64x32-bit is always < 2^32,
 	assume _y and _hi here are 64-bit ints to allow flexibility for caller.
 	*/
-   #if EWM_DEBUG
-	#define  MUL64x32(_x, _y,_lo,_hi)\
-	{\
-		char s0[21],s1[21];\
-		uint64 _t,_a,_b;\
-		\
-		ASSERT(((uint64)(_y) >> 32) == 0,"MUL64x32: ((_y) >> 32) == 0");\
-		MUL_LOHI64((_x), (uint64)(_y), _a, _b);\
-		\
-		_lo = (uint64)((uint32)(_x)) * (uint32)(_y);			/* a*c */\
-		_t  = (uint64)((uint32)((_x) >> 32)) * (uint32)(_y);	/* b*c */\
-		_hi = (_t >> 32);\
-		_t <<= 32;\
-		_lo +=  _t;\
-		_hi += (_lo < _t);\
-		\
-		if(_a != _lo || _b != _hi)\
-		{\
-			printf("x = %s, y = %s\n", s0[convert_uint64_base10_char(s0,_x )], s1[convert_uint64_base10_char(s1,_y)]);\
-			printf("LO= %s, A = %s\n", s0[convert_uint64_base10_char(s0,_lo)], s1[convert_uint64_base10_char(s1,_a)]);\
-			printf("HI= %s, B = %s\n", s0[convert_uint64_base10_char(s0,_hi)], s1[convert_uint64_base10_char(s0,_b)]);\
-			ASSERT(0,"0");\
-		}\
-	}
-   #else
 	#define  MUL64x32(_x, _y,_lo,_hi)\
 	{\
 		uint64 _t;\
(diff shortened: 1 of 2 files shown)


src/imul_macro0.h
```diff
@@ -1397,9 +1397,13 @@ or the with functions using them (if we declare no _-prepended variables local t
...
-		\
-		_lo = ((uint32)((_x) & 0x00000000ffffffff)) * (_y);	/* a*c */\
-		_t  = ((uint32)((_x) >> 32)) * (_y);					/* b*c */\
+		/* v21: cast one factor of each partial product to uint64 so the multiply is done in 64-bit\
+		even when _y is a 32-bit type. The old (uint32)*(_y) form truncated a*c and b*c to 32 bits\
+		whenever the caller passed a uint32 _y (e.g. SQR_LOHI96's `uint32 __tt`), producing wrong\
+		96-bit squares - and hence wrong twopmodq96 results - on every platform that uses this generic\
+		macro (ARM, generic C). x86_64/PPC never hit it because their MUL64x32 is MUL_LOHI64(_x,(uint64)_y,...). */\
+		_lo = (uint64)((uint32)(_x)) * (uint32)(_y);			/* a*c */\
+		_t  = (uint64)((uint32)((_x) >> 32)) * (uint32)(_y);	/* b*c */\
```
ldesnogu, line 1384: The fix looks good to me.

Two comments:
1. it might be possible to merge the debug and non-debug variants to avoid code duplication
2. the C implementation in imul_macro.c seems buggy too (y should be cast)

MarkRose:
Thanks @ldesnogu! Both addressed in 5459faa:

1. **Merged the two `MUL64x32` variants.** The `EWM_DEBUG` variant only existed to recompute a `MUL_LOHI64` reference and catch exactly this truncation bug, so with the arithmetic corrected it's redundant — I kept its one still-useful check (`y < 2^32`) as a `DBG_ASSERT` and dropped the duplicated body. (Directives can't be nested inside a macro definition, so `DBG_ASSERT` is the clean way to keep the debug-only assert in a single definition.)

2. **Good catch on `imul_macro.c` — fixed.** The subroutine form did `(uint32)(x) * y` with the full 64-bit `y`, so it both disagreed with the macro and returned a wrong result if `y` carried bits above 32. Now masks `y` to its low 32 bits and casts one factor to `uint64`, matching the macro, plus the `y < 2^32` `DBG_ASSERT`.

Re-verified the int128-reference 96-bit macro unit test on x86 and aarch64 (qemu) with `EWM_DEBUG` off and on, and added a direct unit test of the subroutine form — passes on both arches and correctly uses only `y`'s low 32 bits.

ldesnogu (approved):
Thanks, that looks better now!

PR #150, opened 2026-07-23
CI: skip jobs that a change cannot affect (path-based gating)

MarkRose:
## What

Adds a small `changes` gate job (`dorny/paths-filter`) that classifies the diff, and gates every build/test/lint job on the relevant category so a change only runs the jobs it can actually affect. A documentation-only PR currently spends ~70 matrix runs building and testing every platform for nothing.

Categories:

| filter | globs | gates |
|---|---|---|
| `code` | `src/**`, `makemake.sh`, `config-fermat.sh`, `.github/workflows/**` | all build / test / sanitizer / analyzer jobs |
| `shell` | `**/*.sh` | ShellCheck |

A docs-only change (`*.md`, `docs/`, `LICENSE`, `help.txt`, …) matches neither → those jobs skip. Editing `makemake.sh`, `config-fermat.sh`, or the workflow itself is `code`, forcing the full matrix (they change how everything builds).

## Safety

- **Gating is `pull_request`-only.** Pushes to `main` and the monthly `schedule` always run the full matrix (`github.event_name != 'pull_request' || needs.changes.outputs.<cat> == 'true'`), so nothing is ever skipped on the authoritative branch or the cron sweep.
- Skipped jobs report as *skipped* (neutral), not failed.
- **No per-arch / per-ISA gating.** I checked whether an "AVX-512-only change doesn't affect ARM" rule was feasible: it isn't safely — the SIMD asm headers (`carry_gcc64.h` etc.) carry `USE_ARM_V8_SIMD` branches and are compiled into the ARM builds, and no source file is ARM-only, so there's no file set that can gate an architecture out without risking an untested regression. Omitted deliberately.

TSan-macOS is left exactly as-is (still `if: false` on main).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,13 @@ on:
     branches:
       - main
   pull_request:
+    # Skip the matrix for changes that cannot affect any build or test. pull_request only:
+    # pushes to main and the monthly schedule always run everything. A denylist rather than
+    # an allowlist, so an unrecognised new file type still runs CI rather than silently not.
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'LICENSE'
   schedule:
     - cron:  '0 0 1 * *'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
       - '**.md'
       - 'docs/**'
       - 'LICENSE'
+      - 'help.txt'
   schedule:
     - cron:  '0 0 1 * *'
 


tdulcet (changes requested):
Thanks, this is a good idea, but I believe we could do it without needing to add the complexity of a new job.

When any code is changed, effectively the entire workflow would need to be run and [GitHub already supports specifying](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore) for PRs which modified files trigger the workflow. We should use that built-in functionality here.

The ShellCheck job is already the fastest (it takes just a few seconds), so separating that would not save us any runtime.

MarkRose:
Agreed on all three points — rewritten to use the built-in filter. The `changes` job and its fourteen `needs:` / `if:` pairs are gone, replaced by seven lines on the trigger:

```yaml
  pull_request:
    paths-ignore:
      - '**.md'
      - 'docs/**'
      - 'LICENSE'
```

Net effect on the PR: **+7 lines instead of +49**, and it no longer touches the parts of `ci.yml` that #103, #155 and #216 also edit — so it drops out of that conflict cluster.

Also dropped the separate `shell` category and its ShellCheck gating, per your point that the job takes seconds anyway.

Two deliberate choices worth stating:

- **`paths-ignore` rather than `paths`.** A denylist keeps running CI for anything unrecognised; an allowlist would fail open to *skipping* the first time someone adds a file type nobody thought to list.
- **`pull_request` only.** Pushes to `main` and the monthly schedule still run everything, so nothing is ever skipped on the authoritative branch or the cron sweep.

GitHub skips a run only when *every* changed file matches, so a PR touching both docs and source still runs the full matrix. I checked the cases that matter:

| change | result |
| --- | --- |
| docs only (`README.md`, `docs/**`) | **skip** |
| source, workflow, or shell script | run |
| `help.txt` only | run |
| docs + one source file | run |
| an unrecognised new file type | run |

One thing I confirmed while looking at the original concern: `main` has no branch protection configured, so there are no required status checks that a skipped run could leave permanently pending as "expected — waiting for status". That was the only real hazard with workflow-level path filtering, and it does not apply here.

tdulcet (dismissed):
Thanks, this looks much better.

> `help.txt` only

Maybe add `help.txt` as well, since it is also just documentation.

MarkRose:
Added. You're right that it belongs in the same bucket — I checked before adding it, and `help.txt` is documentation in a stronger sense than I expected: it is never read by the build, never compiled in, and not consumed by any job. Its only appearance anywhere in the tree is a message string telling the user to go read it:

```
src/Mlucas.c:4930: fprintf(stderr, "Please refer to the help.txt file for the full list of command line options.\n");
```

so a change confined to it genuinely cannot affect a build or test.

Worth noting for anyone reading the rule later: `paths-ignore` skips the run only when *every* changed file matches, so a PR that touches `help.txt` alongside code still runs the full matrix. That is the case that matters here, since documenting a new flag normally arrives together with the flag.

Also rebased onto current `main` — the branch was 62 behind. Sorry that this dismisses the approval; the `help.txt` commit would have done so regardless.

tdulcet (approved):
Thanks.

PR #151, opened 2026-07-23
CI: experimental RISC-V (nosimd) build + self-test

MarkRose:
## What

A first look at whether Mlucas builds and runs on **RISC-V** — the org's `ubuntu-24.04-riscv` / `ubuntu-26.04-riscv` runners were just enabled, and as far as I know a riscv64 build has never been attempted. Mlucas has no RISC-V SIMD kernels, so this exercises the portable **scalar-double (`nosimd`)** path.

A **standalone workflow** (`.github/workflows/riscv.yml`) so it neither gates nor is gated by the main CI matrix:

- matrix: `{ubuntu-24.04-riscv, ubuntu-26.04-riscv} × {gcc, clang}`, `fail-fast: false`
- prints toolchain + target-arch macros up front (so a build/arch-detection misfire is obvious)
- `makemake.sh nosimd` (the one mode with no arch-specific asm and no `-march` flag)
- self-test: the `-s tt` (teensy) tier only. It still checks each radix set's residue against Mlucas's built-in reference table, so it is a real known-answer correctness check, just at tiny FFT lengths. (The second commit dropped the heavier fixed `-fftlen 192` × 100-iteration run — `Res64 = 71E61322CCFB396C` — because the RISC-V runners may be slow; an earlier revision of this description still advertised it.)
- uploads `run.log` for first-attempt debugging. The artifact block also lists `obj_nosimd/build.log`, which nothing currently writes — the build step needs a `tee`, or that path should come out.

## Expectations

This is exploratory — it may well fail on the first pass (untested libm/`long double` behavior, `-flto` on the riscv toolchains, missing packages, etc.), and that's the point: the job is designed to surface *where* it breaks rather than to be green immediately. If it passes, RISC-V scalar builds are viable and we can consider promoting it into the main matrix later.

### DIFF
--- a/.github/workflows/riscv.yml
+++ b/.github/workflows/riscv.yml
@@ -51,11 +51,11 @@ jobs:
       run: |
         set -x
         cd obj_nosimd
-        # Canonical known-answer check: FFT length 192 must give this exact Res64 on any correct build.
-        ./Mlucas -fftlen 192 -iters 100 -radset 0 | tee run.log
-        grep -q 'Res64: 71E61322CCFB396C' run.log
-        # And the built-in tiny self-test tier as a broader smoke test:
-        ./Mlucas -s tt
+        # RISC-V runners may be slow, so use only the smallest 'tt' (teensy) self-test tier. It still
+        # validates each radix set's residue against Mlucas's built-in reference table, so it is a real
+        # known-answer correctness check, not just a compile/smoke test - it just does it at tiny FFT
+        # lengths rather than the heavier fixed fftlen-192 x 100-iteration run.
+        ./Mlucas -s tt | tee run.log
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/.github/workflows/riscv.yml
+++ b/.github/workflows/riscv.yml
@@ -1,0 +1,65 @@
+name: RISC-V (experimental)
+
+# Exploratory: RISC-V has never been built/tested before, and the org's riscv64 runners were only
+# just enabled. Mlucas has no RISC-V SIMD kernels, so this builds the portable scalar-double
+# ("nosimd") mode and runs the standard self-test to see whether the C code base works at all on
+# riscv64. Kept in its own workflow so it neither gates nor is gated by the main CI matrix.
+on:
+  push:
+    branches:
+      - main
+      - ci-riscv-experimental
+  pull_request:
+    paths:
+      - '.github/workflows/riscv.yml'
+      - 'src/**'
+      - 'makemake.sh'
+  workflow_dispatch:
+
+jobs:
+  RISC-V:
+    name: RISC-V nosimd
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


ldesnogu:
A generic comment: RISC-V extension space is messy.

Though I have no experience with R-V, I recommend using the RVA23 profile: https://docs.riscv.org/reference/rva23/v1.0/rva23-profiles.html

It might limit applicability only to newer CPUs (end of 2025), but supporting older variants would be pointless IMHO as this could mean entering a mess of variants that would be useless for any work.

QEMU has support for RVA23.

MarkRose:
These runners are provided by [Scaleway](https://www.scaleway.com/) and use their Elastic Metal RV1 servers with [T-HEAD C910 CPU](https://chipsandcheese.com/p/alibabat-heads-xuantie-c910)s with RV64GC cores. It's an experiment. We probably won't merge this given how slow it's going.

I've got a RISC-V build running a tiny self test locally through QEMU currently.

ldesnogu:
> These runners are provided by [Scaleway](https://www.scaleway.com/) and use their Elastic Metal RV1 servers with [T-HEAD C910 CPU](https://chipsandcheese.com/p/alibabat-heads-xuantie-c910)s with RV64GC cores. It's an experiment. We probably won't merge this given how slow it's going.

This isn't RVA23 alas. My point is mainly that if we start supporting R-V, we should be careful not to support too old standards, or any proprietary extension, or we risk to have to create some other support burden for an already obsolete implementation.

> I've got a RISC-V build running a tiny self test locally through QEMU currently.

Out of curiosity how much slower than C910 is it?

MarkRose:
Both points taken.

## On the baseline

You are right, and it is the more important of the two questions. RV64GC on a C910 is not what we would want to be committing to, and picking up a dependency on an old baseline — or worse a vendor extension — buys a support burden for hardware that is already behind. This should not turn into "Mlucas supports RISC-V" on the strength of whatever runner happened to be available.

That is part of why this is still a draft. I am going to try to get access to an RVA23 machine; testing against that rather than against a C910 seems like the right basis for deciding whether to support RISC-V at all, and if the answer is yes then RVA23 is the floor to state rather than whatever builds today.

## On QEMU

I can give you half an answer honestly, and I would rather flag the missing half than paper over it.

Same source, same `nosimd` build mode, same default thread count, same 44 residues out of `-s tt`, on the same host (an i9-10885H):

| | `-s tt` |
| --- | --- |
| native x86-64 | 1 s |
| `qemu-riscv64` (user-mode, binfmt) | 21 s |

So roughly **20x slower than native x86-64**. For scale, building Mlucas `nosimd` under the same emulation took 285 s.

What I cannot give you is the comparison you actually asked for, against the C910. All three runs of this workflow were cancelled or failed and their logs have since expired, so there is no completed self-test from the real hardware to measure against — which is itself a fair summary of "how slow it's going". If the RVA23 access comes through I will have a real number to put beside the QEMU one, and at that point the interesting comparison is native RVA23 against native x86-64, not either against the emulator.

PR #152, opened 2026-07-23
Fix every broken leading FFT radix in scalar (nosimd) builds: shift-unaware carries, odd-radix0 dispatch, and three routine-local bugs

### LINKED ISSUE #153: Fermat-mod with leading radix 15 computes wrong residues (right-angle-path defect)
Split out from #152, which fixed every *Mersenne*-reachable leading radix in scalar (nosimd) builds and two of the three radix-15/30 defects. One documented remainder is Fermat-specific:

**Symptom:** A Pepin test of F18 forced onto 15K with a leading-radix-15 set produces `Res64 = A7645AE04D6FECDF`, disagreeing with the 4-way cross-check reference `7C6B681485EB86DB` (obtained independently at 14K r0/r1 and 16K r0/r1, and — post-#152 — at 15K with leading radix 30 and 60). Both lead-15 radsets (`{15,32,16}`, `{15,16,32}`) agree with *each other*, i.e. the computation is deterministic and self-consistent but semantically wrong.

Repro (nosimd build, after #152):
```
./Mlucas -f 18 -fftlen 15 -radset 2 -iters 100 -shift 0   # A7645AE04D6FECDF (wrong)
./Mlucas -f 18 -fftlen 15 -radset 1 -iters 100 -shift 0   # 7C6B681485EB86DB (lead-30, correct)
```

**What is already ruled out** (diagnosis from the #152 work):
- The fused `RADIX_15_DIT`/`DIF` pair is an exact DFT pair (verified via an impulse/phase-matrix harness), and its natural-order argument convention matches the mid-pass slot convention (verified against the empirically-extracted slot→bin map of the working radix30 routine).
- The Mersenne side of the same routine is fully fixed by #152 (`13e5b97`): the wraparound mini-pass was being skipped due to a missing `jstart`/`khi` reset. After that fix the *Fermat* mini-pass also demonstrably runs, with exact-integer data (Pepin iterates 9, 81, … observed at the carry step) — yet the final residue is still wrong. So the residual defect is in the right-angle bookkeeping itself, not the transforms and not roundoff.
- One structural suspect: radix15's Fermat init sets only `bjmodn0 = n`, whereas the (now-working) radix30 routine seeds *two* half-chain entries (`bjmodn[0] = n; bjmodn[15] = n;`) for its right-angle "double helix" carry scheme; how that scheme should specialize to an odd radix (15 chains, no natural halving) is exactly the open question.

**Severity:** low — this configuration is off-table (the Fermat self-test tables and any timing-based `fermat.cfg` selection use the lead-30/60 radsets at 15·2^k lengths, both fixed by #152), so no production Pepin run should land on it. But it is a silent-wrong-result path, so it should either be fixed or explicitly rejected.

### LINKED ISSUE #154: SIMD builds crash when the radix tables select a scalar-only leading radix (e.g. 9, 30)
Found during the #152 investigation. The small/odd leading-radix carry routines (radix5–15, 30, …) are scalar-only — they have no SSE2/AVX code paths — but SIMD builds still dispatch them whenever the user (or a radix-set iteration) selects an FFT decomposition with such a leading radix. The result is a hard crash rather than a clean rejection:

```
$ ./makemake.sh avx2 && cd obj_avx2
$ ./Mlucas -m 88019  -fftlen 9  -radset 2 -iters 100 -shift 0   # {9,32,16}
timeout: the monitored command dumped core
$ ./Mlucas -m 130873 -fftlen 15 -radset 1 -iters 100 -shift 0   # {30,16,16}
timeout: the monitored command dumped core
```

Control cases at the same FFT lengths with SIMD-capable leading radices (36, 60, …) produce correct reference residues in the same build, so this is specifically the scalar-only-carry-routine + SIMD-data-layout combination (the SIMD main arrays are permuted/padded differently than the scalar carry code assumes, and the routines have no guard).

Note these crashes were observed *before* the #152 fixes and are independent of them — #152 fixed what the scalar routines compute in scalar builds; this issue is about SIMD builds reaching those routines at all.

**Suggested fix direction:** either gate the affected radsets out of `get_fft_radices`/self-test iteration for SIMD builds, or have the scalar-only carry routines detect a SIMD build and return the same clean `ERR_ASSERT`-with-WARN used elsewhere (cf. the radix<16 shifted-residue guard), so self-tests skip them gracefully instead of dumping core.

**Severity:** low-medium — production runs choose radix sets from timing files that will never prefer these decompositions, but a self-test sweep or an explicit `-radset` invocation in a SIMD build crashes the process.

MarkRose:
## Summary

While validating the first-ever RISC-V (nosimd) build, `-s tt` flagged wrong residues for a few radix sets. Pulling that thread uncovered — and this PR fixes — **five distinct long-standing bugs** in the scalar-double code paths, several dating back a decade or more. After this PR, **every Mersenne-reachable leading FFT radix produces correct reference residues in nosimd builds**, verified on x86-64, arm64 and riscv64 (the latter two via qemu).

A per-(fftlen, radix-set) sweep of the full self-test tables plus a `grep -c RES_SHIFT` survey of all 58 carry routines drove the enumeration; cross-radix-set / cross-fftlen residue agreement (Res64 is invariant to both) served as the correctness oracle throughout.

## The five bugs, in commit order

1. **`869a026` + `38d1f60` — radices 18/22/26: shifted residues silently ignored.** These scalar carry routines predate the v20 shifted-residue feature: they inject the LL subtract-2 unconditionally into word 0, ignoring `RES_SHIFT`. Since v20 applies a *random default shift*, every LL run landing on a leading-radix-18/22/26 decomposition computed a wrong (deterministic, low-roundoff) residue — same defect class as #119, but with no guard, so it corrupted silently. Fixed by mirroring the standard `target_idx`/`target_cy` machinery, then (2nd commit) reworking the carry-loop dataflow from `aj1pNN` register temporaries onto `a[]` to match every sibling radix; aliasing-prone stages snapshot into dead `u`-temps first. 18/18 reference-residue KAT on all three architectures.

2. **`6b9edc9` — radices 30/63: same shift-unaware class.** Of the 12 remaining `RES_SHIFT`-free routines, only these two are reachable (9 are undispatched stubs; 512 never leads). Same target-injection treatment; zero-shift behavior preserved bit-for-bit.

3. **`c99b416` — radix30: `MOD_ADD32` fed a negative addend.** Its `bjmodn[]` chain init passed `bjmodnini - n` (the old explicit-renormalize idiom's constant) to `MOD_ADD32`, whose unsigned-compare reduction requires inputs in `[0,n)`. Every `bjmodn[l]`, `l ≥ 2` came out congruent-but-negative, silently corrupting the carry macros' big/smallword flags for the first words of every sub-chain ≥ 2. Broke lead-30 in **both** Mersenne and Fermat mod; both now match references. (Found via an iteration-divergence probe: corruption began exactly when the residue first reached carry sub-chain 2, after a standalone impulse-matrix harness had proven the DIT/DIF math itself exact.)

4. **`922062f` — odd leading radices (5, 7, 9, 11, 13, 15, 63): final work chunk never dispatched.** The bug behind the Apr-2014 *"mers-mod barfs immediately"* note in `radix63_main_carry_loop.h`: `nchunks = radix0>>1` undercounts the work units for odd radix0 (the block pairing `{0,-},{1,2},…` needs ceil), so the threadpool dispatch (`tid → ii = 2·tid`) never ran the last chunk-pass — the final pair of data blocks was never transformed or squared, every iteration, in every threaded build back through at least v20. (The serial fallback loop was correct; wrapper call/return tracing of radix0 = 9 vs the structurally identical working radix0 = 18 exposed the missing pass directly.) Fix: `(radix0+1)>>1` plus skipping the nonexistent `ws_*[radix0]` wrapper state-slot on the restored pass. `fermat_mod_square` is unaffected (`nchunks = radix0`, no pairing).

5. **`13e5b97` — radix15: wraparound mini-pass never ran.** The routine reset only `jhi` for the cleanup mini-pass where every sibling also resets `jstart`/`khi`; with `jstart` stale the mini-pass loop never executed, so the `radix_inv` pre-scaling it exists to cancel was applied un-cancelled — every iteration multiplied the first words of each block by 1/15. (Fingerprint: iteration-2's word-0 input was `(14/15)² = 0.8711` instead of 196, while iteration-1's was an exact 16.0.)

## Verification (all nosimd, cross-checked on x86-64 / arm64 / riscv64)

| Check | Result |
|---|---|
| 18-case radix-18/22/26 reference KAT, default random shift | 18/18 on all three architectures |
| Leads 5, 9, 11, 13, 15 vs cross-radset references (`-shift 0`; their <16 routines predate shift support, cf. #119/#146) | all correct — first correct results in the git era |
| Lead-30 (15K/30K Mersenne; F18 Fermat) | matches references, both modes, default shift |
| Lead-63 (1008K, M18776473), default shift **and** `-shift 0` | `CA7D81B22AE24935` ✓ — fixes the 2014 bug |
| Full `-s t` self-test tier | 0 incorrect residues, 0 crashes* |
| Even-leading radices / Fermat spot-checks | unchanged |

\* with #125's radix-12 `j1` fix applied locally to get past the unrelated pre-existing `radix12_dif_pass1` segfault (see the backtrace posted on #125); reverted before committing.

## Two further fixes in the same area

- **`d721088` — radix15: cleanly reject Fermat-mod** (fixes #153). The residual Fermat-15 wrong-residue defect is structural: the routine's pre-icycle right-angle carry scheme pairs sub-chain c with c + RADIX/2, which only exists for an even leading radix (radix30 seeds `bjmodn[0]` and `bjmodn[15]`; for odd 15 the n/2 point falls mid-chain). Odd leading radices need the icycle-based scheme radix63 uses; until that is ported, Fermat-mod on this radix set gets the same WARN + `ERR_ASSERT` clean-reject idiom as the existing shifted-residue guard, instead of a silently-wrong residue. Mersenne lead-15 is unaffected and still matches its reference.
- **`796e2db` — clean exit on out-of-range `-radset`/`-fftlen`** (fixes #154). Investigation corrected the original diagnosis: SIMD builds never dispatch the scalar-only carry routines — `get_fft_radices`' per-build tables already exclude those radix sets — so the observed "crashes" were main's `ASSERT(0,...)` aborting with a core dump on explicitly-invalid user input. Bad CLI input now gets the informative message + `exit(EXIT_FAILURE)`.

---

*Correction: the list above originally omitted **7**. Enumerating `rvec[0]` across `get_fft_radices.c` gives exactly {5, 7, 9, 11, 13, 15, 63} as the odd leading radices, so 7 is affected too. The fix is `(radix0+1)>>1` and was already generic — only the description was short.*

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -4382,7 +4382,7 @@ just below the upper limit for each FFT lengh in some subrange of the self-tests
 
 		if(iarg == 0) {
 			sprintf(cbuf  , "*** ERROR: Must specify a valid FFT length on command line before -radset argument!\n");
-			fprintf(stderr,"%s", cbuf);	ASSERT(0,cbuf);
+			fprintf(stderr,"%s", cbuf);	exit(EXIT_FAILURE);	// v21: bad user input - clean error exit, not ASSERT/abort+core
 		}
 
 		/* Make sure it's a valid radix set index for this FFT length: */
@@ -4394,7 +4394,10 @@ just below the upper limit for each FFT lengh in some subrange of the self-tests
 			else
 				sprintf(cbuf  , "ERROR: Unknown error-code value %d from get_fft_radices(), called with radix set index %d, FFT length %d K\n",i,radset, iarg);
 
-			fprintf(stderr,"%s", cbuf);	ASSERT(0,cbuf);
+			// v21: an out-of-range radix set for this build is a *user input* error (e.g. SIMD builds offer
+			// fewer radix sets per FFT length than scalar ones, since the small-leading-radix carry routines
+			// are scalar-only) - report it and exit cleanly rather than ASSERT, which aborts with a core dump:
+			fprintf(stderr,"%s", cbuf);	exit(EXIT_FAILURE);
 		}
 
 	}
--- a/src/mers_mod_square.c
+++ b/src/mers_mod_square.c
@@ -241,7 +241,12 @@ The scratch array (2nd input argument) is only needed for data table initializat
 #endif
 
 	radix0 = RADIX_VEC[0];
-	nchunks = radix0>>1;
+	nchunks = (radix0+1)>>1;	// v21 bugfix: for ODD radix0 the number of independent work units is
+			// ceil(radix0/2), not floor: the last unit processes the final pair of data blocks (e.g. blocks
+			// {5,6} of the radix0 = 9 pairing {0,-},{1,2},{3,8},{4,7},{5,6}). The old radix0>>1 value meant
+			// that in threaded builds (task tid -> ii = 2*tid) the ii = radix0-1 chunk-pass simply never ran,
+			// leaving those blocks un-transformed and un-squared every iteration - i.e. all Mersenne runs with
+			// an odd leading radix silently computed garbage. (The serial fallback loop was correct.)
 	ASSERT(TRANSFORM_TYPE == REAL_WRAPPER, "mers_mod_square: Incorrect TRANSFORM_TYPE!");
 
 /*...initialize things upon first entry */
@@ -2265,6 +2270,13 @@ void mers_process_chunk(
 	for(j = 0; j < jhi; j++)
 	{
 		l = ii + j;
+		// v21 bugfix: for ODD radix0 the last chunk-pass (ii = radix0-1) has jhi = 2 because its DIF/DIT
+		// loops must process both blocks of the final block_index pair, but the wrapper/square step covers
+		// that entire pair via the single call for slot l = radix0-1 (one block via the j1-indices, its
+		// partner via the mirrored j2-indices). There is no state-slot l = radix0 - reading ws_*[radix0]
+		// would run off the end of those arrays - so skip the second wrapper call:
+		if(l >= radix0)
+			continue;
 
 		switch(RADIX_VEC[NRADICES-1])
 		{
--- a/src/radix15_ditN_cy_dif1.c
+++ b/src/radix15_ditN_cy_dif1.c
@@ -718,6 +718,13 @@ for(outer=0; outer <= 1; outer++)
 	/*
 	For right-angle transform need *complex* elements for wraparound, so jhi needs to be twice as large
 	*/
+	// v21 bugfix: the wraparound mini-pass must restart the loop indices - every sibling radix resets
+	// jstart/jhi/khi here (cf. radix9_ditN_cy_dif1.c), but this routine only reset jhi. With jstart still
+	// holding its end-of-full-pass value (>> jhi) the mini-pass j-loop never executed, so the radix_inv
+	// pre-scaling of the first few words of each block was never cancelled by the mini-pass's re-DIF -
+	// every iteration silently multiplied those words by 1/RADIX, corrupting both Mers- and Fermat-mod.
+	jstart = 0;
+	khi = 1;
 	if(TRANSFORM_TYPE == RIGHT_ANGLE)
 	{
 		jhi =15;
--- a/src/radix15_ditN_cy_dif1.c
+++ b/src/radix15_ditN_cy_dif1.c
@@ -84,6 +84,18 @@ int radix15_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 		return(ERR_ASSERT);
 	}
 
+	// v21: The right-angle Fermat-mod carry scheme used in this routine (the older per-chain ii/bjmodn
+	// form of fermat_carry_norm_errcheck) structurally requires an EVEN leading radix: the acyclic
+	// weighting pairs word j with word j + n/2, i.e. carry sub-chain c with sub-chain c + RADIX/2, and
+	// for odd RADIX the n/2 point falls mid-chain (contrast radix30_ditN_cy_dif1.c, which seeds both
+	// bjmodn[0] and bjmodn[15]). Odd leading radices need the newer icycle-based scheme (cf.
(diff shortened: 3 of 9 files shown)


MarkRose:
Cross-referencing a merge hazard, flagged in more detail on #210: **this PR and #210 merge with zero conflicts and the resulting tree does not compile** — `src/radix63_ditN_cy_dif1.c:557: error: 'stride' undeclared`, measured on `418f365` with a real 3-way merge.

Neither is wrong alone. #210 removes the `stride` declaration (replacing `j += stride` with a logical `j += 2` walk); this PR independently adds a new use of `stride` in the same file. Whichever merges second needs that one use reconciled — either re-derive it as `(int)RE_IM_STRIDE << 1` locally, or retain the declaration.

Flagging it because git gives no warning at all, so it would otherwise surface as a red CI run after both had landed.

MarkRose:
Follow-up: the two Fermat-mod rejections this PR adds now return `ERR_RADIX0_UNAVAILABLE` rather than `ERR_ASSERT`.

`ERR_ASSERT` makes the self-test count the radix set as a failure. At F18, FFT length 15K, three of the four radix sets are radix15 or shifted radix30, so the length dropped to 1 of 4 passing - below the "half the sets must pass" threshold - and got no `fermat.cfg` entry. `config-fermat.sh` runs under `bash -e`, so it stopped there and left F19 through F29 unconfigured.

`ERR_RADIX0_UNAVAILABLE` is the code for "this build has no working implementation of this leading radix", which is what these rejections are saying, and `Mlucas.c` already discounts it from the pass fraction for exactly this reason. 15K is now 1 of 1 tried with 3 skipped, and gets its entry.

PR #155, opened 2026-07-24
Updated CI to build for K1OM and KNL

tdulcet:
* Updated the CI to build for K1OM (Knights Corner) and KNL (Knights Landing).
  * Long term, it has been discussed if we should remove support for these older platforms. This just adds some basic testing to prevent regressions, especially for major changes like #34 that affect the `USE_IMCI512` code path. Ernst added K1OM support in v21.

There are errors when attempting to build Mlucas for K1OM (looks like missing line continuation characters in the inline assembly):
```
*** There were build errors - see 'obj_k1om/build.log' for details. ***

../src/carry_gcc64.h:17995:3: error: expected identifier or '(' before string constant
   "vmovapd %%zmm3,(%%rsi){%%k6%} \n\t"
   ^
--
../src/carry_gcc64.h:17995:3: error: expected identifier or '(' before string constant
   "vmovapd %%zmm3,(%%rsi){%%k6%} \n\t"
   ^
--
../src/carry_gcc64.h:18081:2: error: expected identifier or '(' before '}' token
  }
  ^
--
../src/carry_gcc64.h:18081:2: error: expected identifier or '(' before '}' token
  }
  ^
```
The KNL builds would also need to be gated to older GCC/Clang compilers that support it. Ubuntu 26.04 and Alpine 3.24 are failing with both GCC and Clang due to the `-mavx512er` option. It looks like KNL support was removed in [GCC 15](https://gcc.gnu.org/gcc-15/changes.html) and [Clang 19](https://releases.llvm.org/19.1.0/tools/clang/docs/ReleaseNotes.html).

@MarkRose - Maybe you would be interested in fixing these issues. If so, please feel free to take over this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,11 @@ jobs:
             MODES+=( ASIMD )
         else
             MODES+=( SSE2 AVX AVX2 AVX512 )
+            if echo 'int main(void){return 0;}' | $CC -march=knl -mavx512f -mavx512er -x c -o /dev/null -; then
+                MODES+=( AVX512_KNL )
+            else
+                echo "::warning::Skipping the AVX512_KNL build mode as '$($CC --version | head -n1)' does not support -march=knl."
+            fi
         fi
         for mode in "${MODES[@]}"; do
             echo -e "\n$mode\n"
@@ -102,6 +107,11 @@ jobs:
             MODES+=( ASIMD )
         else
             MODES+=( SSE2 AVX AVX2 AVX512 )
+            if echo 'int main(void){return 0;}' | $CC -march=knl -mavx512f -mavx512er -x c -o /dev/null -; then
+                MODES+=( AVX512_KNL )
+            else
+                echo "::warning::Skipping the AVX512_KNL build mode as '$($CC --version | head -n1)' does not support -march=knl."
+            fi
         fi
         for mode in "${MODES[@]}"; do
             echo -e "\n$mode\n"
@@ -196,6 +206,11 @@ jobs:
             else
                 echo "::warning title=AVX-512 build skipped::$("${CC:-gcc}" --version | head -n1): its assembler can't handle the AVX-512 extended registers (zmm16-31/xmm16-31/k0-7) Mlucas's kernels need; skipping the AVX-512 build mode."
             fi
+            if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm31,%%zmm31,%%zmm31\n\tvmovdqa64 %%xmm16,%%xmm17\n\tkmovw %%k1,%%eax":::"zmm31","xmm16","xmm17","k1","eax");return 0;}' | $CC -march=knl -mavx512f -mavx512er -x c -o /dev/null -; then
+                MODES+=( AVX512_KNL )
+            else
+                echo "::warning::Skipping the AVX512_KNL build mode as '$($CC --version | head -n1)' does not support -march=knl."
+            fi
         fi
         for mode in "${MODES[@]}"; do
             echo -e "\n$mode\n"
@@ -210,6 +225,52 @@ jobs:
                 )
             done
         done
+
+  Linux-K1OM:
+    name: Mlucas K1OM
+
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v7
+    - name: Download
+      run: |
+          # https://web.archive.org/web/20210307201727/https://software.intel.com/content/www/us/en/develop/articles/intel-manycore-platform-software-stack-mpss.html
+          wget -nv 'https://web.archive.org/web/20200818141734id_/http://registrationcenter-download.intel.com/akdlm/irc_nas/15904/mpss-3.8.6-linux.tar'
+          tar -xvf mpss-3.8.6-linux.tar
+    - name: Script
+      run: |
+        docker run --rm -i -v "$GITHUB_WORKSPACE:/workspace" -w /workspace centos:7 bash -s <<'EOF'
+        set -e -o pipefail
+        sed -i '/^mirrorlist/d;/^#baseurl=/{s|^#||;s|/mirror|/vault|;}' /etc/yum.repos.d/CentOS*.repo
+        # yum update -y
+        yum install -y make
+        yum localinstall -y mpss-3.8.6/mpss-sdk-k1om-3.8.6-1.x86_64.rpm
+        source /opt/mpss/3.8.6/environment-setup-k1om-mpss-linux
+
+        export CC=k1om-mpss-linux-gcc
+        unset CFLAGS CPPFLAGS
+        $CC --version
+        k1om-mpss-linux-ld --version
+        $CC -dM -E - </dev/null
+
+        bash -e -o pipefail -- makemake.sh no_gmp k1om
+        (cd obj_k1om; make clean)
+        for word in '' 1word 2word 3word 4word nword; do
+            echo -e "\nMfactor $word\n"
+            bash -e -o pipefail -- makemake.sh no_gmp mfac $word k1om || true
+            (
+                cd obj_mfac${word:+_$word}_k1om
+                make clean
+            )
+        done
+        EOF
+    - uses: actions/upload-artifact@v7
+      if: always()
+      with:
+        name: linux_k1om_mlucas
+        path: |
+          ${{ github.workspace }}
+          !mpss-*
 
   ASan-Linux:
     name: AddressSanitizer Linux
--- a/makemake.sh
+++ b/makemake.sh
@@ -267,7 +267,6 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 			;;
 		avx512_knl)
 			echo "Building for avx512_knl SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
-			echo "Warning: The 'avx512_knl' option is deprecated, use 'avx512' instead."
 			ARGS+=(-DUSE_AVX512 -march=knl -mavx512f -mavx512cd -mavx512er -mfma)
 			;;
 		avx512)


MarkRose:
Two findings from getting the K1OM target actually building locally — I reproduced this job's environment under podman (CentOS 7 + MPSS 3.8.6 + `k1om-mpss-linux-gcc`), since the runners make it hard to iterate.

## 1. This job is masked, and what it's masking is real

The job carries `continue-on-error: true`, so it reports green while failing. On stock `main` the KNC build produces **compile errors, not a binary** — and it always has: the `jmp 2f` line in the `USE_IMCI512` arm of `AVX_cmplx_carry_fast_errcheck_X8` has never had its line-continuation backslash, dating to the v21 import (`c9fef37a`, 2023-12-12). Earlier releases (v14.1 → v20) don't contain the construct at all, so **KNC has never built in this repository**.

Worth knowing before the mask comes off: it will go red immediately, for reasons that predate every open PR.

Note also that the error count is misleading — `makemake.sh` runs `make -j` without `-k`, so it stops at the first failures. What looks like 2 defects is actually **5**:

1. three missing line continuations (above)
2. two unescaped `{` in asm templates — `(%%rsi){%%k5%}` needs `%{`, since GCC reads a bare `{` as a dialect alternative and the matching `%}` is escaped, so it never closes
3. a `jz 1f` with no `1:` label anywhere in the macro — gas rejects it
4. `int l,k1,k2` declared behind `#ifndef USE_AVX512` in `radix16/32_dyadic_square.c`, but `platform.h:263` makes `USE_IMCI512` *imply* `USE_AVX512`, so the IMCI-only scalar arm that uses them loses its declaration
5. `tmp,tm1,c_tmp,s_tmp` behind `#ifdef USE_PRECOMPUTED_TWIDDLES` in `radix16_wrapper_square.c` — a flag nothing defines

With those five fixed (8 insertions, 7 deletions in 4 files, all inside IMCI-only arms) the build produces `ELF 64-bit LSB executable, Intel K1OM`. I'll put that up as its own PR.

## 2. The job needs CFLAGS set, or it builds at -O0

This one bites regardless of the source fixes. `/opt/mpss/3.8.6/environment-setup-k1om-mpss-linux` exports `CFLAGS` and `CPPFLAGS`, and the generated Makefile uses `CFLAGS ?=` / `CPPFLAGS ?=`, which defer to the environment by design (`makemake.sh:512-514`). So sourcing the SDK env silently drops `-O3 -std=gnu99` and `-D_GNU_SOURCE`, and the whole tree builds at **-O0** — where `radix16_dif_dit_pass_asm.h` fails with `impossible constraint in 'asm'`, because the `"e"` (immediate) constraint on `pfetch_dist` only folds to a constant once the optimiser runs. Not KNC-specific: a host `gcc -DUSE_AVX512` build of that file fails identically at `-O0` and is clean at `-O1`/`-O3`. The tree already warns about it at `carry_gcc64.h:31`.

Separately, without `-D_GNU_SOURCE`, `threadpool.c` loses `CPU_ZERO`/`CPU_SET`/`sched_setaffinity` and the link fails.

So this job wants something like:

```sh
source /opt/mpss/3.8.6/environment-setup-k1om-mpss-linux
export CC=k1om-mpss-linux-gcc
export CFLAGS="-Wall -g -O3 -std=gnu99 -m64 --sysroot=/opt/mpss/3.8.6/sysroots/k1om-mpss-linux"
export CPPFLAGS="-D_GNU_SOURCE -m64 --sysroot=/opt/mpss/3.8.6/sysroots/k1om-mpss-linux"
```

restoring what makemake.sh would have chosen while keeping the SDK's sysroot.

## Caveat

All of this is build-only — no KNC hardware here, and SDE doesn't emulate IMCI. Nothing validates that KNC *computes* correctly, and the IMCI arm still carries its own `#warning`s to that effect ("Fermat-mod carry macros in k1om / IMCI-512 mode are only stubs for now!", plus radix-{28,36,44,52,60} disabled via a stub macro).

MarkRose (commented):
Agreed this is worth having — the `USE_IMCI512` path has no coverage at all otherwise, and #34-scale changes are exactly what would break it silently.

## The K1OM build error you documented should be fixed already

The failure in the description:

```
../src/carry_gcc64.h:17995:3: error: expected identifier or '(' before string constant
   "vmovapd %%zmm3,(%%rsi){%%k6%} \n\t"
```

is the unescaped-brace bug from #236, which merged on 2 August. That PR changed exactly that line:

```diff
-		"vmovapd	%%zmm3,(%%rsi){%%k6%}	\n\t"
+		"vmovapd	%%zmm3,(%%rsi)%{%%k6%}	\n\t"
```

`main` now has only the escaped form, and this branch's merge-base (`4a21413a`) already includes it — so the blocker predates the fix and a re-run should get further. Worth confirming, since it determines whether the K1OM job can be unmasked.

## Overlap with #239

Both PRs add KNL CI and both touch the KNL flag handling in `makemake.sh`. I merged each onto current `main` and then the other on top: they conflict in `.github/workflows/ci.yml` and `makemake.sh`. Not a hard blocker — whichever lands second needs a small rebase in the KNL job — but worth knowing they're not independent.

The main functional difference: #239 makes the three KNL-specific flags (`-march=knl`, `-mavx512er`, `-mavx512pf`) degrade individually via `try_flag`, warning per dropped flag, because GCC 14 deprecated and GCC 15 removed them. Measured locally: GCC 16 rejects all three, Clang 22 accepts `-march=knl` and rejects the ER/PF pair. This PR's probe tests `-march=knl -mavx512f -mavx512er` as a single unit, so a toolchain that kept `-march=knl` but dropped ER would skip the mode entirely rather than build it degraded. Might be worth combining the two approaches.

## Minor

- 48 commits behind `main`, so it'll want a rebase regardless.
- The `AVX512_KNL` probe block appears three times with slightly different asm bodies (the third also exercises `zmm31`/`xmm16`/`kmovw`). I assume that's one per job rather than accidental duplication — worth a look that they're all the intended variant.

tdulcet:
I will rebase this PR as requested.

> This PR's probe tests `-march=knl -mavx512f -mavx512er` as a single unit, so a toolchain that kept `-march=knl` but dropped ER would skip the mode entirely rather than build it degraded. Might be worth combining the two approaches.

In the CI, it may not be worth testing the separate build without AVX512 ER support. Both GCC and Clang dropped both KNL and AVX512 ER support at the same time.

> * The `AVX512_KNL` probe block appears three times with slightly different asm bodies (the third also exercises `zmm31`/`xmm16`/`kmovw`). I assume that's one per job rather than accidental duplication — worth a look that they're all the intended variant.

Yes, that was intended. The third one duplicates the existing AVX512 extended register check, which is needed to prevent build errors in older compilers that support KNL but not the extended registers.

.github/workflows/ci.yml
```diff
@@ -211,6 +226,52 @@ jobs:
...
+        unset CFLAGS CPPFLAGS
+        $CC --version
+        k1om-mpss-linux-ld --version
+        $CC -dM -E - </dev/null
+
+        bash -e -o pipefail -- makemake.sh no_gmp k1om
+        (cd obj_k1om; make clean)
+        for word in '' 1word 2word 3word 4word nword; do
+            echo -e "\nMfactor $word\n"
+            bash -e -o pipefail -- makemake.sh no_gmp mfac $word k1om || true
```
MarkRose, line 260: This `|| true` is new, and it's masking six real build failures — the job now reports success while producing no `Mfactor` binary at all. From the K1OM job log on this head:

```
Mfactor
*** There were build errors - see 'obj_mfac_k1om/build.log' for details. ***
../src/factor.c:3217: Error: `vmovups' is not supported on `k1om'
../src/factor.c:3223: Error: invalid instruction suffix for `kmov'
../src/factor.c:3229: Error: no such instruction: `vprolvd %zmm0,%zmm31,%zmm1{%k1}'
make: *** [Mfactor] Error 1
```

…and the same for `1word` through `nword`.

The cause is a missing IMCI carve-out. `factor.c:3176` opens the sieve bit-clearing block with a bare `#ifdef USE_AVX512`, and `platform.h:263` makes `USE_IMCI512` imply `USE_AVX512` — so KNC is handed AVX-512F (`vpbroadcastd`, `vmovups`, `vprolvd`, `kmovw`), none of which exist in IMCI. `factor.c` never mentions `USE_IMCI512`, whereas `mi64.c` already uses the right idiom for this at lines 2414 and 4221.

One line fixes it:

```diff
--- a/src/factor.c
+++ b/src/factor.c
@@ -3176,7 +3176,7 @@
-		#ifdef USE_AVX512	// Use vector-int math and gather-load/scatter-store to accelerate the bit-clearing
+		#if defined(USE_AVX512) && !defined(USE_IMCI512)	// Use vector-int math and gather-load/scatter-store to accelerate the bit-clearing
```

The `#else` arm below it is the portable scalar path, which is what KNC wants.

I built the MPSS toolchain in a container and measured it against this branch. Before: all six word variants fail at `factor.c:3212`, matching CI. After the one-line change: all six build, and they're real KNC binaries (`e_machine=181`, K1OM). It's also inert everywhere else — preprocessed `factor.c` for an x86 AVX-512 build is byte-identical before and after, 520,528 bytes each, checked against a `#if 0` control that did show a difference.

Happy to put that up as its own PR since it's a source change and this one is CI-only — then this line can drop the `|| true` and the job would actually be testing something. Let me know which you'd prefer.

MarkRose:
Thanks for the rebase, and the `obj_mfac${word:+_$word}_k1om` fix is right.

On the KNL flags:

> Both GCC and Clang dropped both KNL and AVX512 ER support at the same time.

Not quite — measured just now:

| flag | clang 22.1.8 | gcc 16.1.1 |
|---|---|---|
| `-march=knl` | **accepted** | rejected |
| `-mavx512f` | accepted | — |
| `-mavx512er` | rejected (`unknown argument`) | rejected |
| `-mavx512pf` | rejected | rejected |

So on current Clang `-march=knl` still works while ER and PF are gone. Since the probe tests `-march=knl -mavx512f -mavx512er` as a single unit, that's exactly the toolchain where the mode gets skipped altogether rather than built without ER. Whether the degraded build is worth having is your call — I just wanted the decision made against the right numbers.

Your explanation of the third probe block makes sense, thanks.

I've left a separate note on the `|| true` in the K1OM job — it's currently hiding six `Mfactor` build failures, with a one-line source fix that clears all of them.

.github/workflows/ci.yml
tdulcet: This job is successful without the `|| true` in the #238 integration branch (see [here](https://github.com/primesearch/Mlucas/actions/runs/30757814797/job/91522942399)), so I believe this issue is already fixed in one (or more) of the pending PRs. The  `|| true` should be removed after those PR(s) are merged.

tdulcet:
> Whether the degraded build is worth having is your call — I just wanted the decision made against the right numbers.

Personally, I do not believe testing a KNL build is worth it without AVX512 ER support, as otherwise it is effectively the same as the plain AVX512 (F + CD) build. See #254. However, this could be reevaluated by #239 if needed.

MarkRose (approved):
Agreed, and I withdraw the suggestion.

Once a KNL build without ER is effectively the plain AVX-512 (F + CD) build, the all-or-nothing probe here is the right behaviour — skipping the mode is better than producing something that adds no coverage. My suggestion to combine the two approaches would have made that worse, not better.

For completeness on the one nuance: `-march=knl` still changes tuning and scheduling even with ER and PF gone, so the object code isn't literally identical —

```
plain avx512 (F+CD+FMA): 17c75e1f997f83ae
knl minus ER           : 629dee4ebd382d08
```

— but that's codegen variation, not ISA coverage, and it isn't worth a CI job. It also doesn't affect #239, which pins ubuntu-24.04 precisely because gcc 13 and clang 18 still have ER.

Approving. The `|| true` note I left inline stands as a follow-up rather than a blocker: what it's hiding is a source defect (`factor.c:3176` has a bare `#ifdef USE_AVX512` where `platform.h` makes `USE_IMCI512` imply it), not anything this PR introduced, and this PR is CI-only. Once that one-liner lands the mask can come off and the job will start producing real KNC binaries instead of discarding them.

Worth flagging that this and #239 both touch the same region of `ci.yml`, so whichever lands second will need a rebase.

.github/workflows/ci.yml
```diff
@@ -211,6 +226,52 @@ jobs:
...
+  Linux-K1OM:
+    name: Mlucas K1OM
+
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v7
+    - name: Download
+      run: |
+          # https://web.archive.org/web/20210307201727/https://software.intel.com/content/www/us/en/develop/articles/intel-manycore-platform-software-stack-mpss.html
+          wget -nv 'https://web.archive.org/web/20200818141734id_/http://registrationcenter-download.intel.com/akdlm/irc_nas/15904/mpss-3.8.6-linux.tar'
```
tdulcet, line 238: Similar to the Intel SDE, we may want to host MPSS as well, instead of downloading it from archive.org here. Maybe something for a follow up.

PR #161, opened 2026-07-26
threadpool: fix unsynchronized busy-wait on free_tasks_queue (data race)

MarkRose:
## Problem

Every call site that waits for a threadpool's dispatched tasks to complete polls
`tpool->free_tasks_queue.num_tasks` directly in a sleep-loop, e.g.:

```c
while(tpool && tpool->free_tasks_queue.num_tasks != pool_work_units) {
    ASSERT(0 == mlucas_nanosleep(&ns_time), "nanosleep fail!");
}
```

`free_tasks_queue.num_tasks` is only ever mutated under `pool->free_tasks_mutex` by worker
threads (`threadpool.c`'s enqueue/dequeue paths in `worker_thr_routine`), but every caller reads
it with **no lock at all**. That's a data race by construction, present at all ~38 call sites:
`mers_mod_square.c`, `fermat_mod_square.c`, `pm1.c`, `factor.c`, and all 34
`radixNN_ditN_cy_dif1.c` leading-radix carry routines.

`threadpool.c` already ships a correctly-synchronized primitive for exactly this,
`threadpool_drain(pool, blocking)` — it locks `free_tasks_mutex` and blocks on the pool's own
`tasks_done_cond` until the free-tasks queue is full. It's declared in `threadpool.h` and defined
in `threadpool.c`, but was never called anywhere in the codebase; every caller hand-rolled an
unsynchronized version instead.

## How this was found

While investigating an intermittent AVX-512-only SIGSEGV (`radix32_wrapper_square.c`, `a[rdum]`
read), a ThreadSanitizer build reproduced the crash and additionally flagged a genuine data race:
a worker thread still executing a carry-step chunk (`cy16_process_chunk`, `radix32_dif_pass`) racing
against the main thread concurrently `realloc()`-ing the shared DWT-weights/roots arrays
(`wt0`/`wt1`/`tmp`) for the next FFT length in `mers_mod_square.c`'s `first_entry` re-init path —
racing on the exact "is the pool done yet" check this PR replaces.

(That specific SIGSEGV turned out to have a second, deeper, not-yet-root-caused issue beyond this
race — see follow-up investigation — but this race is real, independently confirmed, and worth
fixing regardless.)

## Fix

Mechanical, uniform substitution at all ~38 sites:

```c
// before
while(tpool->free_tasks_queue.num_tasks != N) {
    ASSERT(0 == mlucas_nanosleep(&ns_time), "...");
}
// after
ASSERT(0 == threadpool_drain(tpool, TRUE), "threadpool_drain failed!");
```

(preserving the `if(tpool)` guard where the original had `tpool &&`), plus removal of the
now-dead `ns_time`/`ns_retval` locals whose only use was the removed loop.

Semantics are unchanged: `threadpool_drain` checks the free-tasks queue against its `max_tasks`
(== `queue_size`), which every caller's own `threadpool_init` call already sized to exactly the
value (`pool_work_units` or `NTHREADS`) it was comparing against — confirmed file-by-file.

## Testing

- Clean build (clang 18, AVX-512, no sanitizers): compiles with 0 errors, binary links.
- ThreadSanitizer + clang 18 + AVX-512, under Intel SDE (`sde64 -skx`), full `-s t` self-test
  sweep: the realloc-vs-worker-read data race no longer appears (previously reproduced on every
  run at multiple FFT-length transitions). No new races introduced.

## Scope

`threadpool.c`/`threadpool.h` themselves are untouched — `threadpool_drain` already existed and
already did the right thing. This PR only replaces callers' hand-rolled unsynchronized wait with
the existing correct API.

### DIFF
--- a/src/factor.c
+++ b/src/factor.c
@@ -2225,9 +2225,6 @@ candidate factors that survive sieving.	*/
   #ifdef MULTITHREAD
 
 //	printf("start; #tasks = %d, #free_tasks = %d\n", tpool->tasks_queue.num_tasks, tpool->free_tasks_queue.num_tasks);
-	struct timespec ns_time;	// We want a sleep interval of 0.1 mSec here...
-	ns_time.tv_sec  =      0;	// (time_t)seconds - Don't use this because under OS X it's of type __darwin_time_t, which is long rather than double as under most linux distros
-	ns_time.tv_nsec = 10000000;	// (long)nanoseconds - Get our desired 0.1 mSec as 10^5 nSec here
 
 	// Populate the work-unit-encoding data structs which will get done by our pool of threads.
 	// Current assignment may be restart of a partially-completed run, in which case npass < TF_PASSES
@@ -2349,10 +2346,7 @@ candidate factors that survive sieving.	*/
 		#endif
 		}
 
-		while(tpool->free_tasks_queue.num_tasks != NTHREADS) {
-			// Posix sleep() too granular here; use finer-resolution, declared in <time.h>; cf. http://linux.die.net/man/2/nanosleep
-			ASSERT(0 == mlucas_nanosleep(&ns_time), "nanosleep fail!");
-		}
+		ASSERT(0 == threadpool_drain(tpool, TRUE), "threadpool_drain failed!");
 		fprintf(stderr,"\n");	// For pretty-printing, have the inline-pass-printing reflect || work, newlines reflect sync-points
 	};	// wave-loop
 
--- a/src/fermat_mod_square.c
+++ b/src/fermat_mod_square.c
@@ -1350,17 +1350,7 @@ for(iter=ilo+1; iter <= ihi && MLUCAS_KEEP_RUNNING; iter++)
 	}
 
 //	printf("start; #tasks = %d, #free_tasks = %d\n", tpool->tasks_queue.num_tasks, tpool->free_tasks_queue.num_tasks);
-	struct timespec ns_time;	// We want a sleep interval of 0.1 mSec here...
-	ns_time.tv_sec  =      0;	// (time_t)seconds - Don't use this because under OS X it's of type __darwin_time_t, which is long rather than double as under most linux distros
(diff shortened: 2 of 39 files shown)


MarkRose:
Cross-checking this against #192 under ThreadSanitizer, since the two turn out to be complementary halves of the same problem.

Built `Mfactor_1word` `-O1 -fsanitize=thread` and ran `-m 2147483647 -bmax 50 -nthread 4` (4 passes, *Performed 657696 trial divides*), with #193 + #82 applied throughout so Mfactor builds at all:

| applied | TSan reports | worker↔worker | main↔worker | exit |
| build fixes only | 6 | 3 | 3 | 66 |
| + #192 | 6 | 2 | 5 | 66 |
| + #192 + **this PR** | **0** | **0** | **0** | **0** |

The races this PR removes are the main↔worker ones — TSan names the spin directly, reporting `As if synchronized via sleep: nanosleep ... main factor.c:2372` against a worker's access to the shared q-tried counter. #192 removes the worker↔worker ones (`PerPass_tfSieve` declared its mutexes as non-static locals, so each thread locked a private copy). Neither is sufficient on its own; with both, the run is TSan-clean.

No change requested here — just evidence that this PR does what it says, and a note that it pairs with #192.

tdulcet (approved):
Thanks, this is a huge improvement.

I am curious, did you benchmark how much this change improves performance with Mlucas using all CPU cores (or maybe deceases power usage), as busy waits are typically very wasteful of CPU cycles.

On an Unix system with Mlucas fully loaded, I have noticed that the load average is always greater than the number of CPU threads, so maybe this accounts for that. For example, on a 4 thread CPU with nothing but Mlucas running, the load average is consistently ~4.7.

MarkRose:
I did not benchmark this, but it's better to be slightly slower and correct than fast and wrong. The problem is that the lock wasn't be used for reads and often incorrect data was being read by other threads. This led to incorrect results or sigsegvs.

PR #14, opened 2024-03-24
Fixed buffer overflow errors in mi64.c

tdulcet:
Patch submitted by Gary Gostin [on the forum](https://www.mersenneforum.org/showpost.php?p=653544&postcount=8).

* Fixed buffer overflow errors in `mi64.c`. Fixes #13

PR #191, opened 2026-07-29
mi64_div_binary: size the scratch for the left-justified y-copy (heap buffer overflow)

MarkRose:
### The bug

`mi64_div_binary()` sets up its scratch space as

```c
xloc = scratch       ; mi64_set_eq(xloc, x, lenX);
yloc = scratch + lenX; mi64_set_eq(yloc, y, lenY); mi64_clear(yloc+lenY, lenX-lenY);
```

The y-copy is **left-justified to `lenX` limbs**, so `yloc` spans `scratch[lenX, 2*lenX)` and the routine needs **`2*lenX`** limbs. The allocation asks for `lenX + lenY + 16`, which is short whenever `lenX > lenY + 16` — and the trailing `mi64_clear()` then writes past the end of the buffer.

The commented-out sizing immediately above had it right — *"Alloc yloc same as x to allow for left-justification of y-copy"* — and the line that replaced it did not.

### Probably the cause of the corruption already recorded here

The comments in this same block document two heap-corruption incidents:

> May 2022: … hit SIGABRT here with `realloc(): invalid next size` …
> Jun 2022: Again hit error … `incorrect checksum for freed object - object was probably modified after being freed.`

Those are the classic signatures of a write just past a heap block, and they were hit on exactly this code path (the cofactor-is-prime-power GCD on F25 with known factors). Both were worked around — switching `realloc`→`malloc`, then adding a slack term — rather than by correcting the size.

### Evidence

Standalone probe: `mi64_div()` with a 127-limb numerator and a 36-limb odd divisor — the shape `Suyama_CF_PRP` feeds it for a 2276-bit factor product. Built `-Og -fsanitize=address -DUSE_THREADS` against each revision:

| revision | result |
| --- | --- |
| `upstream/main` @47d75d3 | `AddressSanitizer: heap-buffer-overflow`, `WRITE of size 8`, `0 bytes after 992-byte region`, in `mi64_clear` (`mi64.c:300`) |
| with this fix | no ASan report; returns `0`, `q[0] = 4934415073450713659` |

### Why this cannot change any result

The change only ever *enlarges* the allocation: `MAX(lenX,lenY) >= lenY`, so both the threshold test and the new size are `>=` their previous values. No arithmetic is touched.

### Not verified here

- I did not reproduce the multi-exponent production symptom end to end. It was observed independently while preparing #183: four multi-exponent PRP-CF sweeps aborted with `_int_malloc assertion failed` / `realloc(): invalid pointer` immediately after the `Checking prime-power-ness via GCD(A - B,C) ...` line, on the **pre-fix** binary. Re-running the single failing exponent alone completes cleanly, so triggering it needs several assignments in one process.
- Only the `static` scratch's *size* is corrected here. Whether that scratch should be `static` at all (it is not thread-safe, and it is what makes the corruption span calls) is a separate question I have not addressed.

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -5685,7 +5685,12 @@ int mi64_div_binary(const uint64 x[], const uint64 y[], uint32 lenX, uint32 lenY
 	allow for left-justification of y-copy"); the sizing that replaced it did not: */
 	if(lens < (lenX + MAX(lenX,lenY))) {
 		// lens = MAX(1024,lenX + lenY);	// Alloc yloc same as x to allow for left-justification of y-copy
-		lens = lenX + MAX(lenX,lenY) + 16;        // GG: bug fix: Always add some extra buffer length. 16 is arbitrary and conservative.
+		/* Exactly the requirement, no fudge. The '+ 16' this replaces came from #14's conservative
+		padding, added while the sizing itself was wrong; with the sizing correct it buys nothing and
+		costs detection - the overflow fixed here only bit when lenX > lenY + 16, i.e. those 16 spare
+		limbs are precisely what hid it until the gap grew past them. Sizing exactly means the next
+		undersizing mistake trips ASan on the first limb rather than the seventeenth: */
+		lens = lenX + MAX(lenX,lenY);
 		/*** May 2022: In preparing for the cofactor-is-prime-power GCD on F25/[known factors], build on Linux
 		with GCC 9.2.1, hit SIGABRT 		here with 'realloc(): invalid next size'. Step-thru debug showed
 		the #limbs-allocated counter lens increasing from 0 to 4 to 9, next jump from 9 to 1048574 triggered
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -5676,10 +5676,16 @@ int mi64_div_binary(const uint64 x[], const uint64 y[], uint32 lenX, uint32 lenY
 	ylen = mi64_getlen(y, lenY);
 	ASSERT(ylen != 0, "divide by 0!");
 
-	// Allocate the needed auxiliary storage - the 2 yloc = ... / mi64_set_eq calls below copy (lenX + lenY) limbs into scratch, so alloc at least that much:
-	if(lens < (lenX + lenY)) {
+	/* Allocate the needed auxiliary storage. NOTE the requirement is 2*lenX limbs, not (lenX + lenY):
+	the y-copy below is *left-justified to lenX limbs* -
+		xloc = scratch       ; mi64_set_eq(xloc, x, lenX);
+		yloc = scratch + lenX; mi64_set_eq(yloc, y, lenY); mi64_clear(yloc+lenY, lenX-lenY);
+	- so yloc spans scratch[lenX, 2*lenX), and the mi64_clear() runs off the end of the allocation
+	whenever lenX > lenY + 16. The commented-out line below had this right ("Alloc yloc same as x to
+	allow for left-justification of y-copy"); the sizing that replaced it did not: */
+	if(lens < (lenX + MAX(lenX,lenY))) {
 		// lens = MAX(1024,lenX + lenY);	// Alloc yloc same as x to allow for left-justification of y-copy
-		lens = lenX + lenY + 16;        // GG: bug fix: Always add some extra buffer length. 16 is arbitrary and conservative.
+		lens = lenX + MAX(lenX,lenY) + 16;        // GG: bug fix: Always add some extra buffer length. 16 is arbitrary and conservative.
 		/*** May 2022: In preparing for the cofactor-is-prime-power GCD on F25/[known factors], build on Linux
 		with GCC 9.2.1, hit SIGABRT 		here with 'realloc(): invalid next size'. Step-thru debug showed
 		the #limbs-allocated counter lens increasing from 0 to 4 to 9, next jump from 9 to 1048574 triggered


src/mi64.c
```diff
@@ -5670,10 +5670,16 @@ int mi64_div_binary(const uint64 x[], const uint64 y[], uint32 lenX, uint32 lenY
...
+	the y-copy below is *left-justified to lenX limbs* -
+		xloc = scratch       ; mi64_set_eq(xloc, x, lenX);
+		yloc = scratch + lenX; mi64_set_eq(yloc, y, lenY); mi64_clear(yloc+lenY, lenX-lenY);
+	- so yloc spans scratch[lenX, 2*lenX), and the mi64_clear() runs off the end of the allocation
+	whenever lenX > lenY + 16. The commented-out line below had this right ("Alloc yloc same as x to
+	allow for left-justification of y-copy"); the sizing that replaced it did not: */
+	if(lens < (lenX + MAX(lenX,lenY))) {
 		// lens = MAX(1024,lenX + lenY);	// Alloc yloc same as x to allow for left-justification of y-copy
-		lens = lenX + lenY + 16;        // GG: bug fix: Always add some extra buffer length. 16 is arbitrary and conservative.
+		lens = lenX + MAX(lenX,lenY) + 16;        // GG: bug fix: Always add some extra buffer length. 16 is arbitrary and conservative.
```
tdulcet: Is the `+ 16` still needed here. This was previously fixed in #14, so maybe that needs to be reevaluated.

MarkRose:
Good question — it isn't, and it's worth removing rather than leaving on top of a now-correct size.

The requirement is exactly `2*lenX` limbs, which falls out of three lines:

```c
xloc = scratch       ; mi64_set_eq(xloc, x, lenX);
yloc = scratch + lenX; mi64_set_eq(yloc, y, lenY); mi64_clear(yloc+lenY, lenX-lenY);
```

the last of which writes through index `lenX + lenY + (lenX-lenY) - 1` = `2*lenX - 1`. So `lenX + MAX(lenX,lenY)` is exact — `MAX` rather than plain `lenX` only because the `lenX < lenY` early-return happens *after* this allocation, so `lenY` may still be the larger when the size is chosen.

That arithmetic also shows what the padding was doing. The old `lenX + lenY + 16` falls short exactly when `lenX + lenY + 16 < 2*lenX`, i.e. **when `lenX > lenY + 16`**. So the 16 spare limbs weren't protecting against this overflow — they were hiding it until the length gap grew past them, which fits the two heap-corruption incidents recorded in the comments right here, both intermittent. Keeping the fudge would preserve a 16-limb window in which the next sizing mistake goes unnoticed.

**Verified** that the exact size doesn't regress anything reachable: an ASan (`-fsanitize=address -O1`) Mfactor build gives `mi64_div quicktest passes` twice and `Factoring self-tests completed successfully`, with no ASan diagnostic.

**Not claimed:** that run doesn't demonstrate the original overflow — the *base* sizing passes the same ASan test, because the self-test never reaches `lenX > lenY + 16`. Showing the fault directly needs a standalone driver, which I couldn't get to link against these TUs today. The case for the change rests on the index arithmetic above, which is checkable by inspection.

tdulcet (approved):

> Showing the fault directly needs a standalone driver, which I couldn't get to link against these TUs today.

Maybe try testing again before merging, since this has now occurred twice.

MarkRose:
Fair, and you were right to push — it does reproduce, I just hadn't got the link right before. The blocker was that `Mlucas.c` carries `main()`, so linking the whole object set fails, and dropping it leaves `RADIX_VEC`/`NRADICES`/`cbuf`/`STATFILE`/`SYSTEM_RAM` undefined. Compiling only the translation units `mi64_div_binary` actually needs, and declaring those few externs in the driver with the types the headers give them, links cleanly.

Built with `-Og -fsanitize=address` against `main` (unfixed) and against this branch, same driver both times.

## Unfixed

```
==298103==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7c86a3de0308
WRITE of size 8 at 0x7c86a3de0308 thread T0
    #0 in mi64_clear src/mi64.c:300
    #1 in mi64_div_binary src/mi64.c:5718
allocated by thread T0 here:
    #1 in mi64_div_binary src/mi64.c:5694
```

which is exactly the described path: the `mi64_clear(yloc+lenY, lenX-lenY)` at 5718 running off the allocation made at 5694.

## Fixed

```
calling mi64_div_binary with lenX=64 lenY=1 (gap 63 > 16)
returned, lenQ=64
rc=0, no sanitizer errors
```

## The boundary is where the sizing predicts

Sweeping `lenX` with `lenY = 1` on unfixed code puts the transition exactly at the `+ 16`:

| lenX | gap | unfixed | fixed |
| --- | --- | --- | --- |
| 15 | 14 | clean | clean |
| 16 | 15 | clean | clean |
| 17 | 16 | clean | clean |
| 18 | 17 | **heap-buffer-overflow** | clean |
| 19 | 18 | **heap-buffer-overflow** | clean |
| 24 | 23 | **heap-buffer-overflow** | clean |
| 64 | 63 | **heap-buffer-overflow** | clean |
| 512 | 511 | — | clean |

So the sixteen spare limbs really were the whole of what hid this, which is the argument for sizing exactly rather than keeping a fudge factor.

## Driver

Compile from a checkout root:

```
gcc -std=gnu99 -w -g -Og -fsanitize=address -D_GNU_SOURCE -DINCLUDE_GMP=0 -Isrc \
    driver.c src/mi64.c src/types.c src/imul_macro.c src/util.c src/qfloat.c \
    src/get_fp_rnd_const.c src/rng_isaac.c src/getRealTime.c src/get_cpuid.c \
    -o driver -lm -lpthread -lrt
./driver 64 1
```

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "types.h"
#include "mi64.h"
/* mi64_div_binary left-justifies its y-copy to lenX limbs, so it needs
   lenX + MAX(lenX,lenY) limbs of scratch. Sizing it lenX+lenY+16 overflows
   as soon as lenX > lenY + 16. Pick lenX far past that. */

/* Globals/functions that live in TUs this driver deliberately does not link
   (Mlucas.c carries main(); the twopmodq family is unrelated to mi64_div_binary). */
char cbuf[STR_MAX_LEN*3], g_cstr[STR_MAX_LEN];
char STATFILE[STR_MAX_LEN];
uint32 SYSTEM_RAM = 0, MAX_RAM_USE = 0;
uint64 twopmodq64(uint64 p, uint64 q) { (void)p; (void)q; return 0ull; }
int    test_twopmodq64(void) { return 0; }

int main(int argc, char **argv)
{
	int LX = atoi(argv[1]), LY = atoi(argv[2]);
	static uint64 x[4096], y[4096], q[4096], r[4096];
	uint32 lenQ = 0;
	int i;
	for(i = 0; i < LX; ++i) x[i] = 0x123456789ABCDEFull + i;
	y[0] = 3ull;
	printf("calling mi64_div_binary with lenX=%d lenY=%d (gap %d > 16)\n", LX, LY, LX-LY);
	mi64_div_binary(x, y, LX, LY, q, &lenQ, r);
	printf("returned, lenQ=%u\n", lenQ);
	return 0;
}
```

Not pushed: the branch is unchanged and still merges cleanly, so there is nothing here that should cost the approval.

MarkRose:
One follow-up: the commit message still carried the paragraph saying the fault could not be shown directly, and since this is a single-commit PR that text would have become the squash-merge message on `main` — permanently recording that the fix was never demonstrated, which is no longer true.

Amended it to carry the ASan trace and the boundary sweep instead. **The tree is byte-identical to what you approved** — `git rev-parse HEAD^{tree}` matches the previous head exactly, so this is a message-only change with no code difference to re-review. Sorry it costs the approval regardless.

PR #194, opened 2026-07-29
twopmodq*: make inline-asm labels unique per asm statement (LTO build break)

MarkRose:
### The build break

At `makemake.sh`'s default `-O3 -flto=auto`, Mfactor fails to assemble:

```
/tmp/ccXXXX.s: Assembler messages:
../src/twopmodq96.c:706:  Error: symbol `LoopBeg' is already defined
../src/twopmodq96.c:1023: Error: symbol `twopmodq96_q4_pshiftjmp' is already defined
../src/twopmodq96.c:1036: Error: symbol `LoopEnd' is already defined
```

The jump targets inside these inline-asm blocks are written as ordinary global labels. That survives a non-LTO build, where each translation unit becomes its own assembly file — but LTO merges translation units into one, so a label gets emitted more than once and the assembler rejects the duplicate.

Two distinct ways it bites here:

* the **same label name is used in two different files** — `twopmodq80.c` also defines a label literally named `twopmodq96_q4_pshiftjmp`;
* a function carrying such a label can be **inlined into more than one caller**, which is why even the uniquely-suffixed `LoopBeg4a`/`LoopEnd4a` in `twopmodq.c` are unsafe.

Note `-flto=auto` hides the real error behind `lto-wrapper: fatal error: make returned 2`; building with `-flto=1` surfaces it.

### The fix

Append gcc's `%=` operator to each label. It substitutes a number unique to each instance of the asm statement, so the emitted symbol is unique per instantiation while the source stays readable.

All of these blocks are **extended** asm (named operands, explicit clobber lists), so `%=` is available. And every label's definition and all its references live inside a **single** asm statement — I checked this explicitly, since `%=` is per-statement and a label referenced across two statements would silently break.

Labels updated (live code only; copies inside `#if 0` blocks are left alone):

| file | labels |
| --- | --- |
| `twopmodq96.c` | `LoopBeg`, `LoopEnd`, `twopmodq96_q4_pshiftjmp` |
| `twopmodq.c` | `LoopBeg4a`, `LoopEnd4a` |
| `twopmodq80.c` | `twopmodq96_q4_pshiftjmp` |
| `twopmodq64_test.c` | `twopmodq64_q4_pshiftjmp`, `twopmodq64_q8_pshiftjmp` |

### Verification

**No code change.** Compiling each of the four files at `-O3` with the Mfactor flag set and diffing disassembly, the only differences are objdump's symbol annotations on jump targets — `<LoopEnd+0x6a6>` became `<LoopEnd1312+0x6a6>`. Normalising the label names away, all four files are **instruction-identical**, same addresses and same offsets.

| check | result |
| --- | --- |
| `twopmodq96.c` / `twopmodq.c` / `twopmodq80.c` / `twopmodq64_test.c` disasm, label names normalised | identical |
| `makemake.sh mfac 1word` at default `-O3 -flto=auto` | builds (previously: assembler errors) |
| built binary's startup self-test | passes, 0 errors — including *"Testing 64-bit 2^p (mod q) functions with 100000 random (p, q odd) pairs"*, which exercises precisely these asm blocks |

### Depends on #193

This is only observable once #193 is applied — without it the Mfactor build fails earlier, at the `USE_FMADD` path's calls to the nonexistent `twopmodq100_2WORD_DOUBLE` routines. Both were applied together for the verification above. With #193 + this change, `makemake.sh mfac 1word` completes at the default flags for the first time on an AVX2 host.

### DIFF
--- a/src/twopmodq.c
+++ b/src/twopmodq.c
@@ -1407,8 +1407,8 @@ void twopmmodq64_q4(uint64 p, uint64 *i0, uint64 *i1, uint64 *i2, uint64 *i3, ui
 		"movslq	%[__start_index], %%rcx		\n\t"\
 		"subq $1,%%rcx						\n\t"\
 		"test %%rcx, %%rcx					\n\t"\
-		"jl LoopEnd4a		/* Skip if n < 0 */	\n\t"\
-	"LoopBeg4a:								\n\t"\
+		"jl LoopEnd4a%=		/* Skip if n < 0 */	\n\t"\
+	"LoopBeg4a%=:								\n\t"\
 	"/* SQR_LOHI_q4(x*, lo*, hi*): */	\n\t"\
 		"movq	%%r12,%%rax	/* x0-3 in r8-11. */\n\t"\
 		"mulq	%%rax		\n\t"\
@@ -1510,8 +1510,8 @@ void twopmmodq64_q4(uint64 p, uint64 *i0, uint64 *i1, uint64 *i2, uint64 *i3, ui
 		"/* } endif((p >> j) & (uint64)1) */						\n\t"\
 		"subq	$1,%%rcx	/* j-- */		\n\t"\
 		"cmpq	$0,%%rcx	/* compare j vs 0 */\n\t"\
-		"jge	LoopBeg4a	/* if (j >= 0), Loop */	\n\t"\
-	"LoopEnd4a:							\n\t"\
+		"jge	LoopBeg4a%=	/* if (j >= 0), Loop */	\n\t"\
+	"LoopEnd4a%=:							\n\t"\
 		"movq	%%r12,%[__x0]	\n\t"\
 		"movq	%%r13,%[__x1]	\n\t"\
 		"movq	%%r14,%[__x2]	\n\t"\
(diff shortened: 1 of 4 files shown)


ldesnogu:
> Labels updated (live code only; copies inside #if 0 blocks are left alone)

I wonder if it wouldn't make sense to also fix ifdef'ed-out code too so that in the future if someone wants to re-enable it he won't hit that issue (that's assuming any such code exists).

MarkRose:
Closing as a duplicate of #82, which I had overlooked when opening this.

#82 makes the same fix by the same means — appending gcc's `%=` unique-id operator to the inline-asm labels — in the same four files, and it does so more completely: **34 label sites vs the 16 here**. It also predates this PR by several weeks, was already approved once, and carries the CI commits that wire Mfactor's `test_fac` self-test into the build matrix, which is the more valuable half of that branch.

For the record, since it came up while comparing the two: I could not find any build configuration in which the 18 labels this PR leaves untouched are actually emitted. Compiling all four TUs to assembly at `nosimd`, `sse2` and `avx2` (and with `TRYQ=4`, `factor.h`'s default), none of `Loop8Beg`/`Loop8End`/`Loop4Beg`/`Loop4End`/`LoopBeg4`/`LoopEnd4`/`LoopBeg8`/`LoopEnd8`/`twopmodq96_q8_pshiftjmp`/`twopmmodq64_q4_pjmp`/`twopmodq78_3wdq2_gcc64` appears in the output. So the extra coverage in #82 is defensive rather than fixing an observable collision — worth having, but nobody should expect a behaviour change from it.

A tree-wide scan of an AVX-512 build (100 TUs compiled to assembly, every emitted label classified) also found **zero duplicate global symbols**. The ~106 file-local names shared across TUs are all compiler-generated statics — `__func__.N`, `first_entry.N`, `task_control.N` — plus the `minv8`/`brev8` lookup tables, which are `static const` in `mi64.h` and therefore one private copy per including TU by design. None of that is a collision.

No work is lost by closing this: #82 is a superset.

PR #198, opened 2026-07-29
imul_macro1.h: MULH128x96 and MULH128 drop a real carry out of __b

MarkRose:
Fifteen copies of this construct discard a real carry:

```c
__a  += __c;
__b  += __d + (__a < __c);
DBG_ASSERT((__b >= __d),"...: unexpected carryout of __b");
```

The comment above them claims *"since b and d <= 2^64 - 2, can add carryout of a+c sans ripple-carry check"* — but that is not the shape of the operands. In `MULH128x96`, `(__a,__b)` is the 96-bit product `x.d0*y.d1`, so `__b < 2^32`; `(__c,__d)` is the **full 128-bit** product `y.d0*x.d1`, with `__d` up to `2^64-2`. Their 128-bit sum reaches `2^128 + 2^96`, so `__b` genuinely carries out.

The `DBG_ASSERT` is not a guard — `masterdefs.h` defines it away to nothing unless `FAC_DEBUG`, so production silently returns `hi.d1` one too small.

### The fix

`(__a,__b)` is added at the `[__w1,__w2]` positions, so a carry out of `__b` has weight 2^128 — it belongs in `__w3` (or `__hiN.d1` in the pipelined variants).

`__d` and the carry-in are now added **separately** rather than as `__d + (__a < __c)`, so the carry test cannot itself be fooled: when `__d = 2^64-1` and there is a carry in, that inner sum wraps to 0 and a naive `(__b < __d)` test would miss the carry entirely.

**Sites:** `MULH128x96` (both the `MUL_LOHI64_SUBROUTINE` and portable bodies), all 4 lanes of `MULH128x96_q4`, all 8 of `MULH128x96_q8`, and the same construct in `MULH128`'s `MUL_LOHI64_SUBROUTINE` body — where **both** `(__a,__b)` and `(__c,__d)` are full 128-bit products, so it overflows far more often than the `x96` case rather than only when `__d` is within 2^32 of the top — measured at **7.2% of uniform-random operand pairs** (see the driver results in the comments; an earlier estimate of "roughly half" in this description was wrong). The three no-op `DBG_ASSERT`s go with the code they purported to check.

### Verification

Against an exact `__int128` reference.

A uniform-random sweep finds **nothing** — the defect needs `__d >= 2^64 - 2^32`, i.e. both `y.d0` and `x.d1` near 2^64, which is ~2^-32 of the space. (My first sweep did exactly this and returned 0 wrong on *both* builds, which is why the base run matters as a control.) Biasing those two operands to within 2^24 of 2^64:

| | witness `hi.d1` | biased sweep |
| --- | --- | --- |
| pristine `main` | `0x00000000FFFFDA20` (exact: `…DA21`) — **mismatch** | **389844 / 400000 wrong** |
| with this fix | `0x00000000FFFFDA21` — matches | **0 / 400000 wrong** |

The witness is the one recorded in the original audit. `imul_macro.c` compiles clean with and without `-DMUL_LOHI64_SUBROUTINE`.

### Reachability

`MULH128x96_q4`/`_q8` are called from `twopmodq128_96.c`, reached from `factor.c` only under `-DUSE_128x96=2`; the x86_64/GCC default is `1`. So this is **live but configuration-gated**, and at that setting it produces rare silently-wrong residues — missed factors, not a crash. The `MULH128` copy is on the `MUL_LOHI64_SUBROUTINE` path, which the default x86-64/GCC build never compiles. It is now demonstrated directly rather than by inspection: a standalone driver compiling only `imul_macro.c` and `types.c` with `-DMUL_LOHI64_SUBROUTINE` shows 14403/200000 uniform-random pairs wrong on `main` and 0/200000 with this fix.

Companion to #167, which fixed a different dropped cross-term carry in `SQR_LOHI192`. Several further defects from the same audit remain unfixed and are not included here.

### DIFF
--- a/src/imul_macro1.h
+++ b/src/imul_macro1.h
@@ -3679,8 +3679,13 @@ MUL64x32s, i.e. are significantly cheaper than full-blown MUL_LOHIs on
 		MUL64x32(  __x.d1,__y.d1,&__w2,&__w3);\
 		/* First add [a,b] + [c,d] : since b and d <= 2^64 - 2, can add carryout of a+c sans ripple-carry check: */\
 		__a  += __c;\
-		__b  += __d + (__a < __c);\
-		DBG_ASSERT((__b >= __d),"MULH128x96: unexpected carryout of __b");\
+		/* (__a,__b) is a 96-bit product so __b < 2^32, but (__c,__d) is a FULL 128-bit one with
+		__d up to 2^64-2, so this 128-bit sum genuinely carries out of __b - the old code dropped
+		that carry, and the DBG_ASSERT below it expands to nothing unless FAC_DEBUG. The carry has
+		weight 2^128 here, i.e. it belongs in __w3. Add __d and the carry-in separately so the
+		carry test cannot itself be fooled when __d = 2^64-1 and there is a carry in: */\
+		__b  += __d;			__w3 += (__b < __d);\
+		__cy  = (__a < __c);	__b += __cy;	__w3 += (__b < __cy);\
 		/* Now add [w1,w2,w3] + [a,b,0]: */\
 		__w1 += __a;\
 		__cy  = (__w1 < __a);\
@@ -3739,8 +3744,13 @@ MUL64x32s, i.e. are significantly cheaper than full-blown MUL_LOHIs on
 		MUL64x32(  __x.d1,__y.d1, __w2, __w3);\
 		/* First add [a,b] + [c,d] : since b and d <= 2^64 - 2, can add carryout of a+c sans ripple-carry check: */\
 		__a  += __c;\
-		__b  += __d + (__a < __c);\
-		DBG_ASSERT((__b >= __d),"MULH128x96: unexpected carryout of __b");\
+		/* (__a,__b) is a 96-bit product so __b < 2^32, but (__c,__d) is a FULL 128-bit one with
+		__d up to 2^64-2, so this 128-bit sum genuinely carries out of __b - the old code dropped
+		that carry, and the DBG_ASSERT below it expands to nothing unless FAC_DEBUG. The carry has
+		weight 2^128 here, i.e. it belongs in __w3. Add __d and the carry-in separately so the
+		carry test cannot itself be fooled when __d = 2^64-1 and there is a carry in: */\
+		__b  += __d;			__w3 += (__b < __d);\
+		__cy  = (__a < __c);	__b += __cy;	__w3 += (__b < __cy);\
 		/* Now add [w1,w2,w3] + [a,b,0]: */\
 		__w1 += __a;\
 		__cy  = (__w1 < __a);\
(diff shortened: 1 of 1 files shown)


MarkRose:
Worth recording here: **this bug had a second instance in the tree**, and the fix in this PR does not reach it.

`THREE_OP128_q4` and `THREE_OP128_q8` in `imul_macro1.h` hand-inline the same dropped carry:

```c
__b0 += __wd0 + (__a0 < __wc0);   /* carry out of __b0 discarded */
```

— the same shape as the `__b += __d + (__a < __c);` this PR corrects, carrying the identical (incorrect) justifying comment about the sum not being able to carry out. The fix here does not propagate to them because they do not call `MULH128x96`/`MULH128`; they duplicate the vulnerable arithmetic inline.

Measured against a GMP oracle over 400,000 random inputs in the `MULH128x96` shape: **29,097 wrong `lo` outputs for `_q4`, 28,976 for `_q8`** — the same order as this PR's own defect rate, which is what you would expect from the same arithmetic error.

Those macros are dead (`THREE_OP128` is set nowhere in the build system, and `git log -S` shows no commit ever turning it on), so rather than duplicate this fix into unreachable code, they are being deleted in #213.

No action needed on this PR — the two do not conflict. Checked explicitly: this branch predates current `main` (merge-base `47d75d3`), so rebasing it first is necessary for the test to mean anything; rebased onto `main` and merged with #213, the result is a clean merge that builds `nosimd` rc=0. Either order works.

Flagging it mainly because the second instance was found only by auditing dead code — the sort of thing that is easy to assume is not worth the time.

MarkRose:
The `MULH128` copy no longer rests on inspection — it is demonstrated.

The obstacle before was that `MULH128` lives on the `MUL_LOHI64_SUBROUTINE` path, which the default x86-64/GCC build never compiles, so nothing in a normal build reaches it. Compiling a driver against only the two translation units it actually needs, with that macro defined, gets at it directly:

```
gcc -std=gnu99 -w -O2 -D_GNU_SOURCE -DMUL_LOHI64_SUBROUTINE -Isrc \
    driver.c src/imul_macro.c src/types.c -o driver
```

(`ABORT` is stubbed in the driver — it lives in `util.c`, which would drag in most of the tree, and `DBG_ASSERT` compiles out anyway.) Checked against an exact `__int128` reference:

| sweep | `main` | this PR |
| --- | --- | --- |
| `x.d1`, `y.d0` biased to within 2^24 of 2^64 | 400000 / 400000 wrong | **0** / 400000 |
| uniform random | 14403 / 200000 wrong | **0** / 200000 |

first mismatch on `main`, uniform: `hi.d1` = `0x57FF3F76B1F18387`, exact `0x57FF3F76B1F18388` — the expected off-by-one from the discarded carry.

The fixed build returning 0 across 600k cases is also the control on the reference: if my `__int128` model were wrong it would disagree with both builds, not just the base.

## One correction to my own description

I had written that `MULH128` "overflows for roughly *half* of all operand pairs". That was an estimate and it is too high — the measurement puts it at **7.2%** of uniform-random pairs. Still far more often than the `x96` case, and still a silently wrong high word rather than a crash, but I have corrected the PR description rather than leave the inflated figure standing.

The driver is at `plan-notes/pr-replies/pr198-mulh128-driver.c` on my side if it is useful; happy to fold it into the tree as a test if you would rather it did not just live in a PR comment.

PR #202, opened 2026-07-29
imul_macro1.h: parenthesize the pointer parameters of the _PTR macros

MarkRose:
Every `_PTR` macro writes `__x->d0` with no parentheses around `__x`. Since `->` binds tighter than unary `&` and than `+`, any argument that is not a bare pointer variable mis-parses:

* `&arr[i]` becomes `&(arr[i]->d0)`
* `p+1` becomes `p + (1->d0)`

It is a **compile error**, not a silent wrong answer, so the practical effect is that callers are quietly restricted to bare pointer variables — a restriction that is invisible until someone hits it. (Which is how it was found, while writing a test harness for these macros.)

Parenthesized **122 dereferences across 9 macros**: `CMPULT96_PTR`, `CMPEQ96_PTR`, `RSHIFT_FAST96_PTR`, `LSHIFT96_PTR`, `ADD96_PTR`, `SUB96_PTR`, `MULH96_PTR_q4`, `ADD192_PTR`, `ADD256_PTR`. `CMPUGT96_PTR` is defined in terms of `CMPULT96_PTR` and needs nothing.

### Verification

A caller using exactly the shapes that fail today — `&arr[0]`/`&arr[1]` into `CMPULT96_PTR` and `ADD96_PTR`, `p+0`/`p+1` into `CMPEQ96_PTR`:

| | result |
| --- | --- |
| pristine `main` | **18 compile errors** |
| with this fix | compiles, results correct (`CMPULT96_PTR` = 1, `CMPEQ96_PTR` = 0, `ADD96_PTR` `sum.d0` = 3) |

Existing callers are unaffected — `twopmodq96.c`, `twopmodq80.c`, `twopmodq128_96.c` and `mi64.c` all still compile clean, since a bare pointer variable parses identically either way.

### One trap worth recording

The obvious sweep regex is subtly wrong. Guarding with a lookbehind that excludes `)` — to avoid double-parenthesising — skips exactly the cast-preceded cases like `(uint32)__x->d1`, which is *every* dereference in `CMPULT96_PTR` and `CMPEQ96_PTR`. My first pass silently missed both macros for that reason. A plain word-boundary match is both correct and idempotent, because an already-fixed `(__x)->` contains `__x)` and never `__x->`.

Fifth PR against this file in the batch (#198, #199, #200, #201); disjoint regions.

### DIFF
--- a/src/imul_macro1.h
+++ b/src/imul_macro1.h
@@ -75,14 +75,14 @@ that is advantageous on at least an appreciable set of CPUs.
 #define ADD192_PTR(__x, __y, __sum)\
 {\
 	uint64 __s,__t, __cy0;	/* Need these in case any or all of x, y, sum point to the same address. */\
-	__s      = __x->d0;\
-	__t      = __x->d1;\
-	__sum->d0 = __x->d0 + __y->d0;	__cy0 = (__sum->d0 < __s);\
-	__sum->d1 = __x->d1 + __y->d1;\
-	__sum->d2 = __x->d2 + __y->d2 + (__sum->d1 < __t);\
+	__s      = (__x)->d0;\
+	__t      = (__x)->d1;\
+	(__sum)->d0 = (__x)->d0 + (__y)->d0;	__cy0 = ((__sum)->d0 < __s);\
+	(__sum)->d1 = (__x)->d1 + (__y)->d1;\
+	(__sum)->d2 = (__x)->d2 + (__y)->d2 + ((__sum)->d1 < __t);\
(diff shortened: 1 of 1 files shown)


ldesnogu (approved):
This looks good to me.

I wonder if all these macros should be converted to `static inline` functions at some point. This would careful benchmarking to ensure speed isn't regressing.

MarkRose:
@ldesnogu We should definitely look at that when it comes to performance optimizing.

PR #209, opened 2026-07-29
radix32: declare cy_i and addi to match their unconditional use — build regression from 2bf9930

MarkRose:
**This is a build regression in merged code**, from `2bf9930a` ("radix32: restore the addi carry pointer instead of leaving it dangling"), which arrived via #126. That commit restored the unconditional `addi = (double *)cy_i` and the four `++addi` increments in the `MODULUS_TYPE_GENFFTMUL` branch of `radix32_main_carry_loop.h`, but left the matching declarations in `radix32_ditN_cy_dif1()` behind guards that exclude those builds. Nothing about it is specific to v21.

**Nor is it specific to SSE2.** `cy_i` is declared under `#ifdef USE_AVX`, so every non-AVX SIMD build fails identically — ARM ASIMD included, which is why `main` does not currently compile on Apple Silicon. Reported independently on the forum against `418f365`.

**Verified on real ARM hardware** (Raspberry Pi 4, aarch64, gcc 12.2.0):

| tree | `makemake.sh asimd` |
|---|---|
| `main` @ `418f365` | rc=1 — `radix32_main_carry_loop.h:62: 'cy_i' undeclared` |
| `main` + this PR | **rc=0**, binary builds, `-s tt` self-test runs |

Identical error and identical fix at `makemake.sh sse2` on x86, which is the cheap way to reproduce without ARM hardware.

CI caught the regression at the time — all five macOS jobs went from green on the preceding run to red on #126 — so re-running those legs here is the definitive check for anything macOS-specific behind it.

---

## The break

Mlucas does not compile at `sse2` on **either gcc or clang** — the main binary, not Mfactor:

```
$ gcc -Wall -g -O3 -std=gnu99 -D_GNU_SOURCE -c -DUSE_SSE2 -msse2 src/radix32_ditN_cy_dif1.c
src/radix32_main_carry_loop.h:62:47: error: 'addi' undeclared (first use in this function)
src/radix32_main_carry_loop.h:62:64: error: 'cy_i' undeclared (first use in this function)
```

clang 22.1.8 gives the same at `:62:50`. Without `-DUSE_SSE2` the file compiles clean.

## Cause

Both names are the same regression, from `2bf9930`. That commit restored an unconditional

```c
addi = (double *)cy_i;
```

in the `MODULUS_TYPE_GENFFTMUL` branch of `radix32_main_carry_loop.h`, but left the two declarations in `radix32_ditN_cy_dif1()` behind guards that exclude non-AVX SIMD builds — `cy_i` declared only under `USE_AVX`, and `addi` under `#ifndef USE_SSE2`. The multithreaded sibling `cy32_process_chunk()` already declares both unconditionally, which is why only the single-threaded entry point broke.

## Fix

Declare both unconditionally in `radix32_ditN_cy_dif1()`, matching `cy32_process_chunk()`. One file, +7/−11. No functional change to any build that already compiled.

**Deliberately not touched:** `carry.h`'s V1/V2 GENFFTMUL macro split. It carries a comment warning that "fixing" the unused-`cy` warning by dropping the caller's `cy_i` setup *silently breaks the V1 variant*. This change restores the caller-side declarations only.

## Verification

| | gcc 16.1.1 | clang 22.1.8 |
|---|---|---|
| `radix32_ditN_cy_dif1.c` at `-DUSE_SSE2`, before | 1 error | 1 error |
| same, after | **0** | **0** |
| full `rm -rf obj_sse2 && bash ./makemake.sh sse2` | **clean** | **clean** |

No further rotted files turned up behind this one. `./obj_sse2/Mlucas -s tt` runs to completion, with `Res64` matching an `obj_nosimd` build of the same tree at every FFT length both reach.

(Unrelated, noted so it isn't mistaken for fallout: the nosimd log reports some "N of M radix-sets passed — skipping it" warnings at 14K/15K. That is the separately-tracked nosimd scalar-radix issue, which #152 addresses.)

## Why this went unnoticed

There is no gcc/clang `sse2` job for the **Mlucas** target in CI — the sse2 coverage that exists is for other targets. This is the seventh build configuration found rotted in this campaign.

### DIFF
--- a/src/radix32_ditN_cy_dif1.c
+++ b/src/radix32_ditN_cy_dif1.c
@@ -198,8 +198,8 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 	static double radix_inv, n2inv;
   #ifndef MULTITHREAD
 	double *addr;
-	double *addi;	// v21 build fix: must exist in SSE2 builds too - the MODULUS_TYPE_GENFFTMUL
-					// branch of radix32_main_carry_loop.h dereferences it unconditionally (regression from 2bf9930)
+	double *addi;	// Declared unconditionally: the MODULUS_TYPE_GENFFTMUL branch of
+					// radix32_main_carry_loop.h assigns and increments it in every build.
 	double temp,frac;
   #endif
 	double scale, dtmp, maxerr = 0.0;
@@ -255,8 +255,8 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 		,*r20,*r22,*r24,*r26,*r28,*r2A,*r2C,*r2E
 		,*r30,*r32,*r34,*r36,*r38,*r3A,*r3C,*r3E */
 		,*cy_r	// Need RADIX slots for sse2 carries, RADIX/2 for avx
-		,*cy_i	// v21 build fix: must exist in SSE2/MULTITHREAD builds too - radix32_main_carry_loop.h's
-				// MODULUS_TYPE_GENFFTMUL branch dereferences it unconditionally (regression from 2bf9930)
+		,*cy_i	// Declared unconditionally: radix32_main_carry_loop.h's MODULUS_TYPE_GENFFTMUL
+				// branch reads it in every build.
 		;
   #ifdef USE_AVX
 	static vec_dbl *base_negacyclic_root;
@@ -1994,7 +1994,7 @@ void radix32_dit_pass1(double a[], int n)
 			,*r20,*r22,*r24,*r26,*r28,*r2A,*r2C,*r2E
 			,*r30,*r32,*r34,*r36,*r38,*r3A,*r3C,*r3E */
 			,*cy_r
-			,*cy_i	// v21 build fix: see the same declaration in radix32_ditN_cy_dif1() above
+			,*cy_i	// Declared unconditionally - see the same declaration in radix32_ditN_cy_dif1() above
 			;
 	  #ifdef USE_AVX
 		vec_dbl *base_negacyclic_root;
--- a/src/radix32_ditN_cy_dif1.c
+++ b/src/radix32_ditN_cy_dif1.c
@@ -198,9 +198,8 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 	static double radix_inv, n2inv;
   #ifndef MULTITHREAD
 	double *addr;
-   #ifndef USE_SSE2
-	double *addi;
-   #endif
+	double *addi;	// v21 build fix: must exist in SSE2 builds too - the MODULUS_TYPE_GENFFTMUL
+					// branch of radix32_main_carry_loop.h dereferences it unconditionally (regression from 2bf9930)
 	double temp,frac;
   #endif
 	double scale, dtmp, maxerr = 0.0;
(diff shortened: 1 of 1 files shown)


ldesnogu:
That patch concerns me a bit.

Back when I did my warning patches, I carefully removed unused variables and tested all variants (my Mac can cross-compile to x86; this possibility might have been removed now that Apple officially stopped supporting x86, I will have to check). It's certainly possible I missed something. But I wonder why suddenly cy_i would be needed. Is that due to previous changes that were merged? And are these changes correct?

I'm not arguing, I surely do mistakes, I just want to be sure this was one.

src/radix32_ditN_cy_dif1.c
```diff
@@ -198,9 +198,8 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 	static double radix_inv, n2inv;
   #ifndef MULTITHREAD
 	double *addr;
-   #ifndef USE_SSE2
-	double *addi;
-   #endif
+	double *addi;	// v21 build fix: must exist in SSE2 builds too - the MODULUS_TYPE_GENFFTMUL
```
tdulcet: The issue is not related to v21 or only SSE2. As noted on the forum (and the next line below), it was a regression from https://github.com/primesearch/Mlucas/commit/2bf9930af59df1138f5d90deedfedd028774f0e5.

tdulcet (dismissed):
Thanks for this fix.

src/radix32_ditN_cy_dif1.c
MarkRose: Cleaned up the comments

MarkRose:
@ldesnogu — you didn't miss anything, and this wasn't your mistake. I dug through the history to be sure before answering.

Your commit 1c468934 was self-consistent. In the same change you commented out the addi uses in the MODULUS_TYPE_GENFFTMUL branch of radix32_main_carry_loop.h:
```
// TODO: 'addi' use is commented out since 'genfftmul_carry_norm_pow2_errcheck' doesn't use its cy parameter in its current implementation
l = 0; addr = (double *)cy_r; //addi = (double *)cy_i;
```
and removed the declarations that this made unused. Uses and declarations went away together — exactly right, and it built everywhere at the time.

What broke it is my commit 2bf9930a, which landed in #126 as a follow-up to your review question about that very TODO. It restored the assignment and the four ++addi increments, but I did not widen the declaration guards to match. So the uses came back and the declarations didn't, and every build whose guards excluded them stopped compiling — which is every non-AVX SIMD build, ARM ASIMD as much as SSE2. That's also why main currently fails to build on Apple Silicon.

So the sequence is: your cleanup was correct → my follow-up re-introduced the uses → the guards no longer matched. The gap is mine.

On whether this fix is the right one: there were two options — re-comment the uses, or widen the declarations. I went with widening because the multithreaded sibling cy32_process_chunk() already declares both unconditionally, so this makes the single-threaded entry point match rather than diverge further. Re-commenting would also have re-armed the trap 2bf9930a was written to remove: carry.h's #if 0 can select the V1 GENFFTMUL macro, which does use cy, and with the setup commented out that dereferences an uninitialised pointer — something neither gcc nor clang diagnoses at -O2 -Wall -Wextra.

Verified on real hardware rather than by inspection: Raspberry Pi 4, aarch64, gcc 12.2.0 — makemake.sh asimd fails on main @ 418f365 with radix32_main_carry_loop.h:62: 'cy_i' undeclared and builds clean with this PR, self-test included. makemake.sh sse2 on x86 reproduces the identical error, which is the cheap way to check without ARM hardware.

I've also corrected the code comments and the PR title/description — the earlier wording called this a "v21 build fix" and framed it as SSE2-specific, and @tdulcet was right that both were wrong.

tdulcet (approved):
Thanks.

PR #212, opened 2026-07-29
twopmmodq64/128/192/256: fix the even-modulus p-window underflow (resolves #186)

### LINKED ISSUE #186: twopmmodq64/128/192/256: even modulus makes pshift underflow for p just above the routine width
### Summary

In `twopmmodq64/128/192/256`, the `p <= W` early-out (W = routine width in bits) is evaluated **before** the power of 2 is stripped from an even modulus `q`. For `p` in `(W, W + trailz(q)]` the subsequent `p -= nshift` makes `pshift = p - W` wrap, and `2^p mod q` comes back wrong.

This matters because **even moduli are the normal case at the `Mlucas.c` call sites**, where the modulus is `KNOWN_FACTORS[i] - 1`.

### Reproduction

Standalone driver linking the Mfactor object set, calling the routines directly and comparing against Python `pow(2,p,q)`. Modulus chosen as a multiple of 8, so `trailz(q) = 3`:

- `q64  = 0x7FFFFFFFFFFFFFF8`
- `q128 = 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFF8`

| routine | p | `main` @47d75d3 | with #185 applied |
|---|---|---|---|
| `twopmmodq64` | 63 | ok | ok |
| `twopmmodq64` | 64 | ok | ok |
| `twopmmodq64` | **65** | **wrong** | **wrong** |
| `twopmmodq64` | **66** | **wrong** | **wrong** |
| `twopmmodq64` | 67 | ok | ok |
| `twopmmodq64` | 68 | wrong | ok |
| `twopmmodq128` | 127 | ok | ok |
| `twopmmodq128` | 128 | ok | ok |
| `twopmmodq128` | **129** | **wrong** | **wrong** |
| `twopmmodq128` | **130** | **wrong** | **wrong** |
| `twopmmodq128` | 131 | ok | ok |
| `twopmmodq128` | 132 | wrong | ok |

Concretely, `twopmmodq64(65, 0x7FFFFFFFFFFFFFF8)` returns `0x200000` where the correct value is `0x20`; `twopmmodq128(129, …)` returns `0x2000` for a correct `0x20`.

The p = 68 / 132 rows are a **different** defect — the `start_index` underflow fixed by #185 — and are included to show the separation: #185 fixes those, and leaves the p = 65/66/129/130 window untouched. A sweep across all four widths (19,200 cases per width against `pow(2,p,q)`) puts the analogous window at p ∈ {193,194} for `twopmmodq192` and p ∈ {257,258} for `twopmmodq256`; I measured the 64- and 128-bit rows above directly.

`p = 67` is correct because with `nshift = 3` the effective exponent lands exactly on `W`, taking the early-out legitimately.

### Why this is filed rather than fixed

The fix touches the same lines as #172, and the routines are currently being changed by three open PRs. It is cleanest to land those first:

- **#185** — `start_index` underflow (the p = 68/132 rows above). Same files.
- **#172** — recompute `qhalf` after the even-modulus shift. Same files, and the same lines this fix needs.
- **#167** — `SQR_LOHI192` dropped cross-term carry. Disjoint file (`imul_macro1.h`), but `twopmmodq192` is not fully correct without it: after #185, a 19,200-case sweep still shows 3,024 mismatches at 192 bits, all of which vanish when #167 is applied on top.

With #185 + #172 + #167 all merged, this p-window is the **only remaining even-modulus failure** in these routines.

I'm happy to open the fix as a PR once those three are in — or sooner if you'd prefer it stacked on one of them.

MarkRose:
Resolves #186.

## ⚠ This PR is stacked — the first three commits belong to #167, #185 and #172

| commit | PR | what |
|---|---|---|
| `SQR_LOHI192: propagate the dropped cross-term carry into word 4` | **#167** | `src/imul_macro1.h` |
| `twopmmodq64/128/192/256: fix start_index underflow …` | **#185** | |
| `twopmmodq128/192/256: recompute qhalf after the even-modulus right-justify shift` | **#172** | |
| **`… fix pshift underflow for even moduli with p just above the routine width`** | **this PR** | the #186 fix |
| **`… reduce the aligned seed when the modulus is an exact power of 2`** | **this PR** | |

**Only the last two commits are new here.** The other three are carried because this fix cannot be
expressed without them: the block it relocates *is* #172's fix, so the patch does not apply to a tree
without it. Merge #167, #185, #172 first (in that order) and rebasing this branch will drop them
automatically — they are the same patches, not copies.

`#167` is independent of the other two and may land at any point before this.

## The defect

`twopmmodq64/128/192/256` (`twopmodq.c:1192`, `twopmodq128.c:53`, `twopmodq192.c:53`,
`twopmodq256.c:88`) all share this shape, with W = 64/128/192/256:

```c
    if(p <= W) { ...direct computation...; return x; }   // early-out, tested against UNSHIFTED p
    nshift = trailz(q);
    if(nshift) { q >>= nshift;  p -= nshift; }           // strip 2^nshift from even q
    pshift = p - W;                                      // <-- wraps if p < W here
```

The early-out is tested before the power of 2 is stripped. So for

```
    W < p < W + trailz(q)        (which requires trailz(q) >= 2)
```

`p -= nshift` drops `p` below `W`, `pshift = p - W` wraps to a value near 2^W, and `leadb` and
`start_index` are extracted from that garbage — the powering loop then runs on the wrong exponent.

`p == W + nshift` is *not* affected: `pshift = 0`, so `leadb = 0` and `start_index = 0`, which the
powering loop handles correctly. That is why #186 sees p=67 and p=131 as "ok" sitting between two wrong
values.

**This is reachable from ordinary use, not just from a unit test.** At `Mlucas.c:5286-5300` (PRP-CF) the
modulus is `KNOWN_FACTORS[i] - 1`, which is **always even** because factors are odd, and the exponent is
the iteration count. `Mlucas.c:2424`, `Mlucas.c:6397` and `factor.c:1234` are further even/small-p call
sites.

## Why a uniform sweep will not find this

Two independent rare events have to coincide: `trailz(q) >= 2` (probability 1/4 for random q), and `p`
landing in a window of `trailz(q) - 1` consecutive values immediately above W (for p drawn from any
realistic range, essentially never). The harness therefore enumerates the window directly rather than
sampling it — `p` over `[W-4, W+40]`, `q = odd_part * 2^t` with `t` over `[0,40]`, N random odd parts per
cell.

That predicts a failure fraction of `sum_{t=2..40}(t-1) / (41*45) = 780/1845 = 42.28%`, and the unfixed
code measures **42.28%** — the sweep hits exactly the predicted cells and nothing else, which is what
makes it evidence rather than noise.

## Verification

Oracle is GMP `mpz_powm(2, p, q)`. "before" = the three dependency commits only.

| sweep | cases | before | after |
|---|---|---|---|
| targeted, `q = …F8` (trailz 3), p ∈ [W−2, W+5] | 32 | **8** | 0 |
| biased, p ∈ [W−4,W+40] × trailz ∈ [0,40], 200/cell | 1,476,000 | **624,000** (42.28%) | 0 |
| same, 1500/cell | 11,070,000 | **4,679,998** (42.28%) | 0 |
| broad, p ∈ [1,4W], random trailz ∈ [0,12) | 800,000 | **758** | 0 |
| degenerate, `q = 2^k` ∀k ∈ [1,W), p ∈ [1,W+8] | 122,792 | **59,677** | 0 |
| large p (up to 2^32), trailz ∈ [0,W−2] | 80,000 | 0 | 0 |

The last row is a pure regression check — with p ≫ W it structurally cannot reach the window, and
correctly reports 0 on both. Recorded so it is not mistaken for evidence.

The targeted sweep reproduces #186's table exactly, including both literal values quoted there
(`twopmmodq64(65, 0x7FFFFFFFFFFFFFF8)` → `0x200000` for a correct `0x20`; `twopmmodq128(129, ·)` →
`0x2000` for `0x20`), and confirms by measurement the p ∈ {193,194} and {257,258} windows that #186 had
only inferred. The p=68/132 rows from #186 are absent because #185 is in the base.

Per-width, on the 11.07M biased sweep: `twopmmodq64` 1,169,998 → 0; `128` 1,170,000 → 0;
`192` 1,170,000 → 0; `256` 1,170,000 → 0.

## Why there are two commits, and why both are needed

**Commit 1** is the #186 fix. For 128/192/256 it hoists the even-modulus right-justify shift **above** the
`p <= W` early-out, guarded by `p > W` so that `p >= nshift` still holds and the identity
`2^p mod q = 2^nshift * (2^(p-nshift) mod q')` remains valid, then restores the off-shifted power of 2 on
the early-out's return path. The guard is essential: hoisting unconditionally would underflow `p -= nshift`
for small p with a large `trailz(q)` (e.g. p=5, q=2^100). The 64-bit routine's direct path is a single
`(1 << p) % q`, so its early-out stays put and `p` is simply re-tested after the shift.

**Commit 2** fixes a second, adjacent defect, found while checking for regressions and not part of #186.
The `p <= W` direct path seeds `x = 2^j` with `j = W - leadz(q) - 1`, the largest power of 2 that is
`<= q`. That seed is strictly `< q` for every q **except an exact power of 2**, where `2^j == q`: the seed
is congruent to 0 but unreduced, and the doubling step only subtracts q when `x > qhalf`, so `x == q` is a
fixed point and the routine returns `q` where `0` is correct. One line per file:

```c
    if(CMPEQ128(x, q)) { x.d0 = x.d1 = 0ull; }
```

This is pre-existing — 59,677 of 122,792 `q = 2^k` cases are already wrong without either commit. **But
commit 1 routes `p ∈ (W, W+nshift)` into that path**, where the old `q' = 1` powering-loop route happened
to return the correct 0. Measured on the `q = 2^k` sweep:

| tree | wrong |
|---|---|
| dependencies only | 59,677 |
| **+ commit 1 only** | **64,177** — commit 1 newly breaks 4,500 cases |
| + both commits | **0** |

So commit 1 must not be merged without commit 2. In the other direction they are independent: commit 2
alone is a valid standalone fix for the pre-existing 59,677.

`twopmmodq64` is unaffected by commit 2 — `(1 << p) % q` is exact for any q. For p > 64 with `q = 2^k` it
trips its own `ASSERT(q > 1)` after the strip; that pre-existing hard abort is left alone.

## Other checks

* Full `makemake.sh avx2` build of the tree: clean, 0 errors.
* `./Mlucas -fft 1024 -iters 100 -radset 0` → `Res64: 95AFD7A5269F14F6`, and the built-in IMUL self-test
  (which exercises `imul_macro1.h`, i.e. #167) passes. Smoke test only — the changed routines are not on
  the FFT path.
* `twopmmodq64_q4` (`twopmodq.c:1339`) is **not** affected: it `ASSERT`s an odd modulus and has no nshift
  path at all.

Verified against a GMP oracle across 5 sweep modes.

### DIFF
--- a/src/imul_macro1.h
+++ b/src/imul_macro1.h
@@ -6402,7 +6402,8 @@ On Alpha, this needs a total of 12 MUL instructions and 42 ALU ops.
 		MUL_LOHI64(__x.d0,__x.d2, __e , __f );	/*   x0*x2 */\
 		MUL_LOHI64(__x.d1,__x.d2, __i , __j );	/*   x1*x2 */\
 		/* Add last pair (bottom layer) cross terms: */\
-		__b += __e;	__i += __f + (__b < __e);	/* f is a UMULH output, so f + CY cannot overflow */\
+		__b += __e;	__f += (__b < __e);	/* f is a UMULH output, so f + CY cannot overflow */\
+		__i += __f;	__j += (__i < __f);	/* ...but i + (f + CY) can, so ripple that carry into j */\
 		__hi.d2 = ((int64)__j < 0);\
 		/* Interleave left-shift of the 4-word result with squaring MULs: */\
 		/*** I wish I could say that interleaving the MULs and ALU stuff gives a speedup, but on x86_64 I get same speed if I do */\
@@ -6433,10 +6434,12 @@ On Alpha, this needs a total of 12 MUL instructions and 42 ALU ops.
 		/* Evaluate the cross-terms first, then right-shift those 1 place while we compute the w0-w5 terms: */\
 	"movq	%[__x0],%%rax	\n\t	mulq	%[__x1]	\n\t	movq	%%rax,%%r11	\n\t	movq	%%rdx,%%r12	\n\t"/* _a:_b = x0*x1 */\
 	"movq	%[__x0],%%rax	\n\t	mulq	%[__x2]	\n\t	movq	%%rax,%%rdi	\n\t	movq	%%rdx,%%r13	\n\t"/* _e:_f = x0*x2 */\
-	"movq	%[__x1],%%rax	\n\t	mulq	%[__x2]	\n\t		movq %%rdx,%%r14  \n\t	movq	%%rdx,%%r15	\n\t"/* _i:_j = x1*x2; leave _i in rax, make 2 copies of rdx */\
+	"movq	%[__x1],%%rax	\n\t	mulq	%[__x2]	\n\t		movq %%rdx,%%r14  \n\t"/* _i:_j = x1*x2; leave _i in rax, _j in r14 */\
 		/* Add last pair (bottom layer) cross terms: */\
 		"addq	%%rdi,%%r12		\n\t"/* _b += _e */\
 		"adcq	%%rax,%%r13		\n\t"/*	_i += _f + (_b < _e); _f is a UMULH output, so f + CY cannot overflow */\
+		"adcq	$0,%%r14		\n\t"/*	_j += (_i < _f); ...but _i + (_f + CY) can, so ripple that carry into _j */\
+		"movq	%%r14,%%r15		\n\t"/* copy of the now-final _j, for _w5 */\
 		"shrq	$63,%%r15		\n\t"/* _w5 = (_j >> 63) */\
 		/* Interleave left-shift of the 4-word result with squaring MULs: */\
 	"movq	%[__x0],%%rax	\n\t	mulq	%%rax	\n\t	movq	%%rax,%[__lo0] \n\t	movq	%%rdx,%%r10	\n\t"/* w0:w1 = x0^2 */\
--- a/src/twopmodq.c
+++ b/src/twopmodq.c
@@ -1207,9 +1207,16 @@ uint64 twopmmodq64(uint64 p, uint64 q)
 	// If get here, p > 64: set up for Montgomery-mul-based powering loop:
 	nshift = trailz64(q);
 	if(nshift) {
-		// p >= nshift guaranteed here:
+		// p >= nshift guaranteed here, since p > 64 > nshift:
 		q >>= nshift; p -= nshift;	// Right-shift dividend by (nshift) bits; for 2^p this means subtracting nshift from p
 		if(debug) printf("Removed power-of-2 from q: q' = (q >> %u) = %" PRIu64 "\n",nshift,q);
+		// The (p <= 64) early-out above was taken against the *unshifted* exponent, so for
+		// p in (64, 64+nshift) the subtraction just dropped p below 64 and the (pshift = p - 64)
+		// computed below would wrap. Redo the direct computation, now against the odd part q',
+		// and restore the off-shifted power of 2 on the way out.  p = 64 needs no special case:
+		// it gives pshift = 0, hence leadb = 0 and start_index = 0, which the main path handles.
+		if(p < 64)
+			return ((1ull << p) % q) << nshift;
 	}
 	qhalf  = q>>1;	/* = (q-1)/2, since q odd. */
 	// Extract leftmost 7 bits of (p - 64); if > 64, use leftmost 6 instead:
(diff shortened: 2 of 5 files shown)


MarkRose:
Rebased onto current `main` to clear the merge conflict. No conflicts to resolve in the end — git dropped two commits that have since landed by other routes (`SQR_LOHI192: propagate the dropped cross-term carry` and `recompute qhalf after the even-modulus right-justify shift`, the latter via #172), leaving the three commits that are still this PR's own. Rebuilt and re-tested after: `-s tt` rc=0, 19 residues.

PR #216, opened 2026-07-30
Make the self-test capable of failing, and cover the FFT lengths it never reached

MarkRose:
The self-test has a good correctness check and **throws its verdict away**. This makes it capable of failing, and extends it to the FFT lengths it never reached.

## The core defect

`Mlucas.c:4853-4855`:

```c
DONE:
	fprintf(stderr, "\n  Done ...\n\n");
	return(0);
```

`-s` and `-fft/-radset` **always exit 0**. Measured on a stock `nosimd` build of `418f365`:

```
$ ./Mlucas -s tt ; echo "exit=$?"
WARNING: 1 of 3 radix-sets at FFT length 4 K passed - skipping it. PLEASE CHECK YOUR BUILD OPTIONS.
WARNING: 1 of 3 radix-sets at FFT length 5 K passed - skipping it. PLEASE CHECK YOUR BUILD OPTIONS.
...
exit=0
```

And `ci.yml`'s four self-test steps gate on that status — three of them additionally pipe into `tee`, so even a corrected status would be swallowed by the pipeline. **No self-test regression could ever have turned CI red.**

## Four more, underneath

2. **The new-data consensus anchor is `if(!new_res.sh0)`.** A first-radset `Res64 == 0` leaves the anchor falsy, so *every* radix set is scored as passing whatever it computed. On pristine main, `-m 18776473 -fft 992 -iters 100` prints `Res64: 0000000000000000. AvgMaxErr = 0.000000874` and writes that zero into `mlucas.cfg` as a certified "2 of 2 consensus". Exposure: 42 of 136 `MvecPRP` rows, the 10000-iter column of 17 `MersVec` rows, and every off-table `-fft`.
3. **`resFlag = 0` on a match erases an earlier Res64 mismatch.**
4. **No zero-residue rejection when there is no reference** — the zero is *adopted* as the reference.
5. **The "length not in the table" test is dead code**: `i != MvecPtr[j].fftLength` is evaluated at `j == numTest`, against the row just written a few dozen lines earlier, so it is never true. `-fft 992 -iters 100` refused to run at all.

Note that `fracmax`/`AvgMaxErr` cannot substitute for any of this: it measures *distance from the nearest integer*, and **zero is an integer**, so an all-zero residue reports perfect numerical health.

## Coverage, measured rather than assumed

The "56 of 603" figure in circulation compares two different populations — 56 is the *nosimd* Teensy count, 603 the *SIMD* total. Enumerated properly via `get_fft_radices()` in a harness linked against the real objects: **153 legal lengths / 603 (fftlen, radix_set) pairs** under SIMD, 708 under nosimd. `-s a` covers 454/603 (75%); all tiers together 564/603 (93.5%). Per tier on avx2: tt 26, t 93, s 127, m 145, l 63, h 67, e 43 — reproduced exactly by the new summary line.

**The gap that matters is not the 6.5%.** The 17 uncovered lengths are exactly the 31·2ᵏ and 63·2ᵏ families — the only lengths at which leading radices 63, 992, 1008 and 4032 are ever used, and therefore precisely where every silent-wrong-residue defect found in this campaign lives. All 27 radix sets there fail today in plain AVX2.

## What this changes

* **An exit contract:** 1 = wrong answer, 2 = length untested, 0 = clean.
* **A new `-s c` coverage sweep** over the uncovered lengths. It needs **no new reference data**: `Res64` depends only on (exponent, iters), so each length borrows the nearest tabulated row. Verified — p=17551099 gives `AB1180428ED65EE0` at 896K and at all 12 radsets of 1024K.
* The anchor fix, zero-residue rejection, the `resFlag` fix, the dead-branch fix, an always-printed summary, and CI gating via `${PIPESTATUS[0]}` so the `tee` no longer swallows the status. CI fails on status 1 only, since status 2 fires on today's 1K/2K.

**Plain cross-radix-set agreement was considered and measured insufficient** — at 992K both radsets share leading radix 992 and both return `0000000000000000`, so they agree with each other and are both wrong.

## Validated failing-first

A test that cannot fail is the subject of this PR, so each check was demonstrated to fail before being demonstrated to pass:

| case | before | after |
|---|---|---|
| radix992 (nosimd), broken | — | 8 wrong answers, `Res64 Error`, **exit 1** |
| radix992 with #179 | — | 992/1984/3968/7936/15872 all pass, 0 wrong |
| radix16 zero-cosine (AVX-512/SDE, `-fft 8 -radset 1`) | pristine main **exit 0** on `ERR_CARRY` | **exit 2**; with #214's fix, correct Res64 and **exit 0** |
| `-s t` / `-s s` negative controls | — | 93/93 and 127/127, exit 0 |

Independently reproduced here before pushing: stock main `-s tt` exits **0** on the 4K/5K/7K/9K/10K warnings above; this branch exits **1** on the identical set.

Runtimes (avx2, 4 threads): tt 2.2 s, t 15 s, s 44 s, m 1493 s, and `-s c` ≈2–4 min once those lengths work.

## Known blockers for a stricter gate, not fixed here

1K and 2K are unusable in a stock AVX2 build of main — 1K gives ROE ≈0.4993 on iteration 1, and 2K radset 0 hits `ERR_ASSERT` "radix < 16 does not support shifted residues" against the self-test's own random `RES_SHIFT`. That is why CI gates on status 1 rather than `!= 0`. AVX-512 under SDE loses 8 of 15 teensy lengths. And `-s` never exercises Fermat-mod at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,12 +72,17 @@ jobs:
         # a pipeline; between them no self-test regression could turn this job red. Gate on 1 for
         # now - status 2 currently fires on a stock build at FFT lengths 1K and 2K - and tighten
         # to (( rc == 0 )) once those are fixed.
+        # This job runs with pipefail set, so a self-test exiting 2 would propagate out of the pipe
+        # into tee and, with set -e, end the step before the PIPESTATUS check below classifies it.
+        # Turn pipefail off for the loop and take the self-test's own status from PIPESTATUS[0]:
+        set +o pipefail
         rc=0
         for s in tt t s m l; do
             time ./Mlucas -s $s -cpu "0:$(( $(nproc --all) - 1 ))" |& tee -a test.log
             st=${PIPESTATUS[0]}
             if (( st == 1 )); then rc=1; fi
         done
+        set -o pipefail
         (( rc == 0 ))
         if [[ $HOSTTYPE == aarch64 ]]; then
             cd ../obj_NOSIMD
@@ -410,12 +415,17 @@ jobs:
         # a pipeline; between them no self-test regression could turn this job red. Gate on 1 for
         # now - status 2 currently fires on a stock build at FFT lengths 1K and 2K - and tighten
         # to (( rc == 0 )) once those are fixed.
+        # This job runs with pipefail set, so a self-test exiting 2 would propagate out of the pipe
+        # into tee and, with set -e, end the step before the PIPESTATUS check below classifies it.
+        # Turn pipefail off for the loop and take the self-test's own status from PIPESTATUS[0]:
+        set +o pipefail
         rc=0
         for s in tt t s m l; do
             time ./Mlucas -s $s -cpu "0:$(( $(sysctl -n hw.ncpu) - 1 ))" 2>&1 | tee -a test.log
             st=${PIPESTATUS[0]}
             if (( st == 1 )); then rc=1; fi
         done
+        set -o pipefail
         (( rc == 0 ))
         if (( $(sysctl -n hw.optional.neon) )); then
             cd ../obj_NOSIMD
@@ -613,12 +623,17 @@ jobs:
         # a pipeline; between them no self-test regression could turn this job red. Gate on 1 for
         # now - status 2 currently fires on a stock build at FFT lengths 1K and 2K - and tighten
         # to (( rc == 0 )) once those are fixed.
+        # This job runs with pipefail set, so a self-test exiting 2 would propagate out of the pipe
+        # into tee and, with set -e, end the step before the PIPESTATUS check below classifies it.
+        # Turn pipefail off for the loop and take the self-test's own status from PIPESTATUS[0]:
+        set +o pipefail
         rc=0
         for s in tt t s m l; do
             time ./Mlucas -s $s -cpu "0:$(( NUMBER_OF_PROCESSORS - 1 ))" |& tee -a test.log
             st=${PIPESTATUS[0]}
             if (( st == 1 )); then rc=1; fi
         done
+        set -o pipefail
         (( rc == 0 ))
         if [[ $HOSTTYPE == aarch64 ]]; then
             cd ../obj_NOSIMD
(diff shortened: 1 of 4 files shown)


MarkRose:
Rebased onto current `main` to clear the merge conflict.

Nine conflicts, all in `ci.yml`, because #103 reworked every self-test block this PR gates. Rather than hand-merge two divergent versions nine times, I took `main`'s job structure wholesale and re-applied the gating on top of it, so the CI half of this PR is now written against the loops that actually exist — including the `-prp` loops and the per-word native Mfactor runs that did not exist when this branch was written.

## The gate is now checked end to end

Worth showing, because the first control I ran did not actually test anything.

```
stock build      ./Mlucas -s tt -> exit 2   ("2 FFT lengths had too few usable radix sets")
corrupted build  ./Mlucas -s tt -> exit 1   ("17 cases returned a wrong answer")
```

so the job stays green on 2 and goes red on 1, which is what the gate is for.

Getting there took two false negatives worth recording. Corrupting the 1K row proved nothing: that case halts on a roundoff error in an AVX2 build before any residue is compared. Corrupting nineteen larger rows then *also* proved nothing, because `makemake.sh` refuses to overwrite an existing `obj_avx2/Mlucas` and I had re-run the stale binary. Only after deleting the object directory did the corruption show up. Both failures looked exactly like "the gate works" from the outside.

Build OK, and the stock `-s tt` summary is unchanged from `main` apart from the new pass/fail block.

MarkRose:
Two follow-up commits, both fixing defects in this PR's own gating.

**The status was never examined.** The pipeline form

```bash
time ./Mlucas -s $s -cpu "..." |& tee -a test.log
st=${PIPESTATUS[0]}
```

runs under `set -e` with `pipefail`, so a nonzero status ends the step on the pipeline itself and the `st=` line never executes. The whole point of capturing the status was to tolerate 2; instead every job that hit it died on the first tier of the ladder and ran none of the others. The status is now captured in a `|| st=${PIPESTATUS[0]}` tail, which `set -e` does not apply to.

**The gate was too narrow.** `if (( st == 1 ))` passes everything else, including the statuses the self-test never returns. A SIGSEGV part-way through `-s m` (139) or an abort (134) counted as a clean ladder - I have one of each in the current logs. It is now `st != 0 && st != 2`, so 2 stays tolerated for the stock build's 1K/2K rejections and nothing else does.

**`config-fermat.sh`.** Callers invoke it as `bash -e -o pipefail -- ./config-fermat.sh`, so now that Mlucas returns a nonzero status the first affected length ends the script. On ARM that is F18 at 15K, and F19 through F29 were never configured and never tested - the run simply looked finished. It now carries on past status 2 and exits nonzero at the end for anything else.

PR #218, opened 2026-07-30
Fermat-mod Gerbicz check: make it work with a nonzero residue shift

### LINKED ISSUE #99: Allow using a random shift with FFT lengths larger than 256M and when testing Fermat numbers

MarkRose:
A Pépin test picks a random nonzero residue shift unless the user passes `-shift 0`, and the Gerbicz check is on by default — but the two did not work together. The run failed its first G-check (iteration 1000000 by default), the assignment was dropped from the workfile with no result, and any attempt to resume from the leftover checkpoint aborted on

```c
ASSERT(RES_SHIFT == 0ull, "Shifted residues unsupported for Pépin test with Gerbicz check!")
```

Worth noting that assertion sat in the **resume** branch of `if(ilo == 0)`, so it never prevented the combination from being *started* — it only turned "wrong check" into "cannot resume". A from-scratch Pépin run picked a random nonzero shift and proceeded straight past it.

## Three defects

**1. The `prp_mult` gate applied the Pépin random-bit doubling to every array the carry routines process.** That doubling is the multiplicative partner of the `shift = 2*shift + randbit` update, so it belongs to the shift-carrying main residue chain alone. Applying it to the G-check product update and to the check's L-fold squaring chain — both called with `update_shift = FALSE` — multiplied those by unaccounted-for powers of 2. Fixed with a `FERMAT_RANDBIT_MULT` global, set from `update_shift` on entry to `fermat_mod_square()` and required in the Fermat clause of the gate.

**2. The correction factor `2^{s_L}` is the Mersenne answer.** Writing `x_i` for the shifted residue after *i* squarings, `u_i = 3^(2^i) (mod Fm)` for the true one, `s_i` for `RES_SHIFT` and `q_i = [s_i >= p/2]` for `RES_SIGN`:

```
x_i = (-1)^q_[i-1] . u_i . 2^s_i
```

so the check product `b = 3 . prod_k x_[kL]` satisfies

```
b = 3 . d^(2^L) . 2^E,    E = sum_k (s_[kL] + q_[kL-1].p)   (mod 2p)
```

since `2^(2^L) == 1 (mod Fm)` for `L >= m+1` kills the powers of 2 the d-chain squaring would otherwise carry. In the **Mersenne** case all `q = 0` and `s_[kL] = 2^L . s_[(k-1)L] (mod p)`, so that sum telescopes to the single term `s_L` — which is what the code computed, **for both moduli**. The Fermat-mod shift update is "mod-double *and add a random bit*", so it does not telescope, and `E` genuinely changes from one check interval to the next. `GCHECK_SHIFT` now maintains that running `(mod 2p)` sum, folding sign flips in via `-1 == 2^p (mod Fm)`, and splits it at check time into a negacyclic rotation plus an explicit negation when it lands in the upper half.

This also retires the `filegrep(STATFILE, "ITERS_BETWEEN_GCHECK_UPDATES", ...)` workaround that recovered the old (wrong) correction by parsing the logfile, and the logfile line that existed only to feed it. That hack left `itmp64` unset if the grep missed.

**3. Restart safety.** The G-check product is not a shifted residue, but `convert_res_FP_bytewise` / `convert_res_bytewise_FP` applied `RES_SHIFT` and `RES_SIGN` to it from the globals. For Fermat-mod the two are **not** mutually inverse — write rotates by `(p - shift)` and conditionally negates, read rotates by `shift`, composing to multiplication by `2^p == -1` — so `b[]` could come back from a savefile negated. The shift is now suppressed for the `b[]` and `d[]` conversions in the Fermat-mod case only. Mersenne-mod keeps its current behaviour: there the rotation is cyclic and cancels between the two sides of the comparison, and changing it would alter every in-progress PRP savefile.

## Savefile compatibility

`GCHECK_SHIFT` is already an 8-byte field at a fixed offset, and every existing Fermat savefile has `RES_SHIFT == 0` and `GCHECK_SHIFT == 0`, for which an accumulator starting from 0 stays 0. The Mersenne-mod semantics of the field are untouched.

## Verification (AVX2, gcc 16.1.1)

* **F16** at scaled intervals (L = 100, check every 10000), shifts 0 / 1 / 12345 / 33333 / 65432 / 65535: **6 of 6 G-checks pass in every case**, and the final `Res64,Res35m1,Res36m1 = 40ABB0C5BFF05CB5,173595305,65390296136` matches an independent GMP Pépin computation (and a separate pure-Python one). Before this change every nonzero shift failed the first check; shift 0 is bit-identical to before.
* **F20 at the production intervals** (L = 1000, check at 1000000), shift 654321: `At iteration 1000000, shift = 630647: Gerbicz check passed`, same final residue as the shift-0 run, matching a 53-minute GMP run of the same exponent. Before: `Gerbicz check iteration 1000000 failed!` and the assignment abandoned.
* **Restart coverage**: kill/resume across a checkpoint at iteration 24000 with shift 12345, and a forced G-check failure at 30000 (rollback to the `.G` savefile, which mod-doubles `RES_SHIFT`). Both recover and finish with the correct residue; both abort on the assertion before this change.
* **No-regression**: `-s tt` gives 13K=`E3BC90B0E652C7C0` 14K=`DB8E39C67F8CCA0A` 15K=`B3C5A1C03E26BB17`; a Mersenne 3-PRP of M86243 with G-checks produces **byte-identical** `p86243` and `p86243.G` savefiles and an identical statfile (bar the deleted log line); Fermat self-tests F14–F17 over several FFT lengths are unchanged at shift 0 and give the same residues at shift 12345. Compiles clean at nosimd/avx/avx2/avx512.

## Relationship to issue #99

Addresses the Fermat-number half of #99: a Pépin test with the Gerbicz check no longer requires `-shift 0`. The FFT-length half of that issue (moduli larger than 2^32 bits) is not touched here, so this deliberately does **not** say "fixes #99".

**This must land together with #190.** Measured on `main` + this PR alone: the Gerbicz check passes, but the run still reports a genuine Fermat prime as *composite* at 5 of 6 nonzero shifts and still returns the +2-wrong residue — strictly worse than the current behaviour of refusing to run. `main` + #190 alone still fails the first G-check. Together: 13/13 correct prime verdicts across shifts spanning `RES_SIGN` flips, against 1/13 on `main`.

One limitation neither PR removes: a nonzero shift still fails at iteration 0 when the leading FFT radix is < 16, which the stock `fermat.cfg` selects at 4K (`8 16 16`). That is #119 / PR #146.

## Relationship to other PRs

* **Supersedes #122**, which addressed the same failure by forcing `RES_SHIFT = 0` for this combination. This removes the restriction instead, so nonzero shifts work rather than being disallowed. #122 has been closed.
* No overlapping hunks with **#190**. This change actually *removes* two `mi64_shlc(sign_flip=1)` calls from the G-check path, reducing exposure to the bug #190 fixes.

## Deliberately left alone

* `mi64_shlc_bits_align` is still cyclic-only. With the G-check path no longer calling it in the negacyclic case that is a diagnostic gap rather than a correctness one, but it is still a gap.
* Fermat PRP-CF runs hit a pre-existing, unrelated `Suyama_CF_PRP` assertion ("requires p-1 mod-squarings!") that reproduces on unmodified main. Separate problem, not addressed here.

### DIFF
--- a/docs/Fermat-testing.md
+++ b/docs/Fermat-testing.md
@@ -96,15 +96,14 @@ to be composite. However, the character of their cofactors is unknown so a Pépi
 necessary prerequisite for testing them; and as for _F_<sub>33</sub>, this is yet to be 
 tested for primality.
 
-The Pépin test uses Gerbicz error checking (GEC) to assure the reliability of the computation, however 
-residue shifting is not implemented for Fermat modular squaring with GEC. Thus when invoking Mlucas the 
-flag `-shift 0` _must_ be added to the Mlucas command line:
-```
-./Mlucas -shift 0
-```
-If non-zero residue shifting is used, the Pépin test will not be able to progress beyond the millionth 
-iteration (the default interval for GEC) as error checking will be unable to confirm whether or not the 
-computation is correct.
+The Pépin test uses Gerbicz error checking (GEC) to assure the reliability of the computation. Up to and 
+including v21.0.2, GEC and a nonzero residue shift were incompatible: the check applied a correction 
+factor that is only valid for the Mersenne-mod pure-shift-doubling case, so a Pépin test run with the 
+default (randomly chosen, nonzero) shift failed its first GEC at the millionth iteration and the 
+assignment was abandoned; `-shift 0` therefore had to be added to the command line. That restriction has 
+been lifted - the Fermat-mod GEC now maintains its own correction accumulator - and `-shift 0` is no 
+longer needed for GEC. It is still required for FFT lengths whose leading radix is less than 16 (see the 
+small-FFT note below) and for p-1 runs.
 
 ## Cofactor compositeness testing: Suyama’s test
 Suyama’s test is a Fermat probable primality test on the cofactor of a Fermat number. As a prerequisite, 
--- a/src/Mdata.h
+++ b/src/Mdata.h
@@ -262,6 +262,12 @@ extern uint64 PMAX;	/* maximum exponent allowed */
 extern uint64 RES_SHIFT, GCHECK_SHIFT;
 // Feb 2020: added a uint32 to keep track of the shifted-residue sign, needed for rotated residue Fermat-mod arithmetic:
 extern uint32 RES_SIGN;
+/* Fermat-mod Pépin test only: the random-bit residue-doubling that accompanies the 'shift = 2*shift + randbit'
+update is a property of the *shift-carrying main residue chain*, not of the modulus. Set FALSE by
+fermat_mod_square() while it is squaring/multiplying one of the Gerbicz-check auxiliary arrays (which do not
+carry a shift), so that the carry routines skip the doubling for those. Cf. the prp_mult gate in the
+radix*_ditN_cy_dif1() routines and the update_shift arg of fermat_mod_square(). */
+extern int FERMAT_RANDBIT_MULT;
 extern uint64 *BIGWORD_BITMAP;	/* Needed for fast how-many-residue-bits-up-to-this-array-word lookup, which the carry routines
 								use to figure out where to inject the -2 in a rotated-residue LL test using Crandall/Fagin IBDWT.
 								A 1-bit means a bigword (p/n+1 bits); 0 means a smallword (p/n bits), where p/n is integer-div. */
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -147,6 +147,7 @@ const char *err_code[ERR_MAX] = {
 uint64 RES_SHIFT = 0xFFFFFFFFFFFFFFFFull;	// 0 is a valid value here, so init to UINT64_MAX, which value is treated as "uninited"
 uint64 GCHECK_SHIFT = 0ull;
 uint32 RES_SIGN = 0;	// Feb 2020: uint32 to keep track of shifted-residue sign flips, needed for rotated residue Fermat-mod arithmetic.
+int FERMAT_RANDBIT_MULT = TRUE;	// Fermat-mod Pépin: apply the random-bit residue-doubling? Only for the shift-carrying main chain; cf. Mdata.h.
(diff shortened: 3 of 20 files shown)


MarkRose:
Added the `FERMAT_RANDBIT_MULT` gate for `radix63_ditN_cy_dif1.c`, which this PR's original sweep could not cover: radix63 had no Fermat arm in its `prp_mult` condition at all, so there was nothing to gate. #232 adds that arm against `main`; the commit here writes the same line already gated, so the file ends up consistent with its 15 siblings whichever lands first.

Ordering is recorded as Chain F in #224: **#232 first, then this PR.** If they land the other way round the two edits collide on one line and the resolution is this version, which is the superset.

Verified on this branch alone (nosimd): F24 at FFT length 1008K goes from *2 of 3 radix-sets passed* with radix63 dissenting from the cross-radix consensus, to **3 of 3 with no disagreement**.

PR #219, opened 2026-07-30
Lint inline-asm blocks for registers used but not declared, and fix util.c %rdx

MarkRose:
One commit: a standalone lint for a defect class, plus the one remaining live instance of it. **Not wired into CI** — it is a script you run when you want it.

## The defect class

Inline-asm clobber lists that do not describe what the asm writes. The compiler parks a live value in the undeclared register, the asm destroys it, and you get a wrong answer or a SIGSEGV. Five instances have turned up in this tree, **four of which first presented as compiler bugs**, and two of which reached a published workaround before being retracted:

| where | registers | first diagnosed as | actually |
|---|---|---|---|
| radix240 `SSE2_RADIX_05_DFT_0TWIDDLE` | `zmm8`, `zmm9` | — | #203 |
| radix1024 `carry_gcc64.h` | `zmm18`, `zmm19` + `k1`–`k4` | "gcc-16 LTO drops a clobber" (retracted) | #68 |
| radix32 `SSE2_RADIX32_WRAPPER_DIT` | `r11`, `r12`, `r13` | "clang-specific memory corruption" | #68 |
| radix320 | same macro as radix240 | "GCC 11 auto-vectorizer miscompile" | #203 |
| `util.c:3112` | `%rdx` | — | **fixed here** |

They look like compiler bugs because only *which* registers the allocator picks varies by version. The signature is consistent: `-O0` clean, one compiler affected, disabling the vectorizer doesn't help, `-O1` fails too.

**Sanitizers are structurally blind to it.** A missing clobber is not UB in the sanitizer's sense — there is no C-level operation to instrument — and the instrumentation perturbs register allocation enough that the symptom disappears. Measured: UBSan on the unfixed radix240 source, with 731 `__ubsan_*` call sites inside the faulting function, emitted zero diagnostics and returned the *correct* residue.

## The `util.c:3112` fix, and it is latent

The asm writes `%rdx` with no clobber and no operand binding (all seven operands are `"m"`). Two honest caveats, because "latent" is the accurate word and I would rather say it than imply breakage:

* `test_mul50x50()` **is** compiled into every AVX2/AVX-512 build — `platform.h:506` defines `USE_FMADD` whenever `USE_AVX2` is set — but its only call site is inside a `/* */` comment, so it never runs.
* gcc 16.1.1 at `-O3` emits **byte-identical code** with and without the clobber, so there is no current miscompile even if it were called.

Fixed because the declaration should be correct regardless.

## The lint

`asm-clobber-lint.py` plus `asm-clobber-baseline.txt`. Run from the repo root:

```
python3 asm-clobber-lint.py              # exits nonzero on any finding not in the baseline
python3 asm-clobber-lint.py --regenerate # rewrite the baseline
python3 asm-clobber-lint.py --self-test  # 16 parser cases with paired controls
```

Two passes, because neither alone suffices:

* **Pass A** parses `src/*.c` and `src/*.h` as raw text, so a finding inside a macro definition is reported at the macro, not at whichever translation unit happened to include it.
* **Pass B** preprocesses every `.c` once per ISA mode (`nosimd`/`sse2`/`avx`/`avx2`/`avx512`, flags taken from `makemake.sh`) and records which asm blocks are actually live, so no `#ifdef` branch has to be guessed at. On this tree it suppresses 65 findings in `*_gcc32.h` and other unreachable paths.

**False positives are why a lint like this gets switched off**, so operand constraints are parsed in full: an asm with `"=d" (x)` or `"c" (n)` legitimately uses `%rdx` or `%rcx` without declaring it, and `%0` / `%1` / `%[name]` are operand references rather than registers. Register aliases are canonicalised — `%eax` and `%al` count as `rax`, `%r11d` as `r11`, and `%zmm18` is satisfied by an `"xmm18"` clobber.

A full sweep takes ~35 s, has zero preprocessor failures, and gcc and clang produce identical output.

## Validation

All three fix PRs are still open, so **`main` itself is the pre-fix source** — a convenient natural experiment:

| tree | findings |
|---|---|
| `main` | 29 |
| + #203 + #68 | 1 |
| + this PR's `util.c` fix | **0** |

Every one of the five known instances is reported at the right file and line, and none is reported once its fix is applied. Zero false positives. The check was then deliberately broken five ways — revert the fix, bump a baseline count, drop a baseline row, break a TU — with the exit code confirmed each time.

**Worth noting for reviewers of #68: it accounts for 27 of those 29 findings**, far more than its two headline sites.

## The tree-wide sweep result is negative, and that is the useful part

No new *live* instance exists. #203 + #68 + this fix close the entire live x86 surface of the class. The lint's value from here is preventing regressions, not clearing a backlog — which is also why it is a script rather than a CI job.


MarkRose:
Rebased onto current `main` (108 commits) and brought up to date, folding in the CI wiring from #255, which I opened by mistake as a duplicate of this and have now closed.

Three things changed:

**The baseline is gone rather than empty.** It recorded 24 accepted findings when this branch was written; #68 has since landed and fixed all of them. Regenerating against current `main` gives zero rows, so the file is deleted. With no baseline present, any finding fails the job — a stronger gate than carrying an empty file.

**The lint is now wired into CI.** A job that runs the lint's own self-test first, then the lint. No `continue-on-error`, and no pipe into `tee` that would swallow the exit status.

**The `util.c` %rdx fix is already on `main`**, so the rebase absorbed it and nothing of it remains here. That fix originally came out of this lint, which is a reasonable advert for it.

## Current state

398 inline-asm blocks across 45 files, **0 findings**, self-test 16/16.

A zero-finding run is only meaningful if the check can fail, so: removing the `%rdx` clobber from `util.c` again makes it exit 1 with

```
util.c:3141: inline asm uses rax,rbx,rcx,rdx,xmm0,xmm1,xmm2,xmm3,xmm12,xmm13,xmm14,xmm15
             but neither binds nor clobbers rdx  [avx2,avx512]
```

— correct file, correct register, correct modes.

Worth restating why this class needs a static check rather than a sanitizer: an asm template that touches a hard register it neither binds nor clobbers leaves the compiler free to keep a live value there, and the asm destroys it. Five instances have been found in this tree and **four were first diagnosed as gcc or clang codegen defects** before being traced back. ASan cannot see it — there is no C-level operation to instrument, and the instrumentation perturbs register allocation enough that the symptom often disappears.

PR #221, opened 2026-07-30
CI: extend the ASan matrix to AVX2 and AVX-512 under SDE, with Res64 assertions

MarkRose:
> **No longer stacked.** #68 merged on 2026-08-18, so its commits have dropped out of this branch; what remains is two commits of this PR's own.

Extends the ASan CI matrix from `nosimd` to **avx2 natively and avx512 under Intel SDE**, for both gcc and clang.

## Three requirements, each learned the expensive way

**1. It asserts Res64 — it does not grep for ASan reports.** Every symptom in this defect class produced **zero ASan reports**; a register clobber leaves no bad memory access to detect. A report-grepping job would have passed the entire broken matrix while it silently computed wrong residues.

The job pins residues, pins exponents, **and asserts the selected radices**, so `get_fft_radices()` cannot silently move a case off the path being tested. The ASan-report grep is still there, but explicitly last. Reference values were re-derived here from clang/avx2+ASan and confirmed in five further configurations including a **non-sanitised** build.

**2. A trailing-radix-32 case is present and proven load-bearing.** The discriminator for this whole class is the *trailing* radix, not the leading one. 32K radsets 0–3 give trailing 16/32/32/16 at one exponent and one residue: on unfixed `main`, clang/avx512 fails exactly rs1 and rs2 while both trailing-16 controls pass. A matrix built from default radix sets would miss it entirely.

**3. No `-fsanitize-recover=address`.** That flag alone makes the broken build produce *correct* results — it actively hides the defect class this matrix exists to catch.

## The probe was verified against unfixed code

8 script runs / 40 cases:

| tree | result |
|---|---|
| unfixed clang/avx512 | exit 1, 3 FAIL |
| unfixed gcc/avx512 | exit 1, 1 FAIL (128K) |
| both, fixed | 5/5 pass |

**Zero ASan reports anywhere**, confirming requirement 1 is not theoretical. And **two of the three clang failures exited 0** — so exit status alone catches only 1 of 3, which is why the residue assertion is the primary check.

## The gcc/ASan build break is fixed, not skipped

New measurement: gcc + ASan fails at **avx2 as well as avx512** (earlier work only tested avx512), on gcc 13.3 — i.e. `ubuntu-latest` — and 16.1.1. The gcc leg carries `--param asan-stack=0` plus a `::warning` annotation naming exactly what that costs. Verified that `asan-stack=0` does **not** hide the clobber defect it needs to catch.

The gcc leg is not dropped as redundant: the `carry_gcc64.h` opmask/high-vector defect is gcc-only, so a clang-only matrix would not cover it.

**The workaround is probably temporary, and #71 is why.** `'asm' operand has impossible constraints` is the *over-declared* form of the same clobber-list problem — an asm naming more registers than it needs, which only turns fatal once something else raises register pressure, and ASan's stack instrumentation is exactly such a something. #71 fixes that class and touches `src/sse2_macro_gcc64.h`, the file identified here as the source of the 14-GPR clobber lists. Once #71 lands it is worth re-testing whether gcc can build the SIMD kernels with full stack instrumentation, and if so dropping `--param asan-stack=0` and its `::warning` annotation. Not doing that here, because #71 is not merged and this job should be honest about what it covers today.

## Cost and hygiene

2 jobs, ~19 runner-minutes, 2 queue slots. SDE is pinned by URL + sha256. The logic lives in `.github/scripts/asan-simd-check.sh` (156 lines, ShellCheck-clean with the repo's option set) rather than inline YAML, specifically so it can be run **and falsified** locally.

No `continue-on-error`, no `tee`, anywhere.

The `ci.yml` change is a **pure insertion** — one hunk, 99 lines, zero existing lines touched — sitting between `ASan-Linux` and `TSan-Linux`, so it should reconcile trivially with #219's concurrent `ci.yml` edit.

## Two findings from this work, not addressed here

* **`ASan-Linux`'s gcc leg is dead on `main` right now.** It invokes `makemake.sh` with no mode argument, so it builds AVX2/AVX-512 rather than `nosimd`, and gcc+ASan cannot compile that. Build break plus `continue-on-error` equals fake success — the same pattern that previously hid #156 and #157, recurring. One-line fix, deliberately left out of this PR to keep it reviewable.
* **clang cannot build AVX-512 kernels with `address,undefined` at `-O3`**: `mi64.c:7353: symbol 'loop2d' is already defined` — the asm-label-collision class that #82 addresses. That is why this job is ASan-only.

## On #68's reported `CPU_ZERO` failure — it does not reproduce

Worth recording since it was cited as a blocker. `main` already fixed it: `makemake.sh:541` emits `CPPFLAGS ?= -D_GNU_SOURCE` and `platform.h` gained an `#ifndef` guard. #68 touches neither, nor `threadpool.c`, and the rebased tree built clean ten times. The failure appears only if `threadpool.c` is compiled by hand, or if `CPPFLAGS` is **exported**, which knocks out makemake's `?=` default — a real CI foot-gun, now commented in the script.

---

**Update (2026-09-10):** rebased onto current `main`, and the gcc leg's `--param asan-stack=0` workaround is gone - both legs now build with full instrumentation.

That flag was here because ASan's stack instrumentation needs a register to reserve, and the over-declared 14-GPR clobber lists in `sse2_macro_gcc64.h` left it none, so the build died with `'asm' operand has impossible constraints`. The cost was a real, named blind spot on the gcc leg: no stack-buffer-overflow, use-after-return or use-after-scope detection.

Those clobber lists have since been corrected on `main` - `3c9dbc7` ("Fix 'asm operand has impossible constraints' in hand-tuned SIMD/int asm") and `5eb3420` ("Free a GPR in the AVX-512 SSE2_RADIX16_DIF_0TWIDDLE for newer GCC") - so the workaround no longer buys anything. Measured on current `main` with this job's exact CFLAGS and no `--param asan-stack=0`, gcc 16.2.1:

| build mode | result | `impossible constraints` errors |
| --- | --- | --- |
| `makemake.sh avx2` | rc=0 | **0** |
| `makemake.sh avx512` | rc=0 | **0** |

against 4 and 6 respectively before the clobber fixes.

This supersedes #227, which existed only to remove the workaround afterwards; folding it in here means the flag is never added in the first place, and there is one PR to review rather than two. A short comment is kept at the site recording why it used to be needed.

### DIFF
--- a/.github/scripts/asan-simd-check.sh
+++ b/.github/scripts/asan-simd-check.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+#
+# Build Mlucas for one SIMD mode under AddressSanitizer and assert that it computes the
+# *right answers* - not merely that it emits no sanitizer report.
+#
+# Usage:  .github/scripts/asan-simd-check.sh <avx2|avx512> [runner...]
+#
+#   <mode>       SIMD build mode, passed straight through to makemake.sh.
+#   [runner...]  Optional launcher prefixed to the binary, e.g. "$HOME/sde/sde64 -skx --".
+#                Needed for avx512 because GitHub's Azure runner SKUs are inconsistent
+#                about AVX-512 support; Intel SDE removes that dependency entirely.
+#
+# Reads CC / CFLAGS / LDFLAGS from the environment (makemake.sh's generated Makefile uses
+# "CFLAGS ?=", so an exported CFLAGS wins outright). Do NOT export CPPFLAGS: makemake.sh
+# sets "CPPFLAGS ?= -D_GNU_SOURCE ...", and without -D_GNU_SOURCE threadpool.c loses
+# CPU_ZERO/CPU_SET/sched_setaffinity on current glibc.
+#
+# WHY A RESIDUE ASSERTION, AND NOT A "grep for ERROR: AddressSanitizer" JOB
+# ------------------------------------------------------------------------
+# The defect class this job exists to catch is the undeclared inline-asm clobber: an asm
+# block writes a GPR, opmask or high vector register that it does not name in its clobber
+# list, so the compiler keeps a live value there across the block. AddressSanitizer cannot
+# see that at all - no invalid memory access ever occurs, the arithmetic is simply wrong.
+# Two such defects were live on Mlucas main simultaneously and produced zero sanitizer
+# reports while silently returning wrong residues for most AVX-512 FFT lengths a real user
+# would pick. A report-grepping job passes such a tree with flying colours.
+#
+# Consequently:
+#   * every case is checked against a pinned Res64, and a mismatch fails the job;
+#   * the FFT radices actually selected are checked too, so a future change to
+#     get_fft_radices() cannot silently move a case off the code path it was chosen for;
+#   * the exponent is pinned with -m, so the reference residues cannot drift with the
+#     self-test tables;
+#   * a missing Res64 line is a failure. Exit status alone is NOT sufficient: Mlucas
+#     prints "ERROR ERROR...Halting test" and then still exits 0.
+#
+# DO NOT ADD -fsanitize-recover=address. It is not a neutral convenience flag here: adding
+# it and nothing else makes a *broken* tree produce correct results, on both clang 18 and
+# clang 22 - i.e. it hides precisely the defect class this job exists for.
+#
+set -euo pipefail
+
+MODE=${1:?usage: $0 <avx2|avx512> [runner...]}
+shift
+RUNNER=("$@")
+
+# Per-case timeout. SIGKILL, not the default SIGTERM: a process running under Intel SDE
+# (Pin/ptrace) can end up parked where SIGTERM is never delivered and a plain `timeout` never
+# fires - one such run sat wedged for 13 hours. This is not hypothetical for this job in
+# particular: on a *broken* tree the SIGSEGV case sometimes wedges Pin instead of dying, which
+# is precisely when the timeout has to work.
+#
+# 600 s is ~15x the slowest passing case measured (128K under SDE with ASan), and 5 x 600 s
+# still fits inside the job's timeout-minutes, so a fully wedged run fails the job rather than
+# burning the whole runner allocation.
+CASE_TIMEOUT=${CASE_TIMEOUT:-600}
+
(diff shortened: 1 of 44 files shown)


MarkRose:
Two updates.

**Rebased onto current `main`.** #68 has landed, so the four clobber commits this was stacked on are gone and it is down to the CI change plus the SDE switch below.

**Intel SDE now comes from this repository rather than Intel.** Intel prunes old releases — 9.24, 9.27 and 9.33 all return 403 from `downloadmirror.intel.com` today — and nothing marks 9.53 as exempt, so an upstream pin here is a build break waiting to happen. The kit is now attached as the [`ci-intel-sde-9.24.0`](https://github.com/primesearch/Mlucas/releases/tag/ci-intel-sde-9.24.0) release: byte-identical to Intel's original, redistributed under the Intel Simplified Software License, which permits unmodified redistribution.

9.24.0 rather than something newer because it is the last kit that still has `-knl`/`-knm`, which the KNL job in #239 needs. One version across both jobs means one asset and one hash instead of two.

Checked the swap rather than assuming it:

- teensy and tiny self-tests of an AVX-512 build under `-skx` give **identical residues on 9.24 and 9.53** — 75 residues compared across the two stages — with `mlucas.cfg` matching apart from the `msec/iter` column;
- all five pinned-Res64 cases here pass unchanged under 9.24, run through `asan-simd-check.sh` with a clang ASan build exactly as CI does:

```
PASS avx512 fftlen 32K  radset 0  radices [32 32 16]   Res64 1A4EF8A0D172FBAA
PASS avx512 fftlen 32K  radset 1  radices [32 16 32]   Res64 1A4EF8A0D172FBAA
PASS avx512 fftlen 32K  radset 2  radices [16 32 32]   Res64 1A4EF8A0D172FBAA
PASS avx512 fftlen 32K  radset 3  radices [16 8 8 16]  Res64 1A4EF8A0D172FBAA
PASS avx512 fftlen 128K radset 0  radices [128 16 32]  Res64 040918890E98F8DA
```

Worth adding for the record on why this job earns its keep: runner AVX-512 support is a per-job lottery, not a steady state. In a single CI run I see Intel Xeon Platinum 8370/8573 (AVX-512 present) next to AMD EPYC 7763 (absent) and EPYC 9V74 (present in silicon, masked by the hypervisor). So a green check on a native AVX-512 job tells you nothing about whether any AVX-512 instruction actually executed. Running under SDE replaces that coin flip with a guarantee.

PR #222, opened 2026-07-30
makemake.sh: pass $WORDS to every Mfactor TU, not just factor.o

MarkRose:
> **Merge order: #167 → #215 → this.** Both are correctness fixes to code this change makes reachable. Details at the bottom.

`makemake.sh` passes `$WORDS` only to the `factor.o` rule, not to the generic `%.o` rule:

```make
552: factor.o: ../src/factor.c
553:    $(CC) $(CFLAGS) $(CPPFLAGS) -c ${ARGS[@]} -DFACTOR_STANDALONE $WORDS -DTRYQ=4 $<
554: %.o: ../src/%.c
555:    $(CC) $(CFLAGS) $(CPPFLAGS) -c ${ARGS[@]} $<          <-- no $WORDS
```

So for `mfac 3word avx`, `factor.o` sees `-DFACTOR_STANDALONE -DP3WORD -DTRYQ=4` and **every other translation unit sees none of the three**. `factor.h` keys several decisions off those macros, so `factor.c` and the `twopmodq*.c` routines it calls disagree about which configuration they are part of.

## What that costs today

`factor.h:47-51` sets `PIPELINE_MUL192` only `#ifdef P3WORD`. Since the `twopmodq*.c` TUs never see `P3WORD`, **all pipelined 192-bit `_q4`/`_q8` paths are dead in every shipped build.** Confirmed by preprocessing: `twopmodq192.c` with the real flags contains `__fprod` 0 times; 72 times with `-DP3WORD` added.

`factor_test.h:309-312` then prints `PIPELINE_MUL192 = 1` in the startup banner — **the binary's self-description is false.**

Two further disagreements, both benign today but recorded in the audit: `TRYQ` resolves to 8 in the TUs versus 4 in `factor.c` (`factor.h:53` recomputes it, and `factor.o` is additionally hardwired to 4), and `MAX_BITS_P|Q` to 57/96 versus 178/192. `factor.h:44` defaults to `P1WORD` when nothing is defined, which is silently what every `twopmodq*.c` TU gets regardless of the word-variant being built.

## Why the obvious objection turns out to be backwards

The natural worry is that switching `$WORDS` on activates a large body of pipelined code that has never been compiled into a shipped binary — and in this codebase, never-compiled code has repeatedly turned out to be broken (#213 retires an entire configuration that never compiled on any compiler generation).

**Measured, the opposite holds here: the pipelined macros are the correct ones and the scalar macros that ship today are wrong.**

`SQR_LOHI192` — the macro used by the shipped `PIPELINE_MUL192`-off path — drops a cross-term carry and is wrong for **25% of full-width 192-bit inputs**, in both its portable-C and x86_64-asm bodies. Cross-checked against GMP and against the in-tree `MULH192(x,x)` over 2.4M operand sets: 150,268 mismatches for the scalar macro, **0** for `SQR_LOHI192_q4`/`_q8`.

The consequence is not theoretical. Two genuine Mersenne factors were constructed with 64-bit `k` (`q` = 190 and 188 bits):

| build | `twopmodq192_q4` | `_q8` |
|---|---|---|
| stock `mfac 3word` | `0x0` | `0x00` |
| same, `-DPIPELINE_MUL192=1` | `0xF` | `0xFF` |

So deleting the dead pipelined code — the other candidate resolution — would have removed the only exact 192-bit squaring in the tree and cemented the bug.

**That root cause is #167**, arrived at independently here. This PR does not fix it; it removes the second, separate path to the same symptom.

## The change

`makemake.sh` only, +13/−1:

* the generic `%.o` rule gains `$WORDS -DFACTOR_STANDALONE` (the latter because `Mdata.h` carries `#error PxWORD requires FACTOR_STANDALONE`, which is what blocks the naive one-line version);
* the objdir name gains the word-variant. This is not cosmetic: `makemake.sh mfac 1word` followed by `mfac 3word` currently reuse the same `obj_mfac_<mode>/`, and make's timestamp check silently relinks the *1word* `factor.o` while reporting success. Once every object is variant-dependent, that stale-relink trap turns from a testing annoyance into a correctness hazard.

Verified across 15 build combinations: all build clean with identical warning profiles and pass `test_fac()`.

## Merge order

* **After #167** — it fixes `SQR_LOHI192`. Merging this first would leave the scalar path wrong *and* switch on the pipelined one, which is not harmful but muddies attribution.
* **After #215** — `MULH192_q4`/`_q8` are built on `MULH192_FAST`, whose approximate body is live for non-`YES_ASM` targets. (Note: on x86_64-gcc `MULH192_FAST` is exact in every flag combination tried, so an earlier claim that this configuration reaches the approximation on this platform does not reproduce. The constraint stands for other targets.)

## Also found, deliberately not fixed here

`-DTRYQ=8` does not compile on `main`: three "incompatible type for argument 1 of `twopmodq192_q8`" errors from `factor_test.h:2126/2226/2342`, where `uint192` is passed by value to a `uint64*` parameter. The sibling `TRYQ==4` arms already do it correctly and #215 adds a fourth correct one, so this is a fix rather than a deletion — but low priority, since `factor.o` is hardwired to `-DTRYQ=4` and the path is unreachable through the supported build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
@@ -112,7 +112,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
@@ -195,7 +195,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                 )
@@ -367,7 +367,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
@@ -421,7 +421,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
@@ -548,7 +548,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,6 @@ jobs:
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
@@ -114,7 +113,6 @@ jobs:
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
@@ -197,7 +195,6 @@ jobs:
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    mv -v build.log build_${word}.log
                 )
             done
         done
@@ -369,7 +366,6 @@ jobs:
                 (
                     cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
-                    mv -v build.log build_${word}.log
                     [[ ! -x Mfactor${word:+_$word} ]] || ./Mfactor${word:+_$word} -h
                 )
             done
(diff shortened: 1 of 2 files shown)


MarkRose:
Added the `ci.yml` half of this change, which was missing: appending the word argument to `DIR` means `makemake.sh mfac 1word AVX2` builds into `obj_mfac_1word_AVX2`, but every Mfactor loop in `ci.yml` still entered the shared `obj_mfac${mode:+_$mode}`. From the second iteration of the word loop onwards there was no `build.log` to rename:

```
mv: cannot stat 'build.log': No such file or directory
```

That subshell's status ends the step — the build itself is masked with `|| true`, the bookkeeping after it is not — which is why every job with an Mfactor loop has been red here since this PR was opened. Confirmed fixed: the `mv` failure is gone from the latest run.

The two halves can't be separated into different PRs: against `main`'s `makemake.sh` the per-word directories don't exist, so the `ci.yml` change only makes sense alongside the `DIR` change. That does put this PR into the `ci.yml` conflict set (#103, #216, #82, #73, #83), which is unwelcome but unavoidable.

**What remains red here is inherited from `main`, not from this PR.** `main` is currently 55 of 65 jobs red on `radix32_main_carry_loop.h:62: 'cy_i' undeclared` — the regression #126 merged, which #209 fixes. No PR branched from `main` can go green until #209 lands.

.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
                 echo -e "\nMfactor $word\n"
                 bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                 (
-                    cd obj_mfac${mode:+_$mode}
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                     make clean
                     mv -v build.log build_${word}.log
```
tdulcet: These `mv` commands were just to prevent the `build.log` files for different words from overwriting each other, so they would now be useless and could be removed.

makemake.sh
```diff
@@ -552,7 +564,7 @@ $Mfactor: \$(OBJS_MFAC)
 factor.o: ../src/factor.c
 	\$(CC) \$(CFLAGS) \$(CPPFLAGS) -c ${ARGS[@]} -DFACTOR_STANDALONE $WORDS -DTRYQ=4 \$<
 %.o: ../src/%.c
-	\$(CC) \$(CFLAGS) \$(CPPFLAGS) -c ${ARGS[@]} \$<
+	\$(CC) \$(CFLAGS) \$(CPPFLAGS) -c ${ARGS[@]} $WORDS_TU \$<
```
tdulcet: This could be simplified and then remove the above section:
```suggestion
	\$(CC) \$(CFLAGS) \$(CPPFLAGS) -c ${ARGS[@]} ${WORDS:+$WORDS -DFACTOR_STANDALONE} \$<
```

tdulcet (changes requested):
Thanks for looking into this.

These changed seemed to contradict the recommend instructions to build Mfactor (before the `makemake.sh` script) posted on the forum, even from Ernst himself (for example, see [this post](https://www.mersenneforum.org/node/19144?p=719431#post719431) or [this post](https://www.mersenneforum.org/node/20095?p=765130#post765130)).

However, I checked an old 2022 message from Ernst from when we were originally working to add this to the `makemake.sh script, and this PR does seem to be correct, at least as far the the compiler flags:
```
Been thinking about the best way to add Mfactor support to the current build script, here my thinking:

1. Treat "build Mfactor" much as we currently do a specified architecture: 'bash makemake.sh mfac" would create an obj_mfac directory ex to /src and build Mfactor in it. It would also be nice to support an optional number-of-limbs flag, e.g. '2word', which would only be allowed fo Mfactor builds, not Mlucas ones.

2. Build only the minimal fileset, which for Mfactor is (relative to obj_mfac):

	../src/factor.c.txt ../src/getRealTime.c ../src/get_fft_radices.c ../src/get_cpuid.c ../src/get_fp_rnd_const.c ../src/imul_macro.c ../src/mi64.c ../src/qfloat.c ../src/rng_isaac.c ../src/two*c ../src/types.c ../src/threadpool.c ../src/util.c

All sources can be built using the same preprocessor flags, '-DUSE_THREADS -DFACTOR_STANDALONE -DP[n]WORD', where DFACTOR_STANDALONE is ignored by all input files other than factor.c. Here the link step for default "single-limb-plus factors" build mode, meaning up to 96 bits:

	gcc -o Mfactor *o -lm -lpthread -lrt

For 2-limb it would be good to link to 'Mfactor_2word', and so forth.
```

tdulcet (approved):

PR #225, opened 2026-07-30
Free the known-factor mi64 vectors and the GMP gmp_one temporary (part of #96)

MarkRose:
Part of #96. Fixes the two leaks tdulcet reported from a Suyama (PRP-CF) test on 2026-07-18, which the earlier #105 did not touch — that PR fixed the `mers_mod_square.c` re-init leaks, and #96 was correctly reopened for these.

## The repro, since nothing in CI exercises this path

```sh
printf 'PRP=1,2,2063,-1,99,0,3,5,"4127,53639"\n' > worktodo.txt
ASAN_OPTIONS=detect_leaks=1 ./Mlucas
```
against a `nosimd` build with `-fsanitize=address,undefined`. Needs `-fft 1 -iters 100` first to get the cfg entry.

One trap worth recording: **Mlucas empties `worktodo.txt` on success**, so the file has to be rewritten between A/B runs. Skipping that produces a convincing false "clean".

## 1. `convert_base10_char_mi64` — 32 bytes / 2 objects

`convert_base10_char_mi64()` allocates the vector and hands ownership to the caller, freeing only on its own error path. `extract_known_factors()` never released it, so **every known factor of every PRP-CF or p-1 assignment leaked one vector**. tdulcet's 48-byte figure is the same leak with 3-limb factors.

The other four call sites are already correct — `factor.c:1134` frees at `:2521`, and the `factor_test.h` ones free locally. `util.c:172` would leak, but `print_mp_dec()` has no live caller.

## 2. `gmp_one` — 8 bytes / 1 object, and it is a pattern

`gcd()` inits `gmp_one` with `mpz_init_set_ui()` and omits it from the `mpz_clear` list. This is easy to miss because it is the *only* `mpz_t` in the file inited that way: plain `mpz_init()` has not allocated limb storage since GMP 5, but `mpz_init_set_ui()` does.

**The same omission is in `modinv()`**, so that is fixed here too.

Positive control that it is per-call rather than one-time: a p-1 run making two `gcd()` calls leaks 2 objects before the change and 0 after.

## Verification

Both leaks reproduced before and absent after, with residues byte-identical across the A/B pair. `makemake.sh nosimd` builds clean. Also checked that `fac` has no use after the new `free()` — all three reads precede it, and the one that persists the value (`mi64_set_eq` into `KNOWN_FACTORS`) copies rather than aliases.

## Why "part of #96" and not "fixes #96"

[...]

Issue #96, opened 2026-07-13
Memory leaks


tdulcet:
LeakSanitizer (part of the AddressSanitizer) detects several memory leaks. For example, after the medium self-test on ARM Linux:
```
==748689==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 115488 byte(s) in 144 object(s) allocated from:
    #0 0xb8d2e44ceb3c in __interceptor_calloc (/home/ubuntu/temp/Mlucas/obj/Mlucas+0xa3eb3c) (BuildId: 30c1440c2067790ee0adb39a71701e80680fda47)
    #1 0xb8d2e47d5980 in mers_mod_square /home/ubuntu/temp/Mlucas/obj/../src/mers_mod_square.c:1132:24
    #2 0xb8d2e481f4b4 in ernstMain /home/ubuntu/temp/Mlucas/obj/../src/Mlucas.c
    #3 0xb8d2e483edec in main /home/ubuntu/temp/Mlucas/obj/../src/Mlucas.c
    #4 0xfe5bf88d73fc  (/lib/aarch64-linux-gnu/libc.so.6+0x273fc) (BuildId: 9f32d43c341bff10b9e7196738eedcfc4f3cc36c)
    #5 0xfe5bf88d74d4 in __libc_start_main (/lib/aarch64-linux-gnu/libc.so.6+0x274d4) (BuildId: 9f32d43c341bff10b9e7196738eedcfc4f3cc36c)
    #6 0xb8d2e44577ac in _start (/home/ubuntu/temp/Mlucas/obj/Mlucas+0x9c77ac) (BuildId: 30c1440c2067790ee0adb39a71701e80680fda47)

Direct leak of 114912 byte(s) in 144 object(s) allocated from:
    #0 0xb8d2e44ceb3c in __interceptor_calloc (/home/ubuntu/temp/Mlucas/obj/Mlucas+0xa3eb3c) (BuildId: 30c1440c2067790ee0adb39a71701e80680fda47)
    #1 0xb8d2e47d7a44 in mers_mod_square /home/ubuntu/temp/Mlucas/obj/../src/mers_mod_square.c:1258:34
    #2 0xb8d2e481f4b4 in ernstMain /home/ubuntu/temp/Mlucas/obj/../src/Mlucas.c
    #3 0xb8d2e483edec in main /home/ubuntu/temp/Mlucas/obj/../src/Mlucas.c
[...]

PR #226, opened 2026-07-30
pm1: free the Stage-2 buf[] pointer array

MarkRose:
`pm1_stage2()` allocates `buf[]` as its own array of `num_b*m` `double *` at `pm1.c:1163`, then fills it with addresses pointing into the single large `a_ptmp` allocation:

```c
buf = (double **)calloc(num_b*m,sizeof(double *));
for(i = 0; i < num_b*m; i++) {
    buf[i] = a + i*npad;
    ...
}
```

The cleanup at `ERR_RETURN` frees `a_ptmp` and then merely nulls `buf`:

```c
free((void *)a_ptmp); a_ptmp = a = 0x0; buf = 0x0;
```

That releases the vectors those pointers address, but not the pointer array itself. **Every Stage-2 call leaks it.**

## Size

`num_b*m` pointers, so it scales with B2 and the memory setting rather than being a fixed cost — on the order of 16 KB for a typical run and more for a large one. It accumulates per Stage-2 invocation, including runs that return early through `ERR_RETURN` after a modmul error.

## Why freeing here is safe on every path

`buf` is `static double **buf = 0x0;` at file scope (`pm1.c:1033`) and is reset to `0x0` in this same cleanup, so an error return taken before the `calloc` at `:1163` frees a null pointer. `ERR_RETURN` is the common exit — the banner comment at `:1140` establishes that every error return from that point on routes through it — so no path escapes the free.

The pointers *inside* `buf` must not be freed individually; they alias into `a_ptmp`, which is already freed on the line above.

## Verification

`makemake.sh nosimd` builds clean.

## Relationship to #96 and #225

Found while producing the LeakSanitizer repro for #96, but kept separate: this is a different subsystem and reproduces on **any** p-1 Stage-2 run, not just the PRP-CF cofactor path that #225 addresses. It is not one of the leaks reported on #96, so it does not affect that issue's closure either way.

### DIFF
--- a/src/pm1.c
+++ b/src/pm1.c
@@ -2558,7 +2558,10 @@ S2_RETURN:
 #endif
 ERR_RETURN:
 	// Free the memory:
-	free((void *)a_ptmp); a_ptmp = a = 0x0; buf = 0x0;
+	// v21: buf[] is its own calloc'ed array of pointers - freeing a_ptmp releases the vectors those
+	// pointers address, but not the pointer array itself, which was leaked on every Stage-2 call:
+	free((void *)a_ptmp); a_ptmp = a = 0x0;
+	free((void *)buf); buf = 0x0;
 	free((void *)b); b = 0x0;
 	free((void *)map); map = 0x0;
   #ifdef MULTITHREAD


tdulcet (approved):
Thanks for finding and fixing this! It looks good to me.

PR #227, opened 2026-07-30
CI: restore full gcc ASan instrumentation once the clobber lists are correct

MarkRose:
> **Stacked on #221.** This branch is #221 plus one commit. Merge #221 first, or merge this and #221 goes with it.

The `ASan-SIMD-Linux` job that #221 adds gives its gcc leg `--param asan-stack=0`, which switches off stack-buffer-overflow, use-after-return and use-after-scope detection. That was a workaround for the build dying with `'asm' operand has impossible constraints`: ASan's stack instrumentation needs a register to reserve, and the over-declared 14-GPR clobber lists in `sse2_macro_gcc64.h` left it none.

That was a source defect rather than a gcc limitation - the same "clobber lists do not describe the asm" problem as the undeclared-register bugs, in the over-declared direction. The clobber lists have since been corrected on `main`, so the workaround is no longer buying anything, and this drops it. Both legs then run identical, full instrumentation.

## Verified on current `main`

The claim is directly testable, so here it is measured rather than argued - gcc 16.2.1, the job's exact CFLAGS (`-O3 -std=gnu99 -fsanitize=address -fno-omit-frame-pointer`), **no** `--param asan-stack=0`:

| build mode | result | `impossible constraints` errors |
| --- | --- | --- |
| `makemake.sh avx2` | rc=0 | **0** |
| `makemake.sh avx512` | rc=0 | **0** |

against 4 such errors at AVX2 and 6 at AVX-512 before the clobber lists were fixed.

## What changed in this PR since the last push

It used to be eleven commits: the asm clobber fixes, a duplicate of #221's ASan-matrix commit, and this ci.yml change on top. All of that is now gone except the ci.yml change, because the source half has landed on `main` independently - `3c9dbc7` ("Fix 'asm operand has impossible constraints' in hand-tuned SIMD/int asm") and `5eb3420` ("Free a GPR in the AVX-512 SSE2_RADIX16_DIF_0TWIDDLE for newer GCC") are both in `main` today, which is why the table above is all zeros.

Replaying those stale commits produced nine conflicts, every one in an inline-asm clobber list - precisely the code where a bad merge fails silently at runtime rather than at compile time. Rebuilding on #221 and keeping only the part that still has work to do avoids that entirely, and makes the PR reviewable: it is now a single, self-contained ci.yml diff.

## An alternative worth considering

Since `main` no longer needs the workaround, the tidiest outcome may be for #221 simply not to add it in the first place, at which point this PR has nothing left to do and can be closed. I have kept it as a separate commit rather than making that call unilaterally, because #221 is yours to shape - happy either way.

### DIFF
--- a/.github/scripts/asan-simd-check.sh
+++ b/.github/scripts/asan-simd-check.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+#
+# Build Mlucas for one SIMD mode under AddressSanitizer and assert that it computes the
+# *right answers* - not merely that it emits no sanitizer report.
+#
+# Usage:  .github/scripts/asan-simd-check.sh <avx2|avx512> [runner...]
+#
+#   <mode>       SIMD build mode, passed straight through to makemake.sh.
+#   [runner...]  Optional launcher prefixed to the binary, e.g. "$HOME/sde/sde64 -skx --".
+#                Needed for avx512 because GitHub's Azure runner SKUs are inconsistent
+#                about AVX-512 support; Intel SDE removes that dependency entirely.
+#
+# Reads CC / CFLAGS / LDFLAGS from the environment (makemake.sh's generated Makefile uses
+# "CFLAGS ?=", so an exported CFLAGS wins outright). Do NOT export CPPFLAGS: makemake.sh
+# sets "CPPFLAGS ?= -D_GNU_SOURCE ...", and without -D_GNU_SOURCE threadpool.c loses
+# CPU_ZERO/CPU_SET/sched_setaffinity on current glibc.
+#
+# WHY A RESIDUE ASSERTION, AND NOT A "grep for ERROR: AddressSanitizer" JOB
+# ------------------------------------------------------------------------
+# The defect class this job exists to catch is the undeclared inline-asm clobber: an asm
+# block writes a GPR, opmask or high vector register that it does not name in its clobber
+# list, so the compiler keeps a live value there across the block. AddressSanitizer cannot
+# see that at all - no invalid memory access ever occurs, the arithmetic is simply wrong.
+# Two such defects were live on Mlucas main simultaneously and produced zero sanitizer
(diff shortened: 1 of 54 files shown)


MarkRose:
Closing - this is now folded into #221, which is where it belonged.

This PR existed to remove the `--param asan-stack=0` workaround from the `ASan-SIMD-Linux` job. That job is introduced by #221 and does not exist on `main`, so every line this PR touched existed only in #221: against `main` its diff changed nothing at all. Keeping it separate meant one PR adding a 24-line workaround plus a `::warning` annotation, and a second PR immediately deleting them.

#221 now simply never adds the flag. The tree that results is the same; the review is one PR instead of two.

**Why the workaround is no longer needed**, since that was this PR's substance and it is worth recording. The flag was there because ASan's stack instrumentation needs a register to reserve, and the over-declared 14-GPR clobber lists in `sse2_macro_gcc64.h` left it none, so the build died with `'asm' operand has impossible constraints`. Those lists have since been corrected on `main` - `3c9dbc7` and `5eb3420` - so I measured it rather than assuming, gcc 16.2.1 with this job's exact CFLAGS and no `--param asan-stack=0`:

| build mode | result | `impossible constraints` errors |
| --- | --- | --- |
| `makemake.sh avx2` | rc=0 | **0** |
| `makemake.sh avx512` | rc=0 | **0** |

against 4 at AVX2 and 6 at AVX-512 before the fixes. A short comment at the site in #221 records why the flag used to be there.

The rest of what this branch carried - the asm clobber commits themselves - had also already landed on `main` by other routes, which is what the zeros above demonstrate. Replaying them produced nine conflicts, all in inline-asm clobber lists, i.e. the code where a bad merge fails silently at runtime rather than at compile time. There was nothing to gain from resolving those by hand to re-apply changes `main` already had.

No loss of coverage in closing this; the gcc leg gains full stack instrumentation via #221.

Issue #224, opened 2026-07-30
Merge order for the PRs that conflict or depend on each other


MarkRose:
There are ~100 open PRs against one codebase. Most are independent, but a subset must land in a particular order — either because they conflict textually, or because one is only correct once another is in. This is a reference for that subset.

Two things worth stating up front, because they are not obvious:

- **A clean merge is not a working merge.** Several pairs here merge with *zero* git conflicts and then fail to compile. Git warns about none of it.
- **Nothing enforces any of this today.** The list below was produced by materialising each candidate merge by hand and compiling the affected translation units, so it will go stale as PRs land.

Every `#N` below resolves to its live state. Landed work is pruned from this list rather than annotated: it is a todo list, not a changelog.

---

## Wave 0 — merge this first

This unblocks the most work and carries the most user-visible risk. Much of Wave 1 is downstream of it.

- **#152** — `nchunks = radix0>>1` drops the last work chunk for **every odd radix0** (5, 7, 9, 11, 13, 15, 63) in any `MULTITHREAD` build, including at `-cpu 0`. Broader than its title suggests. Head of the radix-63/1008/4032 chain below.

---

## Wave 1 — ordered chains

Arrows do not all mean the same thing, and it is worth knowing which kind you are looking at:

- **Chain A** — genuinely hard. Out-of-order merging gives a git conflict or a tree that does not compile, both verified.
- **Chain C** — a correctness pairing. Nothing breaks mechanically; merging half of it produces *wrong answers*, which is worse.
- **Chain E** — advisory. Neither entry blocks the other; each notes what the other unlocks.
- **Chain F** — hard, but for a reason git cannot see: the first PR may not reference a symbol the second one introduces, so the reverse order does not compile. Merged the wrong way round they also collide on a single line.
- **Chain G** — not an ordering: the single PR still needed to get `main` green again.
- **Chain I** — stacking that is invisible from the PR list: the second PR already contains the first, so merging it alone silently takes both, and merging them in the wrong order conflicts.

### Chain A — radix 63 / 1008 / 4032 silent wrong residues

**#152 → #204 → #175 → #205 → #210**, with #207 and #156 anywhere in the chain.

[...]

PR #230, opened 2026-07-30
[DRAFT] Integration branch v3: almost all open fix PRs, merged in #224's order

MarkRose:
All 109 open PRs from this fork, merged onto current `main` in the order set out in #224 — Wave 0, then one chain at a time, then everything unchained — with tdulcet's CI commit applied last. Supersedes #163 (which was built before that ordering existed, and before the PR-accuracy pass).

**Draft, and meant to stay that way.** It is a build/test vehicle for the whole set, not something to merge. Review and merge the individual PRs.

## Order

Merge commits, one per PR, in this sequence. Each arrow group is a chain from #224.

| # | group | PRs, in order |
|---|---|---|
| 0 | Wave 0 | #209, #167, #68, #203, #152 |
| 1 | Chain A — radix 63/1008/4032 | #204, #175, #205, #210, #207, #156 |
| 2 | Chain B — twopmodq | #185, #172, #212 |
| 3 | Chain C — Fermat-mod shift | #190, #218 |
| 4 | Chain D — 192-bit | #142, #215, #222, #223 |
| 5 | Chain E + the `ci.yml` hotspot | #103, #216, #82, #73, #83, #71, #75, #221, #227, #150, #151 |
| 6 | `imul_macro1.h` | #173, #198, #199, #200, #201, #202, then #213 |
| 7 | remaining conflict pairs | #69, #79, #72, #196, #112, #113, #179, #211 |
| 8 | unchained | the other 63, ascending |
| 9 | tdulcet | 6c84d6f7 "Updated CI to add tests and remove error masking" |

`rerere` was **off** for the whole rebuild, so every conflict below was resolved from the diffs rather than replayed from a cache recorded against a different tree.

## The 13 PRs that conflicted, and how each was resolved

Nine were mechanical. Four were not, and are the ones worth a reviewer's attention:

- **#216 + #103** (`ci.yml`, 9 hunks) — #216 gates the self-test exit status; #103 adds the `-prp` self-test loops and the AVX-512 execution probe. Neither side's text contains the other's, so the gating was re-applied to #103's loops rather than either side being taken whole. Non-piped call sites use `|| st=$?`, piped ones `${PIPESTATUS[0]}`: with no `shell:` key GitHub runs `bash -e {0}` — `-e` but *not* `pipefail` — so the two need different idioms.
- **#213 + the `imul_macro1.h` fixes** (7 hunks) — every conflict is #213 deleting a `MUL_LOHI64_SUBROUTINE` block that #198/#201/#202 had just fixed. Took the deletion; the fixes to the surviving arms are untouched and still present.
- **#178 + #218** — both rewrite the same G-check shift-recovery block. #218 makes Fermat-mod use a running `GCHECK_SHIFT` accumulator (and deletes the `filegrep` fallback); #178 generalises the Mersenne arm so a run too short to reach `ITERS_BETWEEN_GCHECKS` still computes the shift. Kept #218's structure with #178's two generalisations (`gchk_iter <=` for the guard, `i < ihi` for the halving loop).
- **#229 + #180** — #180 renames the `Suyama_CF_PRP` parameter and asserts the Pépin squaring count; #229 replaces that with a three-way branch on `nsq` (`p-1` → Pépin, `p` → already-PRP, else abort). Took #229 throughout, which is the superset.

The rest: **#212** (relocated block, #172's `qhalf` recompute already inside it), **#82 + #83 + #103** (three separate Mfactor run-guards, combined into one `if/elif/else`), **#79 + #71** (`ATTR_NO_ASAN` kept, #79's `void`/`thread_num` signature adopted), **#69** (insertion into a region #205 had emptied), **#196 + #72** (prototype union: #196's `accum_alloc`, #72's `**arr`), **#113 + #112** (both append to README's Contributing section), **#211 + #179** (#179's `RADIX992_J1` macro subsumes #211's `#ifdef` arms — the arms are dead in that file, its SIMD symbols are `#undef`'d), **#189 + #216** (two comments describing the same fix).

## Three breaks that no PR owns

All three are artifacts of the combination — each PR is fine alone — so they are fixed here, as their own commits, rather than pushed into anyone's branch:

- **`radix63_ditN_cy_dif1.c:557: 'stride' undeclared`** (#152 + #210). #210 removes the declaration with the loop it served; #152 adds a new use. Zero git conflicts. This is the break documented in #224; the fix is the one-liner recorded there, sized off `RE_IM_STRIDE`.
- **`factor.h:170: TRYQ = 2 and TRYQ > 8 only allowed if USE_FLOAT is defined`** (#83 + #193). #83's multiword fallback `#undef`s `USE_FMADD` *after* the `#ifndef TRYQ → 2` in the same block, so `TRYQ` stayed 2 in a build with neither `USE_FLOAT` nor `USE_FMADD`. Every 2word Mfactor build on AVX2, AVX-512 and ARM failed. Hoisting the fallback above the default fixes it.

  Worth noting for #83 and #193: on `main` this configuration fails too, one guard earlier (`P2WORD may not be used together with USE_FMADD`), and CI has never caught it because `main` runs the Mfactor build as `makemake.sh mfac … || true`. That mask is gone in this tree, so the failure became visible for the first time.

- **#219's lint lost sight of `factor.c`** (#219 + #193). The lint preprocesses each `src/*.c` with one common flag set, but `makemake.sh` builds the Mfactor TUs with `-DFACTOR_STANDALONE $WORDS -DTRYQ=4`. Without the pin, `factor.c` takes the `USE_FMADD` default of `TRYQ = 2` and stops at #193's new `#error`, so the lint dropped it in the avx2 and avx512 modes and exited 1 with "coverage is incomplete, so this run cannot be trusted" — the tool refusing to report success while checking less, which is the behaviour #219 argued for. Gave it a per-file flag table.

## What was built and run

Every configuration this hardware can reach, on the final tree.

### x86-64, gcc 16.1.1

| what | result |
|---|---|
| `makemake.sh {nosimd,sse2,avx,avx2,avx512}` | **5/5 build** |
| `makemake.sh mfac {,1word,2word,3word,4word,nword} × 5 modes` | **30/30 build** |
| `Mlucas -s tt` × 5 modes (avx512 under Intel SDE `-skx`) | 2 clean, 3 exit 2 — see below |
| `Mfactor… -m 127 -bmax 20` × 30 (test_fac self-tests) | **30/30 rc=0** |
| `asm-clobber-lint.py` / `--self-test` | **0 findings** / 16/16 |

### aarch64, Raspberry Pi 4, gcc 12.2.0

| what | result |
|---|---|
| `makemake.sh {nosimd,asimd}` + 12 Mfactor variants | **14/14 build** |
| `Mlucas -s tt`, both modes | **rc=0** |
| `Mfactor… -m 127 -bmax 20`, nosimd | **6/6 rc=0** |
| `Mfactor… -m 127 -bmax 20`, asimd | 6/6 SIGABRT — `twopmodq80.c:3248: No TF support on ARMv8!`, upstream, unrelated to this branch, and nothing runs those binaries (the ARM leg of tdulcet's Mfactor sweep uses `obj_mfac_NOSIMD`, which passes) |

Both integration fixes above were confirmed on ARM: before them, `makemake.sh asimd` failed on `stride` and every 2word Mfactor failed on `TRYQ`.

### The three self-tests that exit 2, and why it is not a regression

`sse2` fails at 7K, `avx` at 7K and 14K, `avx512` at 9K — all "excessive roundoff", none a wrong residue, so #216 returns 2 (too few usable radix sets) rather than 1 (wrong answer), and the CI gating tolerates 2.

Compared against `main` + #209 (the minimum that compiles in a non-AVX SIMD mode), the same lengths pass there. That is **not** a numerical difference. Driving both binaries at the same exponent, radix set and shift:

| shift | main + #209 | this branch |
|---|---|---|
| 64792 | `CD9D42BD1FC8898A`, MaxErr 0.375 | `CD9D42BD1FC8898A`, MaxErr 0.375 |
| 109517 | `CD9D42BD1FC8898A`, MaxErr 0.4375 | `CD9D42BD1FC8898A`, MaxErr 0.4375 |
| 0 | `CD9D42BD1FC8898A`, MaxErr 0.40625 | `CD9D42BD1FC8898A`, MaxErr 0.40625 |

Identical residues *and* identical error at every shift. What changed is **which shift the self-test draws**: #104 drops 1K from the sweep and #146 forces shift 0 whenever the leading radix is < 16, so the pseudo-random stream is at a different position by the time 7K is reached (64792 → 109517).

The real content of the finding is that these cases sit at `p/pmax_rec = 0.99997` and the roundoff gate is 0.40625, so 0.375 and 0.4375 both occur within the ordinary spread of one shift draw. `main` passes them by luck of the draw, not by margin — at shift 0 the 7K case is already at 0.40625, i.e. exactly on the gate. Same phenomenon as the 6K/ARM case in #216's notes.

## Deliberate deviations from a straight `git merge`

Three, all in `ci.yml`, all flagged here because they are choices rather than mechanics:

1. **tdulcet's commit removes the AVX-512 execution probe** (#103) and the per-word Mfactor run guards (#82/#83), replacing the latter with one unguarded `Mfactor -m 127 -bmax 20` sweep after the build loop. Applied as written, since applying his commit was the point. Checked on real hardware rather than assumed: the ARM branch of that sweep runs the `obj_mfac_NOSIMD` binaries, which pass — it is only the ASIMD ones that hit "No TF support on ARMv8!", and nothing runs those any more. The x86 side now runs AVX-512 Mfactor binaries unprobed, so a runner that cannot execute AVX-512 will SIGILL where it previously printed a warning.
2. **His commit also drops #216's self-test status gating** in the four jobs where they overlap; it predates #216. The gating was kept and his other changes to those jobs (`-s tt` for TSan, `[[ $HOSTTYPE == arm64 ]]` instead of the `sysctl neon` probe, no `continue-on-error`) applied on top.
3. **One line removed that no conflict required**: `run_avx512=1` in the Windows job. Its only reader was the Mfactor guard tdulcet's commit deletes, so it was left assigned and never read.


ldesnogu:
Isn't this superseded by #238? In which case this could be closed.

PR #233, opened 2026-07-31
radix960: set the serial cy_i carry pointer, fixing a Fermat-mod SIGSEGV at 960K

MarkRose:
radix960's serial setup assigns `cy_r` but leaves `cy_i` commented out:

```c
cy_r = tmp;	/* cy_i = tmp+0x78; */	tmp += 2*0x78;
```

The space is still reserved — `tmp += 2*0x78` — only the pointer assignment is missing, so `cy_i` stays null. Mersenne-mod never touches it. Fermat-mod does, at `radix960_ditN_cy_dif1.c:2066`:

```c
tmp->d0 = _cy_r[l][ithread];   tm2->d0 = _cy_i[l][ithread];   // tm2 = cy_i
```

and the run dies. Reproduced on stock `main` + #209:

```
./Mlucas -f 24 -fft 960 -iters 100      # single-threaded AVX2  ->  SIGSEGV
```

It's the radix set `(960,16,32)` specifically — the other four at that length, `(240,8,16,16)` and the three radix-60 sets, are unaffected. Threaded builds are fine because the worker routine has its own copies of these assignments, already live.

## The fix

Restore the two arms whose declaration guard admits `cy_i`, each wrapped in `#ifndef MULTITHREAD` exactly as **radix240 and radix1024 already do**. The SSE2 arm is deliberately left alone: the declaration at the top of the routine excludes plain SSE2, matching those same siblings.

## Verification

F24 at FFT length 960K, before → after:

| build | before | after |
|---|---|---|
| threaded AVX2 | 5 of 5 radix-sets | 5 of 5 |
| threaded SSE2 | 5 of 5 | 5 of 5 |
| **single-threaded AVX2** | **SIGSEGV** | **5 of 5** |
| single-threaded SSE2 | 5 of 5 | 5 of 5 |

AVX-512 can't be executed on my hardware; under Intel SDE the same case runs to completion without crashing.

## How it surfaced

CI on the combined-PR integration branch, once `config-fermat.sh` started propagating a failing exit status instead of swallowing it. On stock `main` this crash is reachable but silent in that path.

### DIFF
--- a/src/radix32_ditN_cy_dif1.c
+++ b/src/radix32_ditN_cy_dif1.c
@@ -198,9 +198,8 @@ int radix32_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[],
 	static double radix_inv, n2inv;
   #ifndef MULTITHREAD
 	double *addr;
-   #ifndef USE_SSE2
-	double *addi;
-   #endif
+	double *addi;	// v21 build fix: must exist in SSE2 builds too - the MODULUS_TYPE_GENFFTMUL
+					// branch of radix32_main_carry_loop.h dereferences it unconditionally (regression from 2bf9930)
 	double temp,frac;
   #endif
 	double scale, dtmp, maxerr = 0.0;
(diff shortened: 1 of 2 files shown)


MarkRose:
AVX-512 detail, now measured on the final code rather than asserted. Single-threaded, F24 at 960K under Intel SDE (`sde64 -skx`): **no crash**, and the run reports *2 of 5 radix-sets passed*. The three that don't are the radix-60 sets:

```
Using complex FFT radices  60 16 16 32   -> ERR_RADIX0_UNAVAILABLE
Using complex FFT radices  60 32 16 16   -> ERR_RADIX0_UNAVAILABLE
Using complex FFT radices  60  8  8  8 16 -> ERR_RADIX0_UNAVAILABLE
```

That is leading radix 60 being unavailable in an AVX-512 build — a deliberate capability rejection, unrelated to this PR and unchanged by it. The `(960,16,32)` set this PR fixes is one of the two that pass.

Full verification matrix for F24 @ 960K:

| build | before | after |
|---|---|---|
| threaded AVX2 / SSE2 | 5 of 5 | 5 of 5 |
| single-threaded AVX2 | **SIGSEGV** | 5 of 5 |
| single-threaded SSE2 | 5 of 5 | 5 of 5 |
| single-threaded AVX-512 (SDE) | — | 2 of 5, no crash; the 3 misses are `ERR_RADIX0_UNAVAILABLE` |

PR #234, opened 2026-07-31
mi64: stop viewing uint64 arrays through uint32 pointers (fixes an aarch64/clang+LTO wrong answer)

MarkRose:
`mi64_is_div_by_scalar32()` and its `_x4`/`_x8`/`32p` siblings declared their input as `const uint32 x[]` and doubled the caller-supplied (uint64) length internally, so **every call site handed them a uint64 array through a `(uint32 *)` cast.** The comment above the first of them described the resulting incompatible-pointer warnings as *expected*.

That is a strict-aliasing violation, and clang 14 on aarch64 acts on it: with `-flto` the call is inlined, TBAA concludes the `uint32` reads cannot alias the `uint64` object, and the function returns FALSE for inputs that are plainly divisible. `mi64_div_y32()` then trips its own cross-check:

```
Assertion 'mi64_is_div_by_scalar32((const uint32 *)x, y, len)' failed:
Results of mi64_div_y32 and mi64_is_div_by_scalar32 differ!
```

## What the disagreement actually is

Instrumenting the assertion printed the failing triples:

```
len=1 y= 3 rem=0 x[0]=255      255 = 3 × 85
len=1 y= 5 rem=0 x[0]=255      255 = 5 × 51
len=1 y= 7 rem=0 x[0]=763      763 = 7 × 109
len=1 y=11 rem=0 x[0]=2541     2541 = 11 × 231
len=1 y=13 rem=0 x[0]=2795     2795 = 13 × 215
```

All genuinely divisible — so `mi64_div_y32` is right and the *cross-check* is wrong. Only the check is affected; the division results callers consume were always correct. The job aborts, it does not compute anything incorrectly.

## Fix

Take `const uint64 x[]` and read each half by value:

```c
static uint32 mi64_word32(const uint64 x[], uint32 i)
{
	return (uint32)(x[i>>1] >> ((i&1)<<5));
}
```

No `(uint32 *)` view of a `uint64` object, so no aliasing violation — and because the shift addresses the *value* rather than the storage, no dependence on host byte order. The old word indexing assumed little-endian, so this is one less thing for a big-endian port to trip over. `mi64_is_div_by_scalar32p_x8()` is left alone: it takes eight separate arrays and has no callers.

## Verification

aarch64, Ubuntu clang 14.0.0, `-O3 -flto`, same branch and container throughout:

| build | disagreements | outcome |
|---|---|---|
| unfixed | **5** | abort |
| unfixed + `-fno-strict-aliasing` | 0 | clean — confirms the mechanism |
| unfixed + `-fno-lto` | 0 | clean — confirms the mechanism |
| **this fix, no workaround flags** | **0** | **exit 0, factoring self-tests pass** |

The two flag variants are diagnostics, not the fix — they each remove one leg of the mechanism (TBAA, and the inlining that exposes it).

On x86-64: Mlucas builds clean in `nosimd` and `avx2`; Mfactor's self-tests, which call `_x4` and `_x8` directly, pass. Real aarch64 hardware with Debian clang 14.0.6 never exhibited the bug either way — it needs this specific combination, which is why it shows up only in the `ubuntu-22.04-arm` + clang CI cell.

## How it surfaced

CI on the combined-PR integration branch. On stock `main` the same code is equally undefined; whether it misbehaves is up to the compiler.

### DIFF
--- a/src/mi64.c
+++ b/src/mi64.c
@@ -5784,13 +5784,10 @@ static uint32 mi64_word32(const uint64 x[], uint32 i)
 }
 
 /// Fast is-divisible-by-32-bit-scalar using Montgomery modmul and right-to-left modding:
-/*** NOTE *** Routine assumes x[] is a uint64 array cast to uint32[], hence the doubling-of-len
-is done HERE, i.e. user must supply uint64-len just as for the 'true 64-bit' mi64 functions!
-In other words, these kinds of compiler warnings are expected:
-
-	mi64.c: In function ‘mi64_div_y32’:
-	warning: passing argument 1 of ‘mi64_is_div_by_scalar32’ from incompatible pointer type
-	note: expected ‘const uint32 *’ but argument is of type ‘uint64 *’
+/*** NOTE *** len is the uint64 length, just as for the 'true 64-bit' mi64 functions; the
+doubling to a 32-bit word count is done HERE. x[] is read as uint64 throughout and the 32-bit
+words are extracted with mi64_word32(), so there is no uint64-array-viewed-as-uint32[] cast
+and none of the incompatible-pointer warnings that used to be expected here.
 */
 #ifdef __CUDA_ARCH__
 __device__
@@ -5815,7 +5812,7 @@ int mi64_is_div_by_scalar32(const uint64 x[], uint32 q, uint32 len)
 		qinv = qinv*((uint32)2 - q*qinv);
 	}
 	cy = (uint32)0;
-	dlen = len+len;	/* Since are processing a uint64 array cast to uint32[], double the #words parameter */
+	dlen = len+len;	/* Each uint64 limb holds two 32-bit words, so the word loop runs twice the uint64 len */
 	for(i = 0; i < dlen; ++i) {
 		const uint32 xcur = mi64_word32(x,i);
 		tmp  = xcur - cy;
@@ -5838,7 +5835,7 @@ int		mi64_is_div_by_scalar32p(const uint64 x[], uint32 q, uint32 qinv, uint32 le
 
 	ASSERT(qinv == qinv*((uint32)2 - q*qinv), "mi64_is_div_by_scalar32p: bad qinv!");
 	cy = (uint32)0;
-	dlen = len+len;	/* Since are processing a uint64 array cast to uint32[], double the #words parameter */
+	dlen = len+len;	/* Each uint64 limb holds two 32-bit words, so the word loop runs twice the uint64 len */
 	for(i = 0; i < dlen; ++i) {
 		const uint32 xcur = mi64_word32(x,i);
 		tmp  = xcur - cy;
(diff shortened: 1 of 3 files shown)


src/mi64.c
```diff
@@ -5797,11 +5811,12 @@ int mi64_is_div_by_scalar32(const uint32 x[], uint32 q, uint32 len)
...
 	dlen = len+len;	/* Since are processing a uint64 array cast to uint32[], double the #words parameter */
 	for(i = 0; i < dlen; ++i) {
-		tmp  = x[i] - cy;
+		const uint32 xcur = mi64_word32(x,i);
+		tmp  = xcur - cy;
 		/* Add q if had a borrow - Since we low=half multiply tmp by qinv below and q*qinv == 1 (mod 2^32),
 		   can simply add 1 to tmp*qinv result instead of adding q to the difference here:
 		*/
-		cy = (cy > x[i]); /* Comparing this rather than (tmp > x[i]) frees up tmp for the multiply */
+		cy = (cy > xcur); /* Comparing this rather than (tmp > x[i]) frees up tmp for the multiply */
```
ldesnogu, line 5822: The comment needs to be updated.

src/mi64.c
```diff
@@ -5811,19 +5826,20 @@ int mi64_is_div_by_scalar32(const uint32 x[], uint32 q, uint32 len)
...
```
ldesnogu, line 5845: Ditto.

ldesnogu (commented):
Apart from the need for comment updates, this LGTM.

MarkRose:
Both updated, thanks — and the same staleness turned out to reach two more comments.

The two you flagged, on the `dlen` computation in `mi64_is_div_by_scalar32` and `mi64_is_div_by_scalar32p`, said *"Since are processing a uint64 array cast to uint32[]"*. They now say what the doubling is actually for: each uint64 limb holds two 32-bit words.

The `/*** NOTE ***/` blocks above `mi64_is_div_by_scalar32` and `mi64_div_y32` — both in functions this PR changes — had the same problem, but worse: they promised compiler diagnostics.

> `warning: passing argument 1 of 'mi64_is_div_by_scalar32' from incompatible pointer type`
> `note: expected 'const uint32 *' but argument is of type 'uint64 *'`

Those can't occur any more: the parameter is `uint64[]` and the 32-bit words come from `mi64_word32()`. `gcc -Wall -Wextra` on `mi64.c` reports zero of them. A note telling the next reader to expect warnings that never appear seemed worse than no note, so I rewrote both rather than leave them.

Comment-only — every changed line is inside a comment. Verified with a nosimd build: `mi64.c` compiles clean, and the Mfactor self-test (which exercises these routines) reports `mi64_div quicktest passes` and `Factoring self-tests completed successfully`.

src/mi64.c
ldesnogu: The comment should read:

`/* Comparing this rather than (tmp > xcur) frees up tmp for the multiply */`

src/mi64.c
ldesnogu: The comment should read:

`/* Comparing this rather than (tmp > xcur) frees up tmp for the multiply */`

PR #235, opened 2026-07-31
Fix all ShellCheck findings and unmask the ShellCheck CI job

MarkRose:
## The mask

The `ShellCheck` job in `.github/workflows/ci.yml` carries `continue-on-error: true`. Its step has been **exiting 1 on every run** — the CI log shows the confusing combination of a green check mark on the job *and* `Process completed with exit code 1` inside it. So the job has never gated anything.

This PR fixes all of the findings and removes the mask.

## Findings, and how each was fixed

Scope is what CI actually checks out — `git ls-files '*.sh'` — i.e. `makemake.sh` and `config-fermat.sh`.

| Code | Count | Fix |
| --- | --- | --- |
| **SC2310** (`check-set-e-suppressed`) | 17 | **The optional check is removed from the `-o` list** (per review). |
| **SC2248** (`quote-safe-variables`) | 3 | `command -v "$MAKE"` ×2 in `makemake.sh`; `-iters "$ITERS"` in `config-fermat.sh`. |
| **SC2209** | 1 | `MAKE=make` → `MAKE='make'`. |
| **SC2054** | 1 | `args+=(-radset 8,8,16)` — narrow `# shellcheck disable=SC2054`. |

### SC2310 — the check is dropped rather than suppressed

Every one of the 17 is a call of the `try_flag` / `try_avx512_asm` / `try_lto` feature probes. A nonzero return from those is the *"this toolchain does not support that"* **answer**, not an error — so calling them from an `if`/`&&`/`!` condition, which suppresses `set -e` inside them, is exactly the intent. Each probe also does its own error handling (`tmpdir=$(mktemp -d) || return 1`), so none relies on `set -e`.

An earlier revision of this PR suppressed each call site with a narrow `# shellcheck disable=SC2310`. Per review, removing the optional check from the invocation is the better trade: it only ever fired on intentional code, so nine directives were pure noise. The block comment explaining *why* the probes are called conditionally is kept — that is worth documenting regardless of what the linter thinks.

### SC2054 — a genuine false positive

`8,8,16` is a single Mlucas `-radset` argument; the commas are part of the value, not array element separators. Suppressed in place with a comment saying so, and the verification below confirms it is still passed as one `argv` entry.

### SC2248 / SC2209, and the `time` sites

`$MAKE` holds `make` or `mingw32-make`, `$ITERS` holds `100` — whitespace- and glob-free literals set in the script itself, so the quotes are documentation rather than a semantic change. `MAKE='make'` likewise just tells ShellCheck (and the reader) that this is the string `make`, not a forgotten `$(make)`.

`time $MAKE ...` and `time $MLUCAS ...` are also now quoted. ShellCheck does not analyse the command word of a `time`-prefixed pipeline so these were never flagged, but per review the inconsistency was not worth keeping. Both are inert for the same reason as above.

### Not weakened

- Apart from `check-set-e-suppressed`, the `-o` option list is **unchanged**.
- `avoid-nullary-conditions`, `deprecate-which` and `require-double-brackets` produced **zero** findings — both scripts already use `[[ ]]` throughout — so no `[ ... ]` → `[[ ... ]]` conversions were needed. Good, because that rewrite changes word-splitting/glob semantics on the RHS of `=`.

## Verification

Using the `koalaman/shellcheck` container:

- new option list, both scripts: **rc=0, 0 findings**
- **negative control** — same files with `check-set-e-suppressed` re-enabled: rc=1, **18 SC2310 findings**. The check would still catch these if it were ever wanted back; this PR is not hiding them.
- **v0.9.0, v0.10.0 and stable all rc=0**, so `ubuntu-latest` shipping a different version carries no skew risk.
- `makemake.sh avx2` builds rc=0 with the quoted `time "$MAKE"`; `config-fermat.sh`'s quoted `time "$MLUCAS"` invokes the binary correctly.

Rebased onto current main: the branch was 46 commits behind, and since this PR unmasks the job it must be clean against what main contains now. Current main lints at 5 findings under the new option list — all of them the SC2248/SC2209/SC2054 this PR fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,7 +336,7 @@ jobs:
     steps:
     - uses: actions/checkout@v7
     - name: ShellCheck
-      run: shellcheck --version; shopt -s globstar; shellcheck -o avoid-nullary-conditions,check-set-e-suppressed,deprecate-which,quote-safe-variables,require-double-brackets -s bash **/*.sh
+      run: shellcheck --version; shopt -s globstar; shellcheck -o avoid-nullary-conditions,deprecate-which,quote-safe-variables,require-double-brackets -s bash **/*.sh
 
   macOS:
     name: Mlucas macOS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,8 +336,7 @@ jobs:
     steps:
     - uses: actions/checkout@v7
     - name: ShellCheck
-      run: shopt -s globstar; shellcheck -o avoid-nullary-conditions,check-set-e-suppressed,deprecate-which,quote-safe-variables,require-double-brackets -s bash **/*.sh
-      continue-on-error: true
+      run: shellcheck --version; shopt -s globstar; shellcheck -o avoid-nullary-conditions,check-set-e-suppressed,deprecate-which,quote-safe-variables,require-double-brackets -s bash **/*.sh
 
   macOS:
     name: Mlucas macOS
--- a/config-fermat.sh
+++ b/config-fermat.sh
@@ -80,5 +80,5 @@ for fft in "${!FFTS[@]}"; do
 	if [[ $f -le 17 || $f -ge 32 ]]; then
 		args+=(-shift 0)
 	fi
-	time $MLUCAS -f "$f" -fft "$fft" -iters "$ITERS" "${args[@]}" 2>&1 | tee -a config-fermat.log | grep -i 'error\|warn\|assert\|writing\|pmax_rec\|fft radices'
+	time "$MLUCAS" -f "$f" -fft "$fft" -iters "$ITERS" "${args[@]}" 2>&1 | tee -a config-fermat.log | grep -i 'error\|warn\|assert\|writing\|pmax_rec\|fft radices'
 done
--- a/config-fermat.sh
+++ b/config-fermat.sh
@@ -74,10 +74,11 @@ for fft in "${!FFTS[@]}"; do
 	args=("${ARGS[@]}")
 	# First we test the very fiddly F15 and then loop over F16 up to maximum
 	if [[ $f -eq 15 ]]; then
+		# shellcheck disable=SC2054 # '8,8,16' is one comma-separated Mlucas -radset argument, not three array elements
 		args+=(-radset 8,8,16)
 	fi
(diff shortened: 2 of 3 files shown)


tdulcet (changes requested):
Thanks for the fixes. Instead of adding all of these disable comments, I would just remove the `check-set-e-suppressed` lint altogether, as it no longer seems valuable for us. When I previously added it, there were no violations.

> * I deliberately left unflagged code alone, even where it looks inconsistent: `time $MLUCAS ...` and `time $MAKE ...` stay unquoted

For consistency, I would suggest fixing these as well. It looks like a bug in ShellCheck.

MarkRose:
Both done, thanks — the lint removal in particular is a much better trade than what I had.

**`check-set-e-suppressed` removed.** All 17 findings were calls of the `try_flag`/`try_avx512_asm`/`try_lto` probes, so every single one needed a directive — the check was only ever firing on intentional code. Gone from the `-o` list, along with the nine directives and the two lines of the probe block comment that existed solely to explain them. I kept the rest of that comment: why the probes are called conditionally is worth documenting regardless of the linter.

**`time $MAKE` / `time $MLUCAS` quoted.** You're right that the inconsistency wasn't worth keeping. Both are semantically inert — `MAKE` holds `make` or `mingw32-make`, `MLUCAS` holds `./Mlucas`, all single words — so this is documentation, not a behaviour change.

Verified with the `koalaman/shellcheck` container:

- new option list, both scripts: **rc=0, 0 findings**
- **negative control** — same files, lint re-enabled: rc=1, 18 SC2310 findings, so the check would still catch these if it were ever wanted back
- v0.9.0, v0.10.0 and stable all rc=0, so no version skew
- `makemake.sh avx2` builds rc=0 with the quoted `time "$MAKE"`, and `config-fermat.sh`'s quoted `time "$MLUCAS"` invokes the binary correctly

One note on the force-push: I rebased onto current main first. The branch was 46 commits behind, and since this PR *unmasks* the job it needs to be clean against what main contains now, not what it contained at fork time. Current main lints at 5 findings under the new option list — all of them the SC2248/SC2209/SC2054 this PR already fixes, so the rebased branch is clean. The review response itself is isolated in the last commit if that's easier to read.

tdulcet (approved):
Thanks.

PR #237, opened 2026-07-31
Fix "k10m" typos and radix44's mislabelled IMCI-512 warning

MarkRose:
The Xeon Phi Knights Corner target is spelled **`k1om`** — with a letter **O**, not a digit zero. It is the ELF machine name `EM_K1OM` (what `readelf` prints as `Machine: Intel K1OM`), the toolchain triple `k1om-mpss-linux-gcc`, and the build mode `makemake.sh` accepts.

Eleven sites spelled it `k10m`. Five of those are in **user-visible `WARN` messages**:

```
radix20_ditN_cy_dif1: No k10m/IMCI-512 support; Skipping this leading radix.
```

The rest are comments, one `#error` string, and `docs/knc.txt` (which manages to use both spellings in a single line).

Separately, `radix44_ditN_cy_dif1()`'s warning is a copy-paste of radix36's and reports the **wrong function name** to the user:

```c
// src/radix44_ditN_cy_dif1.c:370
WARN(HERE, "radix36_ditN_cy_dif1: No k10m/IMCI-512 support; ...
```

Anyone hitting that warning on a KNC build would go looking in the wrong file. Fixed to `radix44_ditN_cy_dif1`.

## No functional change

Comments and string literals only. Verified rather than asserted: an AVX2 build of this branch and of stock `main` (418f365b) produce **99 / 99 object files with byte-identical machine code** (`objdump -d`, headers stripped). Every changed site is either a comment or inside an `#ifdef USE_IMCI512` block that a non-KNC build never compiles. `./Mlucas -s tt` also returns rc=0.

Found while auditing the k1om cross-build in #236; kept separate so that one stays reviewable. No dependency between the two.

### DIFF
--- a/docs/knc.txt
+++ b/docs/knc.txt
@@ -9,7 +9,7 @@ How to emulate vinsertf128 ?
 
 AVX-512 instructions unavailable in IMCI-512:
 o XORPD(use VPXOR[D|Q] instead),MOVHPD,MOVLPD,PADDQ,PCMPQ
-o radix-16 tangent-DIF: twiddles-precomputation needs VRCP14PD (but k10m has VRCP23PS, 16-way SP reciprocal to full 23-bit precision). On k1om: use non-tangent version simply to avoid major recode effort
+o radix-16 tangent-DIF: twiddles-precomputation needs VRCP14PD (but k1om has VRCP23PS, 16-way SP reciprocal to full 23-bit precision). On k1om: use non-tangent version simply to avoid major recode effort
 o PREFETCHT*
 o CMOV - replace with equivalent conditional-branch (I use mostly where branch should be easily predictable, 2 addresses whose relative magnitudes indicate "is modulus Mersenne or Fermat"?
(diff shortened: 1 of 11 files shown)


docs/knc.txt
```diff
@@ -9,7 +9,7 @@ How to emulate vinsertf128 ?
 
 AVX-512 instructions unavailable in IMCI-512:
 o XORPD(use VPXOR[D|Q] instead),MOVHPD,MOVLPD,PADDQ,PCMPQ
-o radix-16 tangent-DIF: twiddles-precomputation needs VRCP14PD (but k10m has VRCP23PS, 16-way SP reciprocal to full 23-bit precision). On k1om: use non-tangent version simply to avoid major recode effort
+o radix-16 tangent-DIF: twiddles-precomputation needs VRCP14PD (but k1om has VRCP23PS, 16-way SP reciprocal to full 23-bit precision). On k1om: use non-tangent version simply to avoid major recode effort
```
tdulcet: I would preserve this one as is. We are keeping these `.txt` files for posterity, as Ernst left them. They are already outdated in many cases. Any new documentation should go in `.md` Markdown files.

tdulcet (changes requested):
Thanks, I just have a single nit.

tdulcet (approved):
Thanks.

PR #238, opened 2026-07-31
[DRAFT] Integration branch v4: add k1om and shellcheck fixes to v3

MarkRose:
Forked from `all-fixes-integration-v3` (#230) at `0bde7de2`, adding the two pieces that were deliberately held out of v3.

## What's on top of v3

| | |
|---|---|
| Revert `0bde7de2` | v3 dropped tdulcet's `Mlucas K1OM` job because it could not pass. It can now, so the job comes back — in reworked form. |
| **k1om fixes** (#236 + follow-up) | The Knights Corner cross-build works for the first time. |
| **ShellCheck fixes** (#235 + one more) | 23 findings → 0, and the job's `continue-on-error: true` mask removed. |

## k1om

Three defects in `carry_gcc64.h`'s IMCI-512 carry macro (#236), the worst being a `jz 1f` with **no `1:` label anywhere** — placement is load-bearing, since `andq $63,%rdx` sets ZF iff 64-byte-aligned and must reach the `k6` low-half store. Missing `\` continuations had also truncated the `#define`, spilling the cy_out store, maxerr store and the whole 8×8 transpose into file scope. Plus declaration guards narrower than their uses in three `radix*` files.

New here: **`USE_AVX512_SIEVE`**. `platform.h` makes `USE_IMCI512` imply `USE_AVX512`, so Mfactor compiled the vectorized small-primes sieve for KNC — which has no `vprolvd`, no `vmovups`, no `kmovw` and no register-form `vpbroadcastd`. The new macro (AVX-512 **without** IMCI) gates the sieve and the whole `psmall[]` table that feeds it: declaration, struct member, allocation, both population sites, thread-arg plumbing, parameter and prototype — so the guards stay exactly as wide as the uses. KNC falls back to the scalar bit-clearing loop, which is a live path regardless: the AVX-512 arm already defers to it for small exponents.

Proven a no-op off KNC: preprocessed `factor.c` is byte-identical for a real AVX-512 build (line markers aside). Under `-DUSE_IMCI512` the `vprolvd` asm and all 10 sieve-`psmall` references disappear.

**The CI job builds Mlucas only.** Mfactor does now cross-build for k1om, but trial factoring on a first-gen Xeon Phi is not a workload anyone runs, and six word-variant builds roughly quadruple a job that already downloads a 301 MB SDK and runs an emulated CentOS 7. The job also now re-exports `CFLAGS`/`CPPFLAGS` after sourcing the MPSS environment script: that script exports both itself and the Makefile uses `CFLAGS ?=`, so without this the build silently drops to `-O0`, where `radix16_dif_dit_pass_asm.h` fails on the `"e"` constraint for `pfetch_dist` and `threadpool.c` loses `CPU_ZERO`. Neither is a KNC defect and both would have looked like one.

## ShellCheck

#235 fixes 22 findings on the two scripts main has. This branch has a third (`.github/scripts/asan-simd-check.sh`, already clean) and one extra finding #235 could not have fixed — `exit $st`, on a line that comes from #216's self-test exit-status handling. Fixed here.

17 of the findings are `SC2310` on the `try_flag`/`try_avx512_asm`/`try_lto` feature probes, where a nonzero return *is the answer*; those carry narrow per-call-site disables rather than a control-flow rewrite of the build driver. The `-o` option list is unchanged — nothing was weakened to make findings disappear.

## Verification

- ShellCheck: **exit 0** across all three scripts with CI's exact command and option list.
- K1OM: the job's actual script, extracted from the YAML and run in `centos:7` — **rc=0, zero errors**, `Machine: Intel K1OM`.
- x86-64: `nosimd`, `avx2`, `avx512` and `mfac avx512` all build; `./Mlucas -s tt` rc=0 on nosimd and avx2.

Conflicts resolved (both non-mechanical, described in the merge commits): `factor.c`, where this branch's restructured small_p fallback block made git align its closing brace against the incoming `#endif`; and `config-fermat.sh`, where #216's status logic and #235's quoting both had to survive.

## Caveat

KNC still cannot be executed — no hardware, and SDE does not emulate IMCI. This is "compiles clean at `-O3`, and the scalar fallback is a path x86 exercises daily", not "verified correct on KNC".

Supersedes #230, which I'll close if this looks right.


MarkRose:
Superseded by #303, which is rebuilt from scratch on current `main` (`d74ee02`) with the live open-PR set. This branch had drifted 154 commits behind `main` while sitting 328 ahead, so rebuilding was cheaper and safer than rebasing.

Leaving this open rather than closing it, in case there is anything here worth pulling across.

PR #239, opened 2026-07-31
KNL build mode: produce a binary that runs on a Xeon Phi, or refuse to build; test it under SDE in CI

MarkRose:
Makes the `avx512_knl` build mode either produce a binary that runs on a Xeon Phi or refuse to build at all, and adds CI that executes KNL code under emulation so a regression is caught rather than shipped.

## 1. The deprecation notice was wrong, and harmful

`avx512_knl` printed:

```
Warning: The 'avx512_knl' option is deprecated, use 'avx512' instead.
```

The two modes are not interchangeable. KNL and KNM implement **AVX512F + CD (+ ER/PF)** — no **DQ, BW or VL** — while `avx512` targets Skylake-SP and enables all of them. An `avx512` binary dies almost immediately on a Xeon Phi; under Intel SDE's KNL chip check:

```
SDE-ERROR: Executed instruction not valid for specified chip (KNL): kmovd k0, ecx
Function: qfcos
```

`kmovd` is AVX512BW; KNL has only the 16-bit `kmovw`. Anyone with a Knights Landing box who followed that advice got a SIGILL. Notice removed, replaced with an explanation of why the mode has to stay.

## 2. The mode could not build, and then could build something worse

It hardcoded `-march=knl … -mavx512er`. Both were deprecated in GCC 14 and **removed in GCC 15**, so on GCC 16 it simply failed:

```
cc: error: unrecognized command-line option '-mavx512er'; did you mean '-mavx512bw'?
```

The base flag set is now `-mavx512f -mavx512cd -mfma` — the correct KNL feature set, deliberately **no dq/bw/vl**, since adding those is exactly the bug in §1 — with the KNL-specific flags probed individually.

The harder question is which toolchains can be trusted with the mode at all. Counting 256-bit EVEX inserts/extracts (AVX512VL, which KNL does not implement) across a representative set of translation units:

| compiler | `-march=knl` | `-mavx512er` | VL ops, base | VL ops, `+ -march=knl` |
| gcc 11.4 | yes | yes | 0 | 0 |
| gcc 12.4 | yes | yes | 0 | 0 |
| gcc 13.3 | yes | yes | 0 | 0 |
| **gcc 14.2 / 14.4** | **yes** | **yes** | **463** | **149** |
| gcc 15.3 | no | no | 202 | n/a |
| gcc 16.2 | no | no | 186 | n/a |
| clang 14.0 / 16.0 / 18.1 | yes | yes | 0 | 0 |
| clang 19.1 / 22.1 | yes | **no** | 0 | 0 |

This is a GCC bug, now filed as [PR target/127459](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=127459) - with only AVX512F enabled, GCC emits `vinserti32x4` in its EVEX.256 form, which requires AVX512VL, and `-mno-avx512vl` does not suppress it. A regression: clean in GCC 11-13, present in 14 through 16. The gate below is the workaround, and can be relaxed once that is fixed.

**GCC 14 accepts `-march=knl` and is still unsafe.** An earlier revision of this PR gated on whether the compiler still had the flag; that let GCC 14 straight through, and the binary it produces dies under SDE on `vinsertf32x4 ymm0, ymm0, xmm1, 0x1` in `radix40_ditN_cy_dif1`. On GCC 14 the flag actually *raises* the count rather than lowering it — radix144 alone goes from 12 to 76.

So the gate probes the generated code instead of the flag: it compiles one real FFT kernel with the KNL flag set and refuses if any 256-bit EVEX insert/extract comes out. That tests the property we need rather than a proxy for it. If `objdump` is unavailable the check cannot run, and it then falls back to the measured boundary — GCC 14 and later refused, anything else warned — so the mode never silently produces a binary that cannot run.

ER/PF are a separate axis from `-march=knl`, not implied by it: Clang 19 and later accept `-march=knl` while having dropped `-mavx512er`/`-mavx512pf` and not defining `__AVX512ER__`. They stay warnings rather than errors, because the tree's only ER code — the `VRCP28PD` block in `mi64_modmul53_batch()`, gated on `__AVX512ER__` — is unreachable: its sole caller sits inside a commented-out test harness in `mi64.c`. A toolchain without ER therefore builds a functionally identical binary. If that routine is ever wired up, this needs revisiting.

Behaviour on every toolchain tested:

| toolchain | outcome |
| --- | --- |
| gcc 11.4 – 13.3 | builds, no warnings |
| gcc 14.x | **refused** — flag accepted, codegen unsafe |
| gcc 15.x, 16.x | **refused** — no `-march=knl` |
| clang 14 – 18 | builds, no warnings |
| clang 19+, 22 | builds, with two ER/PF warnings |

## 3. New `Linux-KNL` CI job

Builds the mode and runs `-s tt` under `sde64 -knl`, which sets both CPUID *and* SDE's chip check, so any instruction outside the KNL ISA aborts the run instead of quietly succeeding on the AVX-512 host. That is the regression this job exists to catch — it would have caught §1.

Two deliberate pins:

- **`runs-on: ubuntu-24.04`, not `latest`.** GCC 13 is the last GCC that can target KNL at all, and 24.04 ships GCC 13 and Clang 18, both of which pass. On a newer image this job would correctly but unhelpfully refuse to build.
- **SDE 9.24.0, hash-pinned.** Intel removed the `-knl`/`-knm` knobs in 9.38 and no longer serves any pre-9.38 build — `downloadmirror.intel.com` returns 403 for 9.24, 9.27 and 9.33 alike. It is therefore hosted as a release asset on this repository, byte-identical to Intel's tarball, whose licence permits redistribution without modification, and verified by SHA-256 before unpacking.

## Verification

Every row below was run locally, with SDE 9.24.0 — the same version the CI job uses.

| toolchain | build | binary | `sde64 -knl` |
| --- | --- | --- | --- |
| gcc 13.3 | clean, no warnings | 0 VL instructions | `Res64: 5F1F741C1FBDA658` |
| clang 22.1 | clean, two ER/PF warnings | 0 VL instructions | `Res64: 5F1F741C1FBDA658` |
| gcc 14.4 | **refused** | — | (the binary it would have built dies on the first VL instruction) |
| gcc 16.2 | **refused** | — | — |

That residue matches what the same case produces from an ordinary x86-64 build, so this is a correct KNL run, not merely an absence of chip-check violations.

Also verified: the refusal paths with `objdump` hidden from `PATH` (gcc 14 refused by the version fallback, gcc 13 warns and proceeds); the `avx512` build failing on KNL exactly as in §1; and the ShellCheck finding count unchanged, with the new probe call sites carrying the same narrow `SC2310` disables as the existing ones.

The earlier revision of this description reported that SDE segfaulted locally and that the CI run step was therefore unverified. That was SDE 8.69.1, which this PR no longer uses; 9.24.0 runs correctly on this machine, so the run step is now exercised locally as well as in CI.

## A GCC bug found along the way

The VL instructions in the table above are a GCC defect, not something specific to Mlucas. Minimal case:

```c
void sink(int *);

void f(const unsigned *tab, unsigned long long k)
{
	int out[8];
	out[0] = tab[(k >>  0) & 15];
	/* ... eight of these ... */
	out[7] = tab[(k >> 28) & 15];
	sink(out);
}
```

`gcc -O3 -mavx512f -S` emits `vinserti32x4 $0x1, %xmm1, %ymm0, %ymm0` — the EVEX.256 form, requiring AVX512VL — with only AVX512F enabled and `__AVX512VL__` undefined. `-mno-avx512vl` does not suppress it. It is a regression introduced in GCC 14 and still present in 16.2; Clang is unaffected at every version tested. I have a Bugzilla account request pending and will file it upstream, and link it here.

This PR does not depend on that being fixed — the gate refuses the affected compilers regardless.

## KNM

No separate mode. KNM is KNL plus 4FMAPS/4VNNIW/VPOPCNTDQ, none of which Mlucas uses, so an `avx512_knl` binary already runs on it — and `-mavx5124fmaps`/`-mavx5124vnniw` are gone from both GCC 16 and Clang 22 anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,12 @@ jobs:
     # in GCC 15. The build degrades gracefully without them (see makemake.sh), but 24.04's GCC 13 is
     # the last stock toolchain that can target Knights Landing properly, so test with it while it lasts.
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        cc: [gcc, clang]
+      fail-fast: false
+    env:
+      CC: ${{ matrix.cc }}
     steps:
     - uses: actions/checkout@v7
     - name: Install
@@ -223,15 +229,13 @@ jobs:
         tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
     - name: Before script
       run: |
-        gcc --version
+        $CC --version
         ./sde-external-8.69.1-2021-07-18-lin/sde64 --version
     - name: Build
       run: |
-        set -o pipefail
         bash -e -o pipefail -- makemake.sh avx512_knl
     - name: Self-test under emulated Knights Landing
       run: |
-        set -o pipefail
         # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
         # the run rather than silently working on the AVX-512 host. That is the point of this job:
         # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,6 +202,46 @@ jobs:
             done
         done
 
+  Linux-KNL:
+    name: Mlucas KNL
+
+    # Pinned, not 'latest': -march=knl and the ER/PF extensions were deprecated in GCC 14 and removed
+    # in GCC 15. The build degrades gracefully without them (see makemake.sh), but 24.04's GCC 13 is
+    # the last stock toolchain that can target Knights Landing properly, so test with it while it lasts.
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v7
+    - name: Install
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y libgmp-dev
+    - name: Install Intel SDE
+      run: |
+        # Intel SDE 9.38 and later removed the -knl/-knm knobs, and Intel does not host archived
+        # releases, so pin an older kit. This is the mirror the xsimd project uses for its own CI.
+        wget -nv 'https://github.com/xtensor-stack/xsimd-testing-resources/releases/download/1.0.0/sde-external-8.69.1-2021-07-18-lin.tar.bz2'
+        tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
+    - name: Before script
+      run: |
+        gcc --version
+        ./sde-external-8.69.1-2021-07-18-lin/sde64 --version
+    - name: Build
+      run: |
+        set -o pipefail
+        bash -e -o pipefail -- makemake.sh avx512_knl
+    - name: Self-test under emulated Knights Landing
+      run: |
+        set -o pipefail
+        # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
+        # the run rather than silently working on the AVX-512 host. That is the point of this job:
+        # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
+        cd obj_avx512_knl
+        ../sde-external-8.69.1-2021-07-18-lin/sde64 -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
+        st=${PIPESTATUS[0]}
+        echo "Self-test exit status: $st"
+        # 0 = ok, 2 = too few usable radix sets at some length (informational). Anything else fails.
+        if (( st != 0 && st != 2 )); then exit "$st"; fi
+
   ASan-Linux:
     name: AddressSanitizer Linux
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,6 +314,12 @@ jobs:
     # in GCC 15. The build degrades gracefully without them (see makemake.sh), but 24.04's GCC 13 is
     # the last stock toolchain that can target Knights Landing properly, so test with it while it lasts.
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        cc: [gcc, clang]
+      fail-fast: false
+    env:
+      CC: ${{ matrix.cc }}
     steps:
     - uses: actions/checkout@v7
     - name: Install
@@ -328,15 +334,13 @@ jobs:
         tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
     - name: Before script
       run: |
-        gcc --version
+        $CC --version
         ./sde-external-8.69.1-2021-07-18-lin/sde64 --version
     - name: Build
       run: |
-        set -o pipefail
         bash -e -o pipefail -- makemake.sh avx512_knl
     - name: Self-test under emulated Knights Landing
       run: |
-        set -o pipefail
         # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
         # the run rather than silently working on the AVX-512 host. That is the point of this job:
         # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,6 +307,46 @@ jobs:
           ${{ github.workspace }}
           !mpss-*
 
+  Linux-KNL:
+    name: Mlucas KNL
+
+    # Pinned, not 'latest': -march=knl and the ER/PF extensions were deprecated in GCC 14 and removed
+    # in GCC 15. The build degrades gracefully without them (see makemake.sh), but 24.04's GCC 13 is
+    # the last stock toolchain that can target Knights Landing properly, so test with it while it lasts.
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v7
+    - name: Install
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y libgmp-dev
+    - name: Install Intel SDE
+      run: |
+        # Intel SDE 9.38 and later removed the -knl/-knm knobs, and Intel does not host archived
+        # releases, so pin an older kit. This is the mirror the xsimd project uses for its own CI.
+        wget -nv 'https://github.com/xtensor-stack/xsimd-testing-resources/releases/download/1.0.0/sde-external-8.69.1-2021-07-18-lin.tar.bz2'
+        tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
+    - name: Before script
+      run: |
+        gcc --version
+        ./sde-external-8.69.1-2021-07-18-lin/sde64 --version
+    - name: Build
+      run: |
+        set -o pipefail
+        bash -e -o pipefail -- makemake.sh avx512_knl
+    - name: Self-test under emulated Knights Landing
+      run: |
+        set -o pipefail
+        # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
+        # the run rather than silently working on the AVX-512 host. That is the point of this job:
+        # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
+        cd obj_avx512_knl
+        ../sde-external-8.69.1-2021-07-18-lin/sde64 -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
+        st=${PIPESTATUS[0]}
+        echo "Self-test exit status: $st"
+        # 0 = ok, 2 = too few usable radix sets at some length (informational). Anything else fails.
+        if (( st != 0 && st != 2 )); then exit "$st"; fi
+
   ASan-Linux:
     name: AddressSanitizer Linux
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,9 +310,11 @@ jobs:
   Linux-KNL:
     name: Mlucas KNL
 
-    # Pinned, not 'latest': -march=knl and the ER/PF extensions were deprecated in GCC 14 and removed
-    # in GCC 15. The build degrades gracefully without them (see makemake.sh), but 24.04's GCC 13 is
-    # the last stock toolchain that can target Knights Landing properly, so test with it while it lasts.
+    # Pinned, not 'latest': GCC 13 is the last GCC that can target Knights Landing at all. GCC 14
+    # accepts -march=knl but generates AVX512VL anyway, and GCC 15 removed the flag; makemake.sh
+    # refuses both rather than emit a binary that dies on the first illegal instruction (see the
+    # measured table there). 24.04 ships GCC 13 and Clang 18, both of which pass, so pin to it -
+    # on a newer image this job would correctly but unhelpfully refuse to build.
     runs-on: ubuntu-24.04
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -320,6 +320,14 @@ jobs:
       fail-fast: false
     env:
       CC: ${{ matrix.cc }}
+      # Intel SDE 9.38 and later removed the -knl/-knm knobs, and Intel no longer serves any
+      # pre-9.38 build -- downloadmirror.intel.com returns 403 for 9.24, 9.27 and 9.33 alike.
+      # 9.24.0 is therefore hosted as a release asset on this repository, byte-identical to
+      # Intel's original and redistributed under the Intel Simplified Software License, which
+      # permits redistribution without modification. See the ci-intel-sde-9.24.0 release.
+      SDE_URL: https://github.com/primesearch/Mlucas/releases/download/ci-intel-sde-9.24.0/sde-external-9.24.0-2023-07-13-lin.tar.xz
+      SDE_SHA256: 87a36b821d0fb8cbcbfeffa359753445f951b225b5f4747689ae16417c4f04ac
+      SDE_DIR: sde-external-9.24.0-2023-07-13-lin
     steps:
     - uses: actions/checkout@v7
     - name: Install
@@ -328,14 +336,16 @@ jobs:
         sudo apt-get install -y libgmp-dev
     - name: Install Intel SDE
       run: |
-        # Intel SDE 9.38 and later removed the -knl/-knm knobs, and Intel does not host archived
-        # releases, so pin an older kit. This is the mirror the xsimd project uses for its own CI.
-        wget -nv 'https://github.com/xtensor-stack/xsimd-testing-resources/releases/download/1.0.0/sde-external-8.69.1-2021-07-18-lin.tar.bz2'
-        tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
+        curl -fsSL -o sde.tar.xz "$SDE_URL"
+        # Verify before unpacking: the pin is what makes a hosted binary trustworthy.
+        echo "$SDE_SHA256  sde.tar.xz" | sha256sum -c -
+        mkdir -p "$SDE_DIR"
+        tar -xf sde.tar.xz -C "$SDE_DIR" --strip-components=1
+        rm -f sde.tar.xz
     - name: Before script
       run: |
         $CC --version
-        ./sde-external-8.69.1-2021-07-18-lin/sde64 --version
+        "./$SDE_DIR/sde64" --version
     - name: Build
       run: |
         bash -e -o pipefail -- makemake.sh avx512_knl
@@ -345,7 +355,7 @@ jobs:
         # the run rather than silently working on the AVX-512 host. That is the point of this job:
         # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
         cd obj_avx512_knl
-        ../sde-external-8.69.1-2021-07-18-lin/sde64 -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
+        "../$SDE_DIR/sde64" -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
         st=${PIPESTATUS[0]}
         echo "Self-test exit status: $st"
         # 0 = ok, 2 = too few usable radix sets at some length (informational). Anything else fails.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,10 +332,6 @@ jobs:
       SDE_DIR: sde-external-9.24.0-2023-07-13-lin
     steps:
     - uses: actions/checkout@v7
-    - name: Install
-      run: |
-        sudo apt-get update -y
-        sudo apt-get install -y libgmp-dev
     - name: Install Intel SDE
       run: |
         curl -fsSL -o sde.tar.xz "$SDE_URL"
@@ -351,17 +347,33 @@ jobs:
     - name: Build
       run: |
         bash -e -o pipefail -- makemake.sh avx512_knl
+        # Build Mfactor for KNL too - it has its own AVX-512 sieve, a distinct body of code from
+        # Mlucas, so it is worth compile-covering here. Not run yet: under SDE it segfaults on a
+        # tiny exponent because of the AVX-512 small-p sieve bug that #114 fixes. With #114 merged
+        # in locally the same binary completes '-m 521 -bmax 20' cleanly, so the run step can be
+        # added once that lands.
+        bash -e -o pipefail -- makemake.sh mfac avx512_knl
     - name: Self-test under emulated Knights Landing
       run: |
         # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
         # the run rather than silently working on the AVX-512 host. That is the point of this job:
         # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
+        #
+        # Threads are worth using despite the emulator: measured under 'sde64 -knl', '-s t' takes
+        # 810s on one core against 351s on four, a 2.3x speedup. ('-s tt' on its own shows almost
+        # none, but it is short enough to be dominated by startup.) That is what makes 'tiny'
+        # affordable here, and it roughly triples the FFT lengths covered: 16K-240K atop 2K-15K.
         cd obj_avx512_knl
-        "../$SDE_DIR/sde64" -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
-        st=${PIPESTATUS[0]}
-        echo "Self-test exit status: $st"
-        # 0 = ok, 2 = too few usable radix sets at some length (informational). Anything else fails.
-        if (( st != 0 && st != 2 )); then exit "$st"; fi
+        CPUS="0:$(( $(nproc --all) - 1 ))"
+        # No pipe, so no pipefail needed and no knl.log: this job uploads no artifacts, and the
+        # run's output goes to the workflow log either way.
+        for s in tt t; do
+          st=0
+          "../$SDE_DIR/sde64" -knl -- ./Mlucas -s $s -cpu "$CPUS" || st=$?
+          echo "Self-test -s $s exit status: $st"
+          # 0 = ok, 2 = too few usable radix sets at some length (informational). Else fail.
+          if (( st != 0 && st != 2 )); then exit "$st"; fi
+        done
 
   ASan-Linux:
     name: AddressSanitizer Linux
--- a/makemake.sh
+++ b/makemake.sh
@@ -255,9 +255,22 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 			ARGS+=(-DUSE_AVX512 -march=skylake-avx512 -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
 			;;
 		avx512_knl)
-			echo "Building for avx512_knl SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
-			echo "Warning: The 'avx512_knl' option is deprecated, use 'avx512' instead."
-			ARGS+=(-DUSE_AVX512 -march=knl -mavx512f -mavx512cd -mavx512er -mfma)
+			echo "Building for AVX-512 on Knights Landing / Knights Mill in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
+			# NOT interchangeable with the 'avx512' mode: KNL/KNM implement only AVX512F + CD (+ ER/PF),
+			# with no DQ, BW or VL. An 'avx512' build targets Skylake-SP and will SIGILL on a Xeon Phi -
+			# the first thing it hits is 'kmovd' (BW; KNL has only the 16-bit 'kmovw'), and any
+			# EVEX-encoded 256-bit op would need VL. So keep this mode, and keep it free of dq/bw/vl.
+			ARGS+=(-DUSE_AVX512 -mavx512f -mavx512cd -mfma)
+			# -march=knl and the ER/PF extensions were deprecated in GCC 14, removed in GCC 15, and are
+			# gone from recent Clang too. The build does not need them - it compiles and links with
+			# just f/cd/fma - so probe for each and carry on without whichever the toolchain has
+			# dropped, rather than failing outright on a modern compiler:
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -march=knl && ARGS+=(-march=knl)
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -mavx512er && ARGS+=(-mavx512er)
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -mavx512pf && ARGS+=(-mavx512pf)
 			;;
 		avx512)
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -202,6 +202,46 @@ jobs:
...
+        set -o pipefail
+        # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
+        # the run rather than silently working on the AVX-512 host. That is the point of this job:
+        # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
+        cd obj_avx512_knl
+        ../sde-external-8.69.1-2021-07-18-lin/sde64 -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
+        st=${PIPESTATUS[0]}
+        echo "Self-test exit status: $st"
+        # 0 = ok, 2 = too few usable radix sets at some length (informational). Anything else fails.
+        if (( st != 0 && st != 2 )); then exit "$st"; fi
```
tdulcet, line 247: This is redundant with the `set -o pipefail` above. Also, `set -o pipefail` is already the default with Bash on GitHub Actions, so those could be removed as well.

.github/workflows/ci.yml
```diff
@@ -202,6 +202,46 @@ jobs:
...
+        sudo apt-get install -y libgmp-dev
+    - name: Install Intel SDE
+      run: |
+        # Intel SDE 9.38 and later removed the -knl/-knm knobs, and Intel does not host archived
+        # releases, so pin an older kit. This is the mirror the xsimd project uses for its own CI.
+        wget -nv 'https://github.com/xtensor-stack/xsimd-testing-resources/releases/download/1.0.0/sde-external-8.69.1-2021-07-18-lin.tar.bz2'
+        tar -xf sde-external-8.69.1-2021-07-18-lin.tar.bz2
+    - name: Before script
+      run: |
+        gcc --version
```
tdulcet: Maybe test with Clang on KNL as well.

makemake.sh
```diff
@@ -255,9 +255,22 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
+			# -march=knl and the ER/PF extensions were deprecated in GCC 14, removed in GCC 15, and are
+			# gone from recent Clang too. The build does not need them - it compiles and links with
+			# just f/cd/fma - so probe for each and carry on without whichever the toolchain has
+			# dropped, rather than failing outright on a modern compiler:
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -march=knl && ARGS+=(-march=knl)
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -mavx512er && ARGS+=(-mavx512er)
+			# shellcheck disable=SC2310 # a failed probe is an answer, not an error
+			try_flag -mavx512pf && ARGS+=(-mavx512pf)
```
tdulcet: We should output a warning if any of those KNL specific options are unavailable, as the resulting performance would then be less than what the user is expecting.

tdulcet (changes requested):
Thanks, this will be very helpful to prevent regressions. I just have some minor suggestions for improvement.

Note there is some overlap with the changes in #155.

MarkRose:
All three done.

**Clang on KNL** — added, using the same `cc` matrix and `CC` env as the other jobs; "Before script" now prints `$CC` instead of hardcoding gcc. Worth more than symmetry, as it turns out: the two compilers are retiring the KNL-specific flags at different rates, so the legs exercise different degrade paths. Locally GCC 16 rejects all of `-march=knl`, `-mavx512er` and `-mavx512pf`, while Clang 22 still accepts `-march=knl` and rejects only the ER/PF pair.

**Warning on unavailable flags** — added, one per dropped flag. `try_flag` was degrading silently, so a user on a modern toolchain got a working but generically-tuned binary with nothing to say Knights-specific codegen had been skipped. Verified: GCC 16 emits three warnings, Clang 22 emits two, both still build rc=0 — matching exactly what each compiler accepts.

**The two `set -o pipefail` lines** — removed, and you're right that they're redundant. One correction to the reasoning though, because it matters elsewhere in this file: **`set -o pipefail` is not the default for `run:` steps.** GitHub's default shell is `bash -e {0}`; the pipefail variant applies only when a step declares `shell: bash`, which these don't. Only 5 steps in `ci.yml` do, and the repo otherwise compensates by writing `bash -e -o pipefail` explicitly — 17 times.

These two are redundant for their own reasons rather than that one:

- the **Build** step immediately invokes `bash -e -o pipefail -- makemake.sh`, which sets it for the shell that actually matters;
- the **Self-test** step reads `${PIPESTATUS[0]}` directly instead of relying on the pipeline's status.

Confirmed the Self-test guard still behaves without it — exit 0 and 2 pass, 1 and 132 fail the step.

So the removals are safe here, but I'd avoid generalising it: dropping `set -o pipefail` from a step that has a real pipeline and no explicit `shell: bash` would silently stop that pipeline failing the job.

ShellCheck unchanged at 19 findings before and after, none on the new loop — those 19 are the pre-existing ones #235 addresses.

One note on coverage: the job is pinned to `ubuntu-24.04`, whose GCC 13 and Clang 18 both still support the full KNL flag set, so CI won't actually exercise the degrade path. That's deliberate — the pin is there to keep testing real Knights targeting while a stock toolchain can still do it — but it does mean the warning path is verified locally rather than in CI.

.github/workflows/ci.yml
MarkRose: I've left this one as-is, and I think the premise doesn't hold for this job specifically.

> `set -o pipefail` is already the default with Bash on GitHub Actions

That's true only for steps that explicitly say `shell: bash`, which Actions expands to `bash --noprofile --norc -eo pipefail {0}`. Steps with no `shell:` get the implicit default instead, which is `bash -e {0}` — no `pipefail`. This job doesn't set one at any level:

```
job key: Linux-KNL
  workflow-level defaults: None
  job-level defaults:      None
  step shell:              None
```

The five `shell: bash` occurrences elsewhere in the file are all in the Windows/msys jobs, which is probably what makes it look global.

So without `${PIPESTATUS[0]}`, `$?` after

```bash
sde64 -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
```

would be `tee`'s status — always 0 — and the exit-status check below it could never fire. The job would pass unconditionally, including on a KNL build that aborts on an illegal instruction, which is the one thing it exists to catch.

The alternative would be adding `shell: bash` to the step and dropping `PIPESTATUS`. I've kept the current form because it makes the dependency local and visible rather than resting on a default that differs between steps in this same file, but I'm happy to switch if you prefer the consistency.

Your other two points are in: Clang is in the matrix now, and `makemake.sh` warns when a KNL-specific flag is unavailable rather than silently degrading.

MarkRose:
Pushed a follow-up that changes where this job gets Intel SDE.

It was fetching 8.69.1 from `xtensor-stack/xsimd-testing-resources` — a third-party project's CI-assets release — with no integrity check, and then executing it. Intel no longer serves any pre-9.38 build (`downloadmirror.intel.com` returns 403 for 9.24, 9.27 and 9.33 alike), which is what pushed this to a third-party mirror originally.

SDE **9.24.0** is now attached to this repository as the [`ci-intel-sde-9.24.0`](https://github.com/primesearch/Mlucas/releases/tag/ci-intel-sde-9.24.0) release — byte-identical to Intel's original, redistributed under the Intel Simplified Software License, which permits redistribution without modification. The job now fetches from there and runs `sha256sum -c` before unpacking.

Two reasons for 9.24 over 8.69.1: 8.69.1 segfaults on recent kernels, so the job was on borrowed time as runner images advance; and 9.24 is what I validated against locally. `sde64 -knl` executes the ER assembly in `mi64.c` (`vrcp28pd`), while `sde64 -skx` rejects it with

```
SDE-ERROR: Executed instruction not valid for specified chip (SKYLAKE_SERVER): vrcp28pd zmm1, zmm0
```

so the chip check is demonstrably enforcing the KNL ISA rather than quietly running everything.

**Correction to an earlier version of this comment:** I first wrote that no runner in the matrix has AVX-512. That was wrong — it came from sampling a single run. The real picture is more interesting.

Azure's pool is mixed, and each job draws from it independently. From one run on 29 July, six x86 Linux cells:

| job | CPU | `avx512f` |
|---|---|---|
| ubuntu-22.04, gcc | AMD EPYC 7763 | no |
| ubuntu-24.04, clang | AMD EPYC 9V74 | no |
| **ubuntu-24.04, gcc** | **Intel Xeon Platinum 8573** | **yes** |
| **ubuntu-26.04, gcc** | **Intel Xeon Platinum 8370** | **yes** |
| ubuntu-26.04, clang | AMD EPYC 9V74 | no |
| ubuntu-22.04, clang | AMD EPYC 7763 | no |

EPYC 7763 has no AVX-512 in silicon; EPYC 9V74 has it but the hypervisor masks it off; the Xeon Platinum parts have it. Two of six in the same run. Runs I sampled on 13 and 18 August happened to draw all-AMD, which is what misled me.

So native AVX-512 coverage exists, but it is a **coin flip per job** — a green check tells you nothing about whether any AVX-512 instruction executed. That also explains a class of failure we have been calling flaky: an AVX-512-only defect appears when a job lands on a Xeon and vanishes when it lands on an EPYC, same commit, same job.

Which is the argument for emulation rather than against it. `sde64` runs the AVX-512 path on every job regardless of the draw, deterministically, with a chip check that also rejects instructions outside the target ISA.

I verified the new steps verbatim against the hosted asset before pushing: hash matches, `sde64` reports 9.24.0, and a KNL build completes the teensy self-test with exit status 0.

Separately, I ran the full ladder under `-knl` locally to see what this job would cover if extended: `tt` 19 s, `t` 595 s, `s` 8211 s. So adding `-s t` here would cost about ten minutes and roughly triple the FFT lengths exercised (16 K–240 K on top of 2 K–15 K), if you think that's worth it.

MarkRose:
Rebased onto current `main` — it was 124 commits behind and #155 has since landed. Two conflicts:

- **`makemake.sh`** — `main` still had the hardcoded `-march=knl -mavx512er` flag line; this branch's probing version replaces it, which is the point of the PR.
- **`ci.yml`** — this branch's `Linux-KNL` job landed exactly where #155 put `Linux-K1OM`. Those are two different Xeon Phi parts (Knights Landing vs Knights Corner), so both jobs are kept.

Build re-verified after: `makemake.sh avx512_knl` succeeds on gcc 16, dropping `-march=knl`, `-mavx512er` and `-mavx512pf` and emitting the three degradation warnings as intended.

## But the degraded build is not actually KNL-safe on gcc >= 15

Running the new SDE gate locally against that gcc-16 build fails, and it is worth recording because it means one sentence in this PR is wrong:

```
TID 0 SDE-ERROR: Executed instruction not valid for specified chip (KNL):
  0x...: vinserti32x4 ymm5, ymm1, xmm0, 0x1
  Function: ernstMain   File: ../src/Mlucas.c:378
```

`VINSERTI32X4` with a **ymm** destination is the EVEX.256 form, which requires AVX512VL. KNL has F/CD/ER/PF and no VL — exactly the case the comment I added warns about ("any EVEX-encoded 256-bit op would need VL"). Compiling `Mlucas.c` with identical flags (`-mavx512f -mavx512cd -mfma -DUSE_AVX512`, no `-march=knl`) across toolchains:

| compiler | `vinserti32x4` with ymm dest |
| --- | --- |
| gcc 13.3 | 0 |
| gcc 15.3 | 19 |
| gcc 16.2 | 3 |

gcc 13 emits none whether or not `-march=knl` is passed, so this is not the missing `-march=knl` — it is gcc >= 15 using AVX512VL encodings in an AVX512F-only build. Adding `-mno-avx512vl` does not suppress them.

So the warning text is misleading as written:

> The result will still run on Knights Landing/Mill, but slower than a toolchain retaining Knights support would produce.

On gcc >= 15 the result does **not** run on Knights Landing — it builds, then dies on the first `vinserti32x4`. The CI job is unaffected, because it deliberately pins `ubuntu-24.04` (gcc 13) for exactly this class of reason, so the SDE gate there stays meaningful. It is the fallback path on a newer toolchain that quietly produces a broken binary.

Options, and I have no strong preference:

1. Make the fallback refuse: if `-march=knl` is unavailable, error out rather than build something that cannot run on the target. Loud, and arguably what a cross-targeting build mode should do.
2. Keep building but correct the warning to say the result may not be KNL-executable on toolchains that dropped Knights support, and that a KNL-targeting build needs gcc <= 14.
3. Leave as-is and treat it as a compiler bug to report upstream — the `-mno-avx512vl` result suggests gcc is ignoring an explicit disable, which looks reportable on its own.

Happy to implement whichever you prefer; I did not want to change the PR's behaviour unilaterally on top of a rebase.

.github/workflows/ci.yml
tdulcet, line 273: Could you update this to download from GitHub releases as well.

.github/workflows/ci.yml
```diff
@@ -299,6 +299,60 @@ jobs:
...
+      # permits redistribution without modification. See the ci-intel-sde-9.24.0 release.
+      SDE_URL: https://github.com/primesearch/Mlucas/releases/download/ci-intel-sde-9.24.0/sde-external-9.24.0-2023-07-13-lin.tar.xz
+      SDE_SHA256: 87a36b821d0fb8cbcbfeffa359753445f951b225b5f4747689ae16417c4f04ac
+      SDE_DIR: sde-external-9.24.0-2023-07-13-lin
+    steps:
+    - uses: actions/checkout@v7
+    - name: Install
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y libgmp-dev
```
tdulcet: GMP is already installed by default on GitHub Actions.

.github/workflows/ci.yml
```diff
@@ -299,6 +299,60 @@ jobs:
...
+        mkdir -p "$SDE_DIR"
+        tar -xf sde.tar.xz -C "$SDE_DIR" --strip-components=1
+        rm -f sde.tar.xz
+    - name: Before script
+      run: |
+        $CC --version
+        "./$SDE_DIR/sde64" --version
+    - name: Build
+      run: |
+        bash -e -o pipefail -- makemake.sh avx512_knl
```
tdulcet, line 349: Should we build Mfactor as well for KNL?

.github/workflows/ci.yml
```diff
@@ -299,6 +299,60 @@ jobs:
...
+    - name: Self-test under emulated Knights Landing
+      run: |
+        # -knl sets both CPUID and SDE's chip check, so any instruction outside the KNL ISA aborts
+        # the run rather than silently working on the AVX-512 host. That is the point of this job:
+        # an 'avx512' (Skylake-SP) build dies here on kmovd, which is exactly the regression to catch.
+        cd obj_avx512_knl
+        "../$SDE_DIR/sde64" -knl -- ./Mlucas -s tt 2>&1 | tee knl.log
```
tdulcet: Could running multithreaded reduce the runtime under the emulator?

> Separately, I ran the full ladder under `-knl` locally to see what this job would cover if extended: `tt` 19 s, `t` 595 s, `s` 8211 s. So adding `-s t` here would cost about ten minutes and roughly triple the FFT lengths exercised (16 K–240 K on top of 2 K–15 K), if you think that's worth it.

If we could reduce the runtime, I believe we should try to run the tiny self-test and maybe the Mfactor self-test as well.

.github/workflows/ci.yml
tdulcet: I would prefer the opposite, using `set -o pipefail` instead of manually checking the exit status.

However, since we are not currently uploading any artifacts in this job, we do not need to write the `knl.log` file, so we could just remove the `tee` command instead and thus eliminate this pipe altogether.

makemake.sh
```diff
@@ -271,8 +271,27 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
+			ARGS+=(-DUSE_AVX512 -mavx512f -mavx512cd -mfma)
+			# -march=knl and the ER/PF extensions were deprecated in GCC 14, removed in GCC 15, and are
+			# gone from recent Clang too. The build does not need them - it compiles and links with
+			# just f/cd/fma - so probe for each and carry on without whichever the toolchain has
+			# dropped, rather than failing outright on a modern compiler:
+			# Warn per dropped flag: the build still runs on KNL/KNM without them, but it is
+			# tuned for a generic AVX-512 target and ER/PF-backed reciprocal and prefetch
+			# sequences are unavailable, so it will be slower than the user is expecting.
+			for knl_flag in -march=knl -mavx512er -mavx512pf; do
+				# shellcheck disable=SC2310 # a failed probe is an answer, not an error
```
tdulcet: We already disabled the ShellCheck `check-set-e-suppressed` option in #235, so this disable comment is no longer needed.

tdulcet (changes requested):
> a compiler bug to report upstream — the `-mno-avx512vl` result suggests gcc is ignoring an explicit disable, which looks reportable on its own.

This sounds like a serious GCC bug. Is there an existing open issue about it somewhere? If not, please do report it to them.

> 1. Make the fallback refuse: if `-march=knl` is unavailable, error out rather than build something that cannot run on the target. Loud, and arguably what a cross-targeting build mode should do.

In the meantime, I would probably prefer this option.

MarkRose:
Done — the fallback now refuses.

Your instinct was right, and it is stronger than "arguably what a cross-targeting build mode should do": the old warning was actively false. It told the user the degraded build "will still run on Knights Landing/Mill, but slower", and on a compiler that dropped `-march=knl` it does not run there at all.

The reason the two coincide is worth recording, because it is what makes the probe a sound gate rather than a tuning detail. Counting VL-only EVEX-256 ops (`vinserti32x4` with a ymm destination) in `Mlucas.c` built with `-mavx512f -mavx512cd -mfma`:

| compiler | VL instructions | has `-march=knl` |
| --- | --- | --- |
| gcc 13.3 | 0 | yes |
| gcc 15.3 | 19 | no |
| gcc 16.2 | 3 | no |

gcc 13 emits none whether or not `-march=knl` is passed, so the VL encodings come from the newer compilers rather than from the missing flag, and `-mno-avx512vl` does not suppress them. A toolchain that still has the flag is therefore exactly one old enough to be KNL-safe — so gating on it is not a proxy for the real property by luck, it tracks it.

Verified both directions:

```
gcc 16.2 (no -march=knl):  makemake.sh avx512_knl -> rc=1, error, no binary produced
gcc 13.3 (has it):         rc=0, and 0 VL instructions in the resulting binary
```

ER/PF stay warnings rather than errors — they are performance-only, so a toolchain with `-march=knl` but without them is correct, just slower.

tdulcet (changes requested):
Please see the previous review feedback above: https://github.com/primesearch/Mlucas/pull/239#pullrequestreview-5125354582.

> gcc 13 emits none whether or not `-march=knl` is passed, so the VL encodings come from the newer compilers rather than from the missing flag, and `-mno-avx512vl` does not suppress them.

If possible, we should probably test with more than just three versions of GCC before hardcoding this assumption into the `makemake.sh` script. Maybe we should test with Clang as well, and then update the comment and error messages as necessary.

> ER/PF stay warnings rather than errors — they are performance-only, so a toolchain with `-march=knl` but without them is correct, just slower.

Considering that there is actual AVX512 ER code in Mlucas, the `-mavx512er` flag is probably more important than `-march=knl`, if `-march=knl` does not always enable that ER code.

MarkRose:
Both points done, and widening the compiler set was the right call — it found a hole in the gate I had proposed.

## The wider matrix

Eleven toolchains across both families, counting 256-bit EVEX inserts/extracts (AVX512VL, which KNL does not implement) over a representative set of translation units rather than `Mlucas.c` alone:

| compiler | `-march=knl` | `-mavx512er` | VL ops, base | VL ops, `+ -march=knl` |
| gcc 11.4 | yes | yes | 0 | 0 |
| gcc 12.4 | yes | yes | 0 | 0 |
| gcc 13.3 | yes | yes | 0 | 0 |
| **gcc 14.2 / 14.4** | **yes** | **yes** | **463** | **149** |
| gcc 15.3 | no | no | 202 | n/a |
| gcc 16.2 | no | no | 186 | n/a |
| clang 14.0 | yes | yes | 0 | 0 |
| clang 16.0 | yes | yes | 0 | 0 |
| clang 18.1 | yes | yes | 0 | 0 |
| clang 19.1 | yes | **no** | 0 | 0 |
| clang 22.1 | yes | **no** | 0 | 0 |

**GCC 14 accepts `-march=knl` and is still unsafe.** So "does the compiler still have the flag" was the wrong test — it let GCC 14 straight through, and the binary it produces dies under SDE on `vinsertf32x4 ymm0, ymm0, xmm1, 0x1` in `radix40_ditN_cy_dif1`. `-march=knl` there actually *raises* the count rather than lowering it; radix144 alone goes from 12 to 76.

My earlier three-version sample missed this because it measured `Mlucas.c` only, where GCC 14 is clean — the VL instructions are in the radix kernels. That is exactly the assumption you were pushing back on, and it did not survive contact with GCC 14.

So the gate now probes the generated code: compile one real FFT kernel with the KNL flag set and refuse if any 256-bit EVEX insert/extract comes out. That tests the property we need rather than a proxy for it, and it will keep working as toolchains move. If `objdump` is unavailable it degrades to a warning.

Verified on all four paths:

| toolchain | outcome |
| --- | --- |
| gcc 16.2 — no flag | refused, flag message |
| gcc 14.4 — flag accepted, unsafe codegen | refused, codegen message |
| gcc 13.3 — flag, clean, ER present | builds, no warnings; 0 VL ops; `sde64 -knl` → `Res64: 5F1F741C1FBDA658` |
| clang 22.1 — flag, clean, no ER/PF | builds with the two warnings; `sde64 -knl` → same residue |

matching the residue the same case gives from an ordinary x86-64 build. The CI job stays on ubuntu-24.04 (GCC 13 + Clang 18, both of which pass); I corrected its comment, which still said the mode degrades gracefully.

## `-mavx512er` vs `-march=knl`

You were right that these are separate axes — Clang 19 and later accept `-march=knl` while having dropped `-mavx512er`/`-mavx512pf` entirely, and do not define `__AVX512ER__`. So `-march=knl` does not imply the ER code.

It is nonetheless safe to leave ER as a warning, and for a concrete reason rather than "performance only" as I put it before: the tree's only ER code is the `VRCP28PD` block in `mi64_modmul53_batch()`, gated on `__AVX512ER__`, and that function is unreachable — its sole caller sits inside a commented-out test harness in `mi64.c`. A toolchain without ER therefore builds a functionally identical binary, which is what the Clang 22 row above demonstrates. If that routine is ever wired up, this should be revisited.

## The GCC bug

I could not find an existing report. It is real and self-inconsistent: with `-mavx512f -mavx512cd -mfma -mno-avx512vl`, GCC leaves `__AVX512VL__` undefined and still emits

```
62 f3 7d 28 38 c1 01    vinserti32x4 $0x1,%xmm1,%ymm0,%ymm0
```

where EVEX byte `28` sets L'L=01, i.e. 256-bit, which requires AVX512VL. It reproduces on 14.2, 14.4, 15.3 and 16.2, and `-march=knl` does not suppress it either. It does not affect Clang, which emits none at any version tested.

Since it is still present in current GCC I am happy to reduce it to a minimal test case and file it upstream — just say if you would rather I did not, as it is somewhat tangential to this PR.

MarkRose:
Reduced the GCC bug to a minimal test case. It turns out not to need `-mno-avx512vl` at all — plain `-mavx512f` is enough, which makes it a cleaner statement of the problem than what I described earlier:

```c
void sink(int *);

void f(const unsigned *tab, unsigned long long k)
{
	int out[8];
	out[0] = tab[(k >>  0) & 15];
	out[1] = tab[(k >>  4) & 15];
	out[2] = tab[(k >>  8) & 15];
	out[3] = tab[(k >> 12) & 15];
	out[4] = tab[(k >> 16) & 15];
	out[5] = tab[(k >> 20) & 15];
	out[6] = tab[(k >> 24) & 15];
	out[7] = tab[(k >> 28) & 15];
	sink(out);
}
```

```
$ gcc -O3 -mavx512f -S bug.c -o -
	vinserti32x4	$0x1, %xmm1, %ymm0, %ymm0
```

That is the EVEX.256 form, which requires AVX512VL, emitted with only AVX512F enabled and `__AVX512VL__` undefined. The eight stores get SLP-vectorised into a 256-bit vector and the insert used to build it needs VL; seven stores do not trigger it. No inline asm or unusual clobbers are involved — that was a red herring from the original kernel.

Confirmed against the ISA by running the same binary on two emulated chips:

```
$ sde64 -knl -- ./bugtest
TID 0 SDE-ERROR: Executed instruction not valid for specified chip (KNL):
  vinserti32x4 ymm0, ymm0, xmm1, 0x1        Function: f

$ sde64 -skx -- ./bugtest
225 196 169 144 121 100 81 64
```

It is a regression introduced in GCC 14 and still present on current GCC:

| compiler | emits the VL instruction |
| --- | --- |
| gcc 11.4, 12.4, 13.3 | no |
| **gcc 14.2, 14.4, 15.3, 16.2** | **yes** |
| clang 14, 18, 19, 22 | no |

On GCC 14, where `-march=knl` still exists, all of `-mavx512f`, `-mavx512f -mno-avx512vl`, `-march=knl` and `-march=knl -mno-avx512vl` produce it, with `__AVX512VL__` undefined in every case. The `-march=knl` row is the pointed one: that names a CPU with no AVX512VL whatsoever.

I have requested a GCC Bugzilla account — it has been long enough since I last filed anything that the old one is gone — and will file this upstream as a `[14/15/16 Regression]` once the account comes through. I will link the report here when it exists.

None of that blocks this PR: the build-mode gate here already refuses GCC 14 on the strength of the generated code rather than waiting for a fix upstream, so we do not ship a binary that cannot run on the target it names.

MarkRose:
One more hole closed, prompted by re-reading the goal of this mode: never hand back a binary that cannot run on the target it names.

The codegen probe is the authority, but it needs `objdump` (or `llvm-objdump`) and a probe TU that compiles. If it cannot run, the previous behaviour was to warn and build anyway — which on GCC 14 or later means shipping exactly the broken binary this is meant to prevent.

So when the probe is unavailable, it now falls back to the measured version boundary: GCC 14 and later are refused, everything else still gets the warning. That fallback is deliberately narrow — hardcoding a version is the thing the probe exists to avoid, so it applies only when the real check cannot run, and never to Clang, which is unaffected at every version tested.

Verified with `objdump` hidden from `PATH`:

| toolchain | outcome |
| --- | --- |
| gcc 14.4 | refused, naming the version and suggesting installing binutils |
| gcc 13.3 | warns and proceeds, as before |

and with `objdump` present, the two existing paths are unchanged — gcc 14.4 refused by the codegen check, gcc 16.2 refused for having no `-march=knl`.

tdulcet (changes requested):
Thanks for investigating this issue and for filing the bug with GCC. It will be interesting to hear what they say.

I am requesting changes per my previous review two days ago: https://github.com/primesearch/Mlucas/pull/239#pullrequestreview-5125354582

> The codegen probe is the authority, but it needs `objdump` (or `llvm-objdump`) and a probe TU that compiles.

Could you use `gcc -S` (or `clang -S`) instead of requiring `objdump`?

MarkRose:
All of it done — and apologies, six of these were from your 6 September review and I answered only the compiler-matrix question at the time. Going through them:

## `gcc -S` instead of `objdump`

Much better, thank you. Same detection — verified identical on both sides, gcc 14 emitting 4 VL-form `vinserti32x4` and gcc 13 emitting 0 — but it needs only the compiler: no binutils, no assembler, no object file, and it is quicker than `-c`.

It also deletes something I was not happy with. The "probe could not run" fallback, which guessed from the GCC version, existed *only* because `objdump` might be absent. With `-S` that cannot happen, so the fallback is gone entirely and the sole remaining failure mode is a compiler that cannot compile the probe TU — now an error rather than a guess.

## The rest

- **`SC2310` disables** — removed, all three. Dead since #235 turned off `check-set-e-suppressed`.
- **`libgmp-dev`** — removed. Confirmed rather than assumed: the main Linux job installs only `libhwloc-dev`, so GMP is clearly already there. (The one remaining `libgmp-dev` install is in the container job, where it genuinely is not.)
- **`tee` and the pipe** — both gone. Nothing uploads `knl.log`, the output reaches the workflow log anyway, and with no pipe there is no `pipefail` question left to answer.
- **SDE from GitHub releases** — already done in the previous round; I re-checked that the asset downloads and matches the pinned SHA-256.

## Multithreading under the emulator — you were right to ask

I nearly got this wrong. Measuring `-s tt` alone suggested threads barely help (22s → 19s), but `tt` is too short to show anything past startup. On the actual candidate:

| `-s t` under `sde64 -knl` | time |
| --- | --- |
| 1 core | 810 s |
| 4 cores | 351 s |

**2.3x.** So the emulator does not serialise the way the first measurement implied. The self-test step now uses all runner cores.

## Adding `-s t`

Worth it on those numbers: about six minutes, and it roughly triples the FFT lengths exercised — 16K–240K on top of 2K–15K.

## Mfactor for KNL

Added as a **build**, and the answer to whether it should also run is "yes, but not yet".

It compiles fine and is worth covering, since its AVX-512 sieve is a separate body of code from Mlucas. But under SDE it segfaults immediately on a tiny exponent — and that is not a KNL problem: it is the AVX-512 small-p sieve bug, since `-m 521` sits below the sieving-prime bound. With **#114** merged in locally, the same binary gives `M(521) has 0 factors in range k = [0, 16336320], passes 0-15`, rc=0.

So I have left the run step out with a comment saying why, rather than add a step that is red until #114 lands. Happy to add it the moment that merges — it is three lines.

PR #242, opened 2026-08-01
Add self-test residues for the 1008K FFT-length family (1008K through 516096K)

MarkRose:
Adds self-test coverage for the **1008K FFT-length family** — 1008K and every power-of-two multiple through 516096K. Ten entries in `MersVec` (LL) and ten in `MvecPRP` (PRP).

These lengths use leading radix 1008 and 4032 (the `63·2^k` family) and previously had **no self-test coverage whatsoever**, which is a fair part of why the defects in those carry paths went unnoticed for as long as they did.

Residues generated and shared by **Ken (kriesel)** — thank you.

## Verification

Every one of the twenty values was independently recomputed rather than taken on trust: gcc 16, AVX2, `-shift 0`, 100 iterations, thread counts of 2, 4 and 8. All twenty match to the bit.

| FFT length | exponent | LL | PRP |
|---|---|:--:|:--:|
| 1008K | 19694291 | ✅ | ✅ |
| 2016K | 38795051 | ✅ | ✅ |
| 4032K | 76411513 | ✅ | ✅ |
| 8064K | 150481337 | ✅ | ✅ |
| 16128K | 296307203 | ✅ | ✅ |
| 32256K | 583354309 | ✅ | ✅ |
| 64512K | 1148280899 | ✅ | ✅ |
| 129024K | 2259875809 | ✅ | ✅ |
| 258048K | 4446690749 | ✅ | ✅ |
| 516096K | 8747833277 | ✅ | ✅ |

1008K, 2016K and 4032K were additionally cross-checked across two or three different radix sets each, all agreeing — a residue must be invariant under the choice of radix set, so that is the strongest check available short of a second implementation.

## Table bookkeeping

`numTest` 136 → 146, with `numSmall` +2, `numMedium` +3, `numLarge` +2, `numHuge` +2, `numEgregious` +1. After insertion both tables hold 146 rows plus the terminating sentinel, are sorted ascending, and have identical subrange boundaries:

```
Teensy     n=15        1K ..      15K
Tiny       n=32       16K ..     240K
Small      n=26      256K ..    2016K
Medium     n=27     2048K ..   16128K
Large      n=18    16384K ..   64512K
Huge       n=18    65536K ..  258048K
Egregious  n=10   262144K ..  524288K
```

## Depends on #207

Worth stating plainly: **on current `main` these entries are inert**, because `main` cannot compute them.

```
main, 1 thread : Res64: 7898B67995506090   MaxErr 0.281    <- correct
main, 2 threads: Roundoff warning iter 13, maxerr 0.4998
main, 4 threads: Roundoff warning iter 12, maxerr 0.4994
```

With #207 (`radix1008/4032: init the DFT-63 'two'/'one' consts in every thread's local store`) applied and nothing else, 2 and 4 threads both return `7898B67995506090` with `MaxErr` back to 0.281 — error metrics identical to the single-threaded run.

I checked that #204 alone does *not* account for it: `main` + #204 reproduces the failure at the same iteration with the same maxerr to nine digits. #207 is the specific dependency.

Until #207 lands, the self-test will mark these lengths "too few usable radix sets" and skip them — harmless, just not yet useful. After it lands they become live.

Two side observations from tracking that down, neither of which changes this PR:

- #207's own description characterises the defect as *wrong Mersenne residues* with more than one carry thread. At 1008K it instead surfaces as *excessive roundoff* — the corrupted per-thread constants push the error past the threshold before a residue is produced. That may be a more recognisable symptom to add to #207 than the silent-wrong-residue one.
- The practical consequence is that leading radix 1008 is currently unusable for real multi-threaded work on 21.0.2, not merely mis-tested.


MarkRose:
Rebased onto current `main` to clear the merge conflict.

One conflict in `src/Mlucas.c`, and it needed care rather than a side-pick: this branch was written when the 10000-iteration column was still `{0x0ull, 0ull, 0ull}` for most rows, and `main` has since filled those in. Taking this branch's side would have wiped out those reference residues; taking `main`'s would have dropped the PR. The resolution keeps every `main` row as-is and splices in only the FFT lengths this PR adds.

Verified after: no existing row removed, no zeroed triplet reintroduced, and both `MersVec[]` and `MvecPRP[]` come to 147 rows against `numTest+1` = 147 — worth checking explicitly, since too *few* initializers would be silently zero-filled and then adopted as new references rather than caught. Build OK, `-s tt` rc=0 with 19 residues.

PR #243, opened 2026-08-01
Flag zero residues, zero roundoff and repeated residues in self-tests and production runs

MarkRose:
Adds four sanity checks for failure modes that neither the roundoff check nor the reference-residue comparison can see. Every one has shown up in this codebase already.

## Production runs — stalled iteration

In the checkpoint loop: a residue matching **any of the last 32 checkpoints** means the iteration is not advancing.

A ring rather than just the previous checkpoint, because a stall need not sit on a fixed point — it can cycle. LL has an absorbing cycle of exactly that shape (`s == 0 → -2 → 2 → 2 …`), and the degenerate `Res64 = ...0002` seen out of defective carry paths *is* that fixed point. Two distinct intervals cannot share a `(Res64, Res35m1, Res36m1)` triple by chance — a 135-bit coincidence.

## Self-tests

#184 adds the equivalent zero-residue guard for production runs and explicitly scopes itself to `!INTERACT`, noting that *"strengthening that pass criterion is a separate matter"*. This is that matter.

**Zero residue.** The comparison structurally cannot catch it — the `(*sh0 != 0)` tests treat a zero reference as "no data yet" and *store* the computed value rather than checking it. So an all-zero result is accepted, and then becomes the expected answer for every later run.

**Zero roundoff.** `fracmax` measures distance from the nearest integer, and zero is exactly an integer, so a transform running on trivial data reports a *perfect* error while computing nothing. radix1008/4032 did precisely this — `Res64 = 0` with a flawless `AvgMaxErr`. Gated on `AME` having actually been accumulated, since `MaxErr == 0.0` is legitimate for the first few tens of iterations while the residue is still a small integer; that is what `AME_ITER_START = 30` exists for.

**Duplicate Res64 across cases**, where a case is (FFT length, exponent, iteration count). This is the self-test analogue of the stall check: a self-test case runs a *single* checkpoint, so it has no earlier interval of its own to compare against. Sibling radix sets of one case are expected to agree and are deliberately not compared against each other.

Most valuable in new-data mode, where no reference row exists to catch it — and that is exactly how the reference tables are generated, so a stall there would be baked in as ground truth permanently.

## Cost

Negligible, and worth being precise since "more conditionals in the inner loop" is the obvious worry. The ring check runs **once per `ITERS_BETWEEN_CHECKPOINTS`** (10000 by default), doing at most 32 short-circuiting triple-compares — in the same block as two O(n) `convert_res_FP_bytewise` passes over the entire residue and a savefile write. A `p=86243` run performs about 9 checkpoints in total.

## Verification

**No false positives:** `Mlucas -s tt` completes rc=0 over 19 cases with none of the four firing. Across a self-test sweep the smallest `AvgMaxErr` at any FFT length, down to 2K, is 0.216 — a real run is nowhere near zero.

**Each check shown to fire**, by injection against the built binary:

| check | control | result |
|---|---|---|
| zero residue | forced all-zero on one case | fires |
| zero roundoff | forced `AME = MME = 0` | fires |
| stalled iteration | frozen residue, real LL run at p=86243 | caught at iteration 20000 |
| stalled iteration, period-2 | alternating residues | caught at 30000 vs 10000 |

That last one is the reason for the ring: a consecutive-only comparison misses it entirely.

One limitation stated plainly: I have **not** demonstrated the cross-case duplicate check firing. In a normal `-s` run it is nearly unreachable, because a case that returns another's residue fails its own reference row first and is skipped before the check runs. Its live window is new-data mode, argued above but not exercised by a control here.

## Depends on

- **#184** — the production-side zero-residue guard. This is its self-test complement; #184 deliberately left that scope open.
- **#216** — without it the self-test always exits 0, so a detected problem still could not fail the run.

### DIFF
--- a/src/Mdata.h
+++ b/src/Mdata.h
@@ -189,7 +189,12 @@ All but the "else" stuff below is specific to factor.c built in standalone mode:
 #define ERR_SKIP_RADIX_SET			14	// In context of self-testing, not fatal for run overall but skip the current set of FFT radices
 #define ERR_INTERRUPT				15	// On one of several interrupt SIGs, exit iteration loop prematurely, write savefiles and exit
 #define ERR_GERBICZ_CHECK			16
-#define ERR_MAX		ERR_GERBICZ_CHECK
+/* v21: sanity checks for failure modes neither the roundoff check nor the reference-residue
+comparison can see. See ernstMain() and the selfTest driver in main(). */
+#define ERR_ZERO_RESIDUE			17
+#define ERR_ZERO_ROUNDOFF			18
+#define ERR_DUPLICATE_RES64			19
+#define ERR_MAX		ERR_DUPLICATE_RES64
 
 /***********************************************************************************************/
 /* Globals. Unless specified otherwise, these are declared in Mdata.h and defined in Mlucas.c: */
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -140,7 +140,10 @@ const char *err_code[ERR_MAX] = {
 	"ERR_UNKNOWN_FATAL",
 	"ERR_SKIP_RADIX_SET",
 	"ERR_INTERRUPT",
-	"ERR_GERBICZ_CHECK"
+	"ERR_GERBICZ_CHECK",
+	"ERR_ZERO_RESIDUE",
+	"ERR_ZERO_ROUNDOFF",
+	"ERR_DUPLICATE_RES64"
 };
 
 // Shift count and auxiliary arrays used to support rotated-residue computations:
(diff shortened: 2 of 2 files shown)


MarkRose:
Follow-up: the stalled-iteration check now compares distinct intervals only.

It rests on "distinct intervals cannot repeat a residue by chance", but compared the current residue against every history entry - including the one for the current iteration. The same interval reaches the checkpoint block twice when an interrupt arrives *between* intervals - a signal during the checkpoint write, or the run flag being cleared - because the loop goes back through it to write the savefile at the iteration just completed. The second pass finds its own history entry:

```
 ***   Stalled-iteration Error   ***
 Res64 2D0978C0DA521DE2 at iteration 50000 is identical to that at iteration 50000.
 Distinct intervals cannot repeat a residue by chance; the iteration is not advancing.
```

Same iteration number on both sides - the residue is being compared with itself. The savefile was written correctly and the run resumed from it without trouble, so the damage was a diagnostic telling the user to go looking for hardware errors after an ordinary ^C, plus exit 134 where a clean stop should give 0.

Found by `tests/jacobi-check.sh` T7, whose whole subject is an interrupt between intervals; the suite goes from 57 passed / 1 failed to 58 passed / 0 failed.

PR #251, opened 2026-08-04
Added support for FreeBSD

tdulcet:
* Added support for FreeBSD to the `makemake.sh` script and generated Makefile
* Updated the CI
  * Added FreeBSD job using the [vmactions/freebsd-vm](https://github.com/marketplace/actions/freebsd-vm) Action
  * Updated macOS jobs to use the newer macOS 26 Intel runner
* Minor cleanups from #67
  * Made requirement for C99 consistent everywhere

This added error masking to the new FreeBSD CI job, which would for example need to be removed in #103 or a follow up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,10 +191,10 @@ jobs:
             # KNL-only via __AVX512ER__, so plain AVX-512 builds no longer need an ER-capable assembler -
             # which is why clang 5.0 on alpine:3.8 can now build AVX-512.) Probe and skip AVX-512 rather than
             # fail the build on a mode that provably can't work here.
-            if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm31,%%zmm31,%%zmm31\n\tvmovdqa64 %%xmm16,%%xmm17\n\tkmovw %%k1,%%eax":::"zmm31","xmm16","xmm17","k1","eax");return 0;}' | "${CC:-gcc}" -mavx512f -x c -o /dev/null - >/dev/null 2>&1; then
+            if echo 'int main(void){__asm__ __volatile__("vpxord %%zmm31,%%zmm31,%%zmm31\n\tvmovdqa64 %%xmm16,%%xmm17\n\tkmovw %%k1,%%eax":::"zmm31","xmm16","xmm17","k1","eax");return 0;}' | $CC -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma -x c -o /dev/null -; then
                 MODES+=( AVX512 )
             else
-                echo "::warning title=AVX-512 build skipped::$("${CC:-gcc}" --version | head -n1): its assembler can't handle the AVX-512 extended registers (zmm16-31/xmm16-31/k0-7) Mlucas's kernels need; skipping the AVX-512 build mode."
+                echo "::warning::Skipping the AVX512 build mode as the '$($CC --version | head -n1)' assembler cannot handle the required AVX512 extended registers (zmm16-31/xmm16-31/k0-7)."
             fi
         fi
         for mode in "${MODES[@]}"; do
@@ -229,7 +229,7 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=address,undefined" >> "$GITHUB_ENV"
+        sed -i 's/-O3 -flto/-Og -fsanitize=address,undefined/' makemake.sh
         $CC --version
     - name: Script
       run: |
@@ -261,7 +261,7 @@ jobs:
         sudo apt-get install -y libhwloc-dev
     - name: Before script
       run: |
-        echo "CFLAGS=-fdiagnostics-color -Wall -g -Og -fsanitize=thread" >> "$GITHUB_ENV"
+        sed -i 's/-O3 -flto/-Og -fsanitize=thread/' makemake.sh
         $CC --version
     - name: Script
       run: |
@@ -282,7 +282,7 @@ jobs:
       run: |
         set -x
         set -o pipefail
-        gcc -c -fdiagnostics-color -g -O3 -march=native -DUSE_THREADS -fanalyzer src/*.c |& tee analyzer.log
+        gcc -c -fdiagnostics-color -std=gnu99 -g -O3 -march=native -DUSE_THREADS -fanalyzer src/*.c |& tee analyzer.log
         rm -- *.o
         echo -e '## GCC analyzer\n```' >> $GITHUB_STEP_SUMMARY
         grep 'warning:' analyzer.log | sed 's/\x1B\[\([0-9]\+\(;[0-9]\+\)*\)\?[mK]//g' | awk '{ print $NF }' | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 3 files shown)


makemake.sh
```diff
@@ -312,32 +310,24 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
-	# MacOS: sysctl -n prints nothing (and exits nonzero) for an absent key - e.g. all the
-	# hw.optional.avx* keys on Apple Silicon - so capture each with a 0 default. This also lets the
-	# combined AVX-512||AVX2 test below use plain arithmetic without an empty operand tripping
-	# '((: || : syntax error' (as `(( || $(...) ))` would when the first sysctl yields nothing):
-	avx512f=$(sysctl -n hw.optional.avx512f 2>/dev/null || echo 0)
-	avx2_0=$( sysctl -n hw.optional.avx2_0  2>/dev/null || echo 0)
-	avx1_0=$( sysctl -n hw.optional.avx1_0  2>/dev/null || echo 0)
-	sse2=$(   sysctl -n hw.optional.sse2    2>/dev/null || echo 0)
-	neon=$(   sysctl -n hw.optional.neon    2>/dev/null || echo 0)
+	avx512f=$(sysctl -n hw.optional.avx512f)
```
MarkRose, line 324: This looks like a genuine regression rather than a rebase artifact — I checked, and the guarded form is present at this PR's own merge-base (`4a21413a`), so the change actively removes it.

`main` has:

```bash
avx512f=$(sysctl -n hw.optional.avx512f 2>/dev/null || echo 0)
```

with a comment explaining that `sysctl -n` prints nothing *and exits nonzero* for an absent key — which is every `hw.optional.avx*` key on Apple Silicon.

The consequence is worse than a wrong answer. `makemake.sh` sets `set -e` at line 30, so the nonzero status from the assignment terminates the script — silently, with no output:

```
$ bash -c 'set -e; sysctl(){ return 1; }; v=$(sysctl -n hw.optional.avx512f); echo reached'
$ echo $?
1          # "reached" never prints
```

So `./makemake.sh` with no arguments would just exit 1 on an Apple Silicon Mac. The guarded form falls through correctly to the next branch.

For what it's worth I also checked the related `(($(sysctl -n hw.optional.avx2_0)))` change on the line below, and that one is harmless — an empty operand there evaluates false rather than erroring.

makemake.sh
```diff
@@ -264,13 +269,6 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 			ARGS+=(-DUSE_AVX512 -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
 			;;
 		k1om)
```
MarkRose, line 282: Could the cross-build note above this arm stay? It documented that the MPSS SDK's `environment-setup-k1om-mpss-linux` exports its own `CFLAGS`/`CPPFLAGS`, and that the Makefile's `CFLAGS ?=` therefore silently drops `-O3` and `-D_GNU_SOURCE`.

That matters because it produces two failures that look like k1om defects but aren't: `impossible constraint in 'asm'` in `radix16_dif_dit_pass_asm.h` (the `"e"` constraint on `pfetch_dist` needs the optimiser), and `threadpool.c` losing `CPU_ZERO`/`sched_setaffinity` (the missing `-D_GNU_SOURCE`). It took a while to work out the first time.

MarkRose (changes requested):
FreeBSD support is a welcome addition, and the `vmactions/freebsd-vm` approach looks clean. Requesting changes for one item only — the macOS sysctl change, which I think breaks auto-detect on Apple Silicon outright. Details inline.

Two smaller things:

**Rebase.** This doesn't currently merge onto `main` — `makemake.sh` conflicts, and the branch is 48 commits behind. I suspect the conflict is largely the same lines flagged inline, so a rebase may resolve them together.

**Agreed on the masking.**

> This added error masking to the new FreeBSD CI job, which would for example need to be removed in #103 or a follow up.

Agreed. #103 removes the `|| true` from every `makemake.sh mfac` call, so if that lands first this job's would be the only mask left. Happy to take it out there or leave it as a follow-up, whichever order suits you.

makemake.sh
tdulcet: This was just a cleanup from #67. That comment added in #67 was wrong, which is why I removed it. The CI confirms that this works as expected on ARM macOS. Note that this whole section will be removed in #254, so it is somewhat a moot point.

makemake.sh
tdulcet: This PR removed the `CFLAGS ?=`, as it broke BSD make as well, so that comment is no longer needed.

.github/workflows/ci.yml
```diff
@@ -494,6 +494,76 @@ jobs:
...
+            for word in '' 1word 2word 3word 4word nword; do
+                echo -e "\nMfactor $word\n"
+                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
+                (
+                    cd obj_mfac${word:+_$word}${mode:+_$mode}
+                    make clean
+                    [ ! -x Mfactor${word:+_$word} ] || ./Mfactor${word:+_$word} -h
+                )
+            done
+        done
```
MarkRose, line 557: All four FreeBSD jobs are currently failing here, and it's this line:

```
+ [ ! -x Mfactor ]
+ ./Mfactor -h
Illegal instruction
##[error]Process completed with exit code 132.
```

It's the `AVX512` iteration. The `sysctl -a` output above in the same log has the reason — the host is an EPYC 9V74, which is Zen 4 and does have AVX-512, but the hypervisor masks it out of the guest:

```
hw.model: AMD EPYC 9V74 80-Core Processor
Structured Extended Features=0x219c07ab<FSGSBASE,TSCADJ,BMI1,AVX2,SMEP,BMI2,ERMS,
                                        INVPCID,RDSEED,ADX,SMAP,CLFLUSHOPT,CLWB,SHA>
```

No `AVX512*` in leaf 7, so the binary is right to trap. Because the step runs under `set -ex`, the job stops here — which costs the parts I'd most want from a FreeBSD runner, since `-s tt t s m` and `config-fermat.sh` are both below the loop and never execute.

Gating execution rather than the build keeps compile coverage for all six modes:

```suggestion
        set -ex -o pipefail
        sysctl -a
        # CPUID is masked in the VM, so build every mode but only run what the guest can execute
        native=$($CC -dM -E -march=native - </dev/null)
        runnable() {
            case $1 in
                ''|NOSIMD) return 0 ;;
                SSE2)      echo "$native" | grep -q __SSE2__ ;;
                AVX)       echo "$native" | grep -q __AVX__ ;;
                AVX2)      echo "$native" | grep -q __AVX2__ ;;
                AVX512)    echo "$native" | grep -q __AVX512F__ ;;
            esac
        }
        for mode in '' NOSIMD SSE2 AVX AVX2 AVX512; do
          echo -e "\n$mode\n"
            bash -e -o pipefail -- makemake.sh use_hwloc $mode
            (cd obj${mode:+_$mode}; make clean)
            for word in '' 1word 2word 3word 4word nword; do
                echo -e "\nMfactor $word\n"
                bash -e -o pipefail -- makemake.sh mfac $word $mode || true
                (
                    cd obj_mfac${word:+_$word}${mode:+_$mode}
                    make clean
                    if [ -x Mfactor${word:+_$word} ] && runnable "$mode"; then
                        ./Mfactor${word:+_$word} -h
                    fi
                )
            done
        done
```

The subshell inherits both the function and `$native`, and it's all POSIX `sh`, so it should be fine under `usesh: true`.

MarkRose:
On the `sysctl` guard — you're right and I withdraw it. `MODES=( '' NOSIMD )` does exercise auto-detect on macOS, and `macos-14`, `macos-15` and `macos-26` were all green, so an absent `hw.optional.avx512f` evidently isn't fatal on Apple Silicon. I'd built that objection from a simulated failure and never checked it against a real Mac. Sorry for the noise.

`CFLAGS ?=` — confirmed removed, the only remaining mention is in a comment, so the k1om note is unnecessary now. Both of my original review points are resolved.

The change request stands, but for a different reason: all four FreeBSD jobs are red, and have been on every push. Details and a suggested fix are inline — briefly, the `AVX512` iteration runs `./Mfactor -h` and traps, because the hypervisor masks AVX-512 out of the guest even though the EPYC 9V74 host has it. Under `set -ex` the job stops there, so `-s tt t s m` and `config-fermat.sh` never run.

One thing the logs give us for free: `mfac 2word/3word/4word` fail on every release with

```
../src/factor.h:105:4: error: P2WORD may not be used together with USE_FMADD!
```

That's the `#error` I'm removing in #83 — it fires on any FMA-capable host, so those three word variants can't build in native or AVX2 mode anywhere. Here it's absorbed by the `|| true`, but it's worth knowing that's what the mask is currently hiding.

MarkRose (approved):
Approving — both of the items I originally requested changes for are settled: the `sysctl` guard was my mistake, and the `CFLAGS ?=` removal makes the k1om note unnecessary.

The AVX-512 `Illegal instruction` in the mode loop is worth fixing, but I don't think it should hold this up. It's confined to the new job, nothing else depends on it, and the board is red for unrelated reasons anyway. The suggestion is inline whenever you want it — until then the FreeBSD jobs give compile coverage but stop before the self-tests.

.github/workflows/ci.yml
tdulcet: It is not just the FreeBSD jobs that are failing, it is all of the x86 jobs (Linux, macOS and Windows), but they are also filing on the `main` branch with this same error. This should be fixed by #103. The #238 integration branch confirms that this works as expected after the changes in #103.

PR #254, opened 2026-08-08
Updated makemake.sh script to use CPUID to detect SIMD build modes

tdulcet:
* Updated the `makemake.sh` script to use the existing `get_cpuid.c` file to detect the supported SIMD build modes on x86
  * This replaces the separate Linux, macOS and Windows code
  * Also, added support for detecting K1OM IMCI512 and KNL AVX512 build modes
* Removed the AVX512 DQ, BW and VL extensions added in #64, as they are not actually used, just discussed in comments

This depends on #164.

### DIFF
--- a/makemake.sh
+++ b/makemake.sh
@@ -268,7 +268,7 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 		avx512_skylake)
 			echo "Building for avx512_skylake SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
 			echo "Warning: The 'avx512_skylake' option is deprecated, use 'avx512' instead."
-			ARGS+=(-DUSE_AVX512 -march=skylake-avx512 -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
+			ARGS+=(-DUSE_AVX512 -march=skylake-avx512 -mavx512f -mavx512cd -mfma)
 			;;
 		avx512_knl)
 			echo "Building for avx512_knl SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
@@ -277,10 +277,10 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
 			;;
 		avx512)
 			echo "Building for AVX512 SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
-			ARGS+=(-DUSE_AVX512 -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl -mfma)
+			ARGS+=(-DUSE_AVX512 -mavx512f -mavx512cd -mfma)
 			;;
 		k1om)
-			echo "Building for 1st-gen Xeon Phi 512-bit SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
+			echo "Building for 1st-gen Xeon Phi 512-bit IMCI512 SIMD in directory '${DIR}_${arg}'; the executable will be named '${TARGET}'"
 			ARGS+=(-DUSE_IMCI512)
 			;;
 		avx2)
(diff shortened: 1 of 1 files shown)


MarkRose (commented):
Nice consolidation — three platform-specific detectors collapsing into one probe, and it reads much better. Functionally verified here on an i9-10885H (AVX2, no AVX-512): auto-detect correctly picks AVX2 and the build completes, rc=0.

Three things I'd flag, one of which I think is a real gap.

## 1. Can the k1om auto-detection actually work?

The probe is compiled with `$CC` and then **run on the build host**:

```bash
"${CC:-gcc}" -std=gnu99 -Wall -g -O3 -Isrc -o "$tmpdir/simd" "$tmpdir/simd.c"
if ! output=$("$tmpdir/simd"); then
```

But the k1om arm is selected by `#if defined(CPU_IS_K1OM)`, which is only true when `$CC` already targets KNC — i.e. exactly the case where the resulting binary can't execute on an x86 build host. It would fail at the `$(...)`, hit `Error: Unable to detect the SIMD build mode`, and exit 1.

Am I missing something — is the expectation that you run `makemake.sh` on the card itself, or does MPSS provide a host-side runner? Otherwise a KNC cross-build has to pass `k1om` explicitly, which works fine, and the auto-detection for it can't trigger.

Worth noting the same reasoning applies to any cross-compile: the probe assumes host-executable output.

## 2. The auto-detected k1om arm adds `-march=native`, the explicit one doesn't

```
explicit  k1om)   ARGS+=(-DUSE_IMCI512)
auto      k1om)   ARGS+=(-DUSE_IMCI512 -march=native)
```

`main` has just `-DUSE_IMCI512` for the explicit mode, and the cross-build note above it explains why the SDK's environment is delicate about flags. For a cross-compiler targeting KNC, `-march=native` seems at best redundant. Deliberate?

## 3. The Core2Duo workaround is dropped

The macOS SSE2 branch had:

```bash
# On my Core2Duo Mac, 'native' gives "error: bad value for -march= switch":
ARGS+=(-DUSE_SSE2 -march=core2 -msse2)
```

and the new unified `sse2)` arm uses `-march=native -msse2`, so that machine would now fail to build. It's a narrow case, but the comment reads like it was written from a real failure — perhaps keep `-march=core2` for the SSE2 arm, or wrap it in `try_flag -march=native`, as the `asimd` and `nosimd` arms already do.

## Minor

- `// if (has_imci512())` and `// if (has_asimd())` are left in the generated source as commented-out scaffolding.
- The probe compile isn't guarded, but `set -e` is in force at line 30, so a compile failure exits with the compiler's own error rather than reaching the `Unable to detect` message. That's arguably the better outcome; just noting the message is only reachable on the run-fails path, which is case 1 above.

tdulcet:
> Am I missing something — is the expectation that you run `makemake.sh` on the card itself, or does MPSS provide a host-side runner?

Users would just need to select the MPSS compiler and I believe it should work as expected:
```bash
CC=k1om-mpss-linux-gcc bash makemake.sh
```
This reuses the existing K1OM detection in Mlucas, which #155 already confirms that it works.

> For a cross-compiler targeting KNC, `-march=native` seems at best redundant. Deliberate?

This was just for consistency with the other native builds, which all supply `-march=native` or `-mcpu=native`. I am not sure if it is redundant per say, but feel free to remove it if you prefer.

> `// if (has_imci512())` and `// if (has_asimd())` are left in the generated source as commented-out scaffolding

Yes, this was intentional, as I would like to use those functions, but `has_imci512()` is not fully implemented, while `has_asimd()` is defined in `utils.c`, which currently has dependencies on a large subset of Mlucas, and would require a larger refactoring to be able to use independently (as the `get_cpuid.c` file can already be).

> The probe compile isn't guarded, but `set -e` is in force at line 30, so a compile failure exits with the compiler's own error rather than reaching the `Unable to detect` message.

Feel free to move this into that below if statement, if you prefer. We may also want to hide the compiler warnings by default.

makemake.sh
```diff
@@ -321,132 +321,66 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
 	tmpdir=$(mktemp -d)
 	trap 'rm -rf "$tmpdir"' EXIT
 	cat <<'EOF' >"$tmpdir/simd.c"
 #include <stdio.h>
-int main()
+
+#if !defined(__k1om__) && (defined(__i386__) || defined(__x86_64__))
+	#define USE_AVX
+#endif
+#include "get_cpuid.c"
```
ldesnogu, line 332: I feel uneasy at having to force USE_AVX and then include `get_cpuid.c`.
Why is it so? Couldn't it be passed when invoking the compiler (cf my comment below)?

Will the whole script correctly detect IMCI512?

makemake.sh
```diff
@@ -321,132 +321,66 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
-	case $HOSTTYPE in
-		aarch64 | arm*)
-			args+=(-mcpu=native)
-			;;
-		x86_64 | *)
-			args+=(-march=native)
-			;;
-	esac
-	"${CC:-gcc}" -Wall -g -O3 "${args[@]}" -o "$tmpdir/simd" "$tmpdir/simd.c"
+	"${CC:-gcc}" -std=gnu99 -Wall -g -O3 -Isrc -o "$tmpdir/simd" "$tmpdir/simd.c"
```
ldesnogu, line 381: Is it guaranteed that without any further compilation flag the compile-time defined stuff in `platform.h` will be correct?

ldesnogu (commented):
I think I don't have the full picture here, but I'm unsure the auto-detection will work for all hosts (for x86, Arm looks OK).

makemake.sh
tdulcet: > I feel uneasy at having to force USE_AVX and then include `get_cpuid.c`.
> Why is it so? Couldn't it be passed when invoking the compiler (cf my comment below)?

The problem is that the AVX, AVX2 and AVX512 checks are disabled in `get_cpuid.c` without this `USE_AVX` define. There is a comment from Ernst that explains why: https://github.com/primesearch/Mlucas/blob/fe0f0d014c376bd09916e76011c8a9cf1dda6133/src/get_cpuid.c#L456-L460 I am not certain that this avoids the compiler error on ancient systems, but I do not have any such pre-AVX systems to test this on. There should be a better way to fix this, as for example Prime95 does not do this, but it may require a larger refactor of `get_cpuid.c`.

We would not want to pass this define to the compiler with `-DUSE_AVX`, as it should only be enabled on x86 systems.

> Will the whole script correctly detect IMCI512?

Yes, it should. See https://github.com/primesearch/Mlucas/pull/254#issuecomment-5252567997 above.

makemake.sh
tdulcet: Yes, `platform.h` does not require any additional flags. It just relies on the defines that compilers enable by default. The `-march=native`/`-mcpu=native` was only needed because this `simd.c` program previously used special defines like `__AVX512F__`, but now with CPUID those are no longer required. The CI confirms that this works as expected on all 60+ tested combinations.

ldesnogu (dismissed):
This looks good to me.

Note a further improvement could be to factor the settings of `ARGS` no matter whether the target was selected by command-line or auto-detected (differences are only in the setting of `-march` for the various 512-bit derivatives). But even if that's doable or worth it, this shouldn't prevent this change from being committed.

makemake.sh
```diff
@@ -321,132 +321,66 @@ if [[ ${#MODES[*]} -eq 1 ]]; then
...
 	if ! output=$("$tmpdir/simd"); then
 		echo "Error: Unable to detect the SIMD build mode" >&2
 		exit 1
```
MarkRose, line 385: Following up on the k1om question — I built the MPSS cross-toolchain in a container and ran the documented invocation against this branch:

```
$ CC=k1om-mpss-linux-gcc bash makemake.sh
src/platform.h:451:3: warning: #warning "Xeon Phi (1st-Gen) detected"
makemake.sh: line 384: /tmp/tmp.ir2olzxjig/simd: cannot execute binary file
Error: Unable to detect the SIMD build mode
```

The probe compiles correctly — `platform.h` even identifies the target — but this line *runs* it, and a K1OM ELF can't execute on the x86-64 build host, so the mode is never detected.

For the same reason I don't think #155 confirms the path: its K1OM job passes the mode explicitly (`makemake.sh no_gmp k1om`), which exercises the `k1om)` case arm rather than auto-detection.

The cross-compiler predefines `__k1om__`, which is enough to resolve it without running anything:

```suggestion
	# Cross-compilers produce a probe the build host cannot execute, so resolve those
	# targets from the compiler's own predefined macros before falling back to running it.
	if "${CC:-gcc}" -dM -E -x c /dev/null 2>/dev/null | grep -q '__k1om__'; then
		output=k1om
	else
		"${CC:-gcc}" -std=gnu99 -Wall -g -O3 -Isrc -o "$tmpdir/simd" "$tmpdir/simd.c"
		if ! output=$("$tmpdir/simd"); then
			echo "Error: Unable to detect the SIMD build mode (the probe did not run - cross-compiling? Pass the mode explicitly, e.g. 'k1om')" >&2
			exit 1
		fi
```

Applied exactly as suggested and re-run in the container, it prints **"The CPU supports the IMCI512 SIMD build mode"** and proceeds with `-DUSE_IMCI512`. (The build then hits the K1OM compile errors that #155 is about, which is a separate matter.)

The improved message is worth having on its own — "Unable to detect the SIMD build mode" gives no hint that cross-compiling is the cause.

Everything else looks good to me. I checked the reduced `avx512` flag set separately: no DQ/BW/VL-only instructions exist anywhere in the tree, and the whole thing builds clean with just `-mavx512f -mavx512cd -mfma`.

tdulcet: > The probe compiles correctly — `platform.h` even identifies the target — but this line _runs_ it, and a K1OM ELF can't execute on the x86-64 build host, so the mode is never detected.

Yes, you would of course need an actual Knights Corner/K1OM card to be able to test this. It is like trying to run CUDA code without an Nvidia GPU installed. However, if the user has a K1OM card installed and properly setup, unless I am missing something, this should work as expected. The native builds are for compiling on the same system that it is intended to be run on, so for K1OM this would require a K1OM card. Of course, we will not know for 100% sure that this works until someone with a K1OM card is able to test it.

Users that are cross compiling for K1OM should continue to use the explicit `k1om` option to the `makemake.sh` script, as #155 does.

MarkRose (approved):
Withdrawing my k1om suggestion — your framing is right, and the arm you added is a better fix than the one I proposed.

```c
#if defined(CPU_IS_K1OM)
	puts("k1om");
```

together with `#if !defined(__k1om__) && (defined(__i386__) || defined(__x86_64__))` means a genuine on-card native build resolves without needing to run anything CPUID-dependent, which is the case auto-detect is actually for. Cross-compiling to the explicit `k1om` option, as #155 does, is the right split. My reproduction was of the `CC=k1om-mpss-linux-gcc bash makemake.sh` invocation, and that path is now explicitly out of scope rather than broken.

One optional nicety, take it or leave it: when the probe compiles but cannot execute, the failure surfaces as

```
makemake.sh: line 384: /tmp/tmp.XXXX/simd: cannot execute binary file
Error: Unable to detect the SIMD build mode
```

which gives no hint that cross-compiling is the cause. Naming it in the message would save the next person the detour. Not worth another round on its own.

Approving — your rebase dismissed the earlier approval, and everything I raised is resolved. I separately checked the reduced `avx512` flag set: no DQ/BW/VL-only instructions exist anywhere in the tree, and the whole thing builds clean with just `-mavx512f -mavx512cd -mfma`. The detection change is equivalent in practice too, since the only F+CD parts without DQ/BW/VL are KNL/KNM, which the reordered ER branch now catches first.

tdulcet:
Thanks for the thorough review.

> gives no hint that cross-compiling is the cause. Naming it in the message would save the next person the detour.

Please feel free to update the error message if you believe it is needed. I am not sure we would want to mention cross compiling specifically, as this is the native code path, so that could be confusing for users, but I am sure we could still improve this error message.

PR #257, opened 2026-08-19
Mfactor: don't compile the AVX-512 vector sieve for IMCI-512 (KNC)

MarkRose:
`platform.h:263` makes `USE_IMCI512` imply `USE_AVX512`, so the vectorized small-primes sieve in `PerPass_tfSieve()` is compiled for Knights Corner — which has IMCI-512, not AVX-512F, and none of the gather/scatter or mask instructions that block uses.

Building `Mfactor` for K1OM therefore fails outright:

```
../src/factor.c:3212: Error: operand type mismatch for `vpbroadcastd'
../src/factor.c:3217: Error: `vmovups' is not supported on `k1om'
../src/factor.c:3223: Error: invalid instruction suffix for `kmov'
../src/factor.c:3229: Error: no such instruction: `vprolvd %zmm0,%zmm31,%zmm1{%k1}'
make: *** [Mfactor] Error 1
```

This adds a dedicated `USE_AVX512_SIEVE` flag in `factor.h`, set only when `USE_AVX512` is defined and `USE_IMCI512` is not, and keys the sieve on it. `mi64.c` already uses the same `defined(USE_AVX512) && !defined(USE_IMCI512)` idiom at lines 2414 and 4221; a named flag is used here because the carve-out is needed at ten sites in `factor.c` and two in `factor.h` — including the `psmall` parameter in `PerPass_tfSieve()`'s prototype and definition, which have to move together or the build breaks a different way.

KNC falls back to the scalar bit-clearing loop, which is a live path regardless: the AVX-512 arm defers to it for small exponents.

## Verified

**KNC, MPSS 3.8.6 cross-toolchain in a container.** Before, all six `Mfactor` word variants fail at `factor.c:3212`. After:

```
  base   BUILT   e_machine=181 (K1OM)
  1word  BUILT   e_machine=181 (K1OM)
  2word  BUILT   e_machine=181 (K1OM)
  3word  BUILT   e_machine=181 (K1OM)
  4word  BUILT   e_machine=181 (K1OM)
  nword  BUILT   e_machine=181 (K1OM)
```

Real Knights Corner binaries, not host objects — ELF `e_machine` 181.

**x86 AVX-512 is untouched.** Preprocessed `factor.c` for an AVX-512 build is identical before and after, once `# line` directives are discounted (the patch adds ten lines to `factor.h`, which shifts them). Checked against a negative control that disables the new flag and does show a difference, so the comparison is capable of failing.

## Why it matters now

#155 adds a K1OM CI job whose `Mfactor` builds are wrapped in `|| true`, so the job currently reports success while producing no binary at all. With this in, that mask can come off and the job will actually be testing something.

Not a regression — this has been the state since `USE_IMCI512` began implying `USE_AVX512`. It only surfaces once anything tries to build `Mfactor` for K1OM.

### DIFF
--- a/src/factor.h
+++ b/src/factor.h
@@ -28,6 +28,14 @@
 
 #include "util.h"
 
+/* IMCI-512 (1st-gen Xeon Phi / Knights Corner) shares AVX-512's zmm/k register names, and platform.h
+makes USE_IMCI512 imply USE_AVX512 - but it has none of the instructions the vectorized small-primes
+sieve in factor.c needs: vprolvd, vmovups, kmovw and the register form of vpbroadcastd all fail to
+assemble for k1om. That sieve - and the psmall[] table which exists solely to feed it - is therefore
+gated on 'defined(USE_AVX512) && !defined(USE_IMCI512)' rather than on USE_AVX512 alone, the same
+idiom mi64.c already uses, so KNC falls back to the scalar bit-clearing loop (a live path in any
+case: the AVX-512 arm defers to it for small exponents). */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -335,7 +343,7 @@ uint32	CHECK_PKMOD4620(uint64 *p, uint32 lenP, uint64 k, uint32*incr);
 		const uint32 p_last_small,	//largest odd prime appearing in the product; that is, the (nclear)th odd prime.
 		const uint32 nprime,		// #sieving primes (counting from 3)
 		const uint32 MAX_SIEVING_PRIME,
-	#if defined(USE_AVX512) && !defined(USE_GPU)
+	#if defined(USE_AVX512) && !defined(USE_IMCI512) && !defined(USE_GPU)
 		const uint32 *psmall,
 	#endif
 		const uint8 *pdiff,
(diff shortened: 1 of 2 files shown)


src/factor.h
```diff
@@ -28,6 +28,16 @@
...
 #include "util.h"
 
+/* IMCI-512 (1st-gen Xeon Phi / Knights Corner) shares AVX-512's zmm/k register names, and platform.h
+makes USE_IMCI512 imply USE_AVX512 - but it has none of the instructions the vectorized small-primes
+sieve needs: vprolvd, vmovups, kmovw and the register form of vpbroadcastd all fail to assemble for
+k1om. Gate that sieve - and the psmall[] table which exists solely to feed it - on a separate macro,
+so KNC falls back to the scalar bit-clearing loop (which is a live path in any case: the AVX-512 arm
+defers to it for small exponents). */
+#if defined(USE_AVX512) && !defined(USE_IMCI512)
+	#define USE_AVX512_SIEVE
```
tdulcet: I would prefer not to introduce another define here, as there are already so many in Mlucas/Mfactor, and in this case we are not saving very many characters.

tdulcet (commented):
Thanks for fixing this. I just have a nit.

tdulcet (approved):
Thanks.

PR #263, opened 2026-09-07
CI: run the ASan Linux job on ARM too, instrumenting the ASIMD build

MarkRose:
Right now **no job anywhere builds hand-tuned SIMD asm under a sanitizer.** Every sanitizer job is pinned to NOSIMD, because macOS cannot assemble the SIMD under `-Og` (`invalid CFI advance_loc expression`). That pinning is correct for macOS, but it gave up the one configuration that has actually caught a SIMD memory bug in this codebase.

Issue #12 - a SEGV in `radix48_dit_pass1` during `Mlucas -s m` - was reported by AddressSanitizer on an ARM runner. As things stand there is nothing left in CI that could find its like.

This is the ARM counterpart to #221, which covers the x86 side (AVX2 natively, AVX-512 under SDE). The two do not overlap.

## What it does

Adds `ubuntu-24.04-arm` to the existing `ASan-Linux` matrix rather than adding a job. The build mode comes from `$HOSTTYPE`, the way the `Linux` and `Singlethread-Linux` jobs already select theirs:

* **x86 cells** keep the NOSIMD build and behave exactly as before.
* **ARM cells** let `makemake.sh` auto-detect, which selects ASIMD - the build this exists to instrument.
* On ARM, `config-fermat.sh` and the Mfactor self-tests run from the NOSIMD builds. That is the same substitution the `Linux` and `Singlethread-Linux` jobs make: ASIMD's trial-factoring routines are stubs (`No TF support on ARMv8!`, under `#ifdef USE_ARM_V8_SIMD` in `twopmodq80.c`).
* Same tiers as the x86 cells - `tt t s m` plus PRP `tt t s`. My earlier revision gated the medium tier; that is gone, since ASIMD is far faster per iteration than the NOSIMD build the x86 cells already run `-s m` on.
* Adds the artifact upload the sibling Linux jobs have.

**It asserts the binary is really ASIMD.** `makemake.sh` announces the selected mode on stdout but does not record it in `obj/build.log`, so a check against the build log would silently pass while testing nothing - I wrote that version first and it would have failed the job for the wrong reason. The job greps the run itself for `Build uses ARMv8 advanced-SIMD`, which `util.c` prints only under `#ifdef USE_ARM_V8_SIMD` (the near-miss line for a scalar build on ASIMD-capable hardware reads `CPU supports ARMv8 advanced-SIMD ... but using scalar floating-point build`, and does not match). Without this the ARM cells quietly degrade into duplicates of the x86 ones the moment auto-detection changes.

## Verified on aarch64 hardware

Debian 12, gcc 12.2 and clang 14, using the flags `makemake.sh` generates:

* **gcc**: full native (ASIMD) build with `-Og -fsanitize=address,undefined` compiles, links and self-tests clean.
* **clang**: the SIMD-heaviest translation units - `radix48_ditN_cy_dif1.c` (the one from #12), `radix32_wrapper_square.c`, `radix32_dif_dit_pass.c`, `radix16_dif_dit_pass.c` - all assemble with zero diagnostics under `-mcpu=native -Og -fsanitize=address,undefined`. So the macOS assembler problem does not reproduce on ARM Linux with either compiler.

Caveat worth stating: that box runs clang 14, while `ubuntu-24.04-arm` has a much newer one. The `Linux` job already builds ASIMD with clang on those runners without sanitizers, so the only new variable is `-Og -fsanitize=...`, but this PR's own CI run is the real check.

## On #12 itself

Separately I tried to reproduce #12 on aarch64 hardware against current `main`, and it does **not** reproduce: the exact reported configuration passes, including under ASan, and the `-s m` sweep reached `M57984131` at 3072K with radices 48/32/32/32 and passed it. I would not call the issue fixed on that basis - different platform (Debian/gcc vs macOS/clang), different microarchitecture, and the report says "sometimes" - but it is worth recording. This job is what would tell us either way, on an ongoing basis, rather than by my hand-running it once.

## Dropped from the earlier revision

The previous version also failed the job on any `WARNING: 0 of` line (an FFT length with no passing radix set) and on any `runtime error:` from UBSan, on the grounds that the self-test's exit status is not sufficient on its own until #216 lands, and that an ASan report is not sufficient either since a register clobber corrupts results without making a bad memory access. I still think both are worth having, but they are new failure conditions on cells that are green today, so they belong in their own PR rather than riding along on this one. Happy to open that separately.

CI only; no source changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,9 +314,10 @@ jobs:
   ASan-Linux:
     name: AddressSanitizer Linux
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, ubuntu-24.04-arm]
         cc: [gcc, clang]
       fail-fast: false
     env:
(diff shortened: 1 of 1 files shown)


tdulcet (changes requested):
Thanks for the PR.

Instead of adding a new CI job for this, could you instead update the matrix for the existing ASan Linux job to also run on a ARM runner. That would eliminate the duplication.

Issue #97, opened 2026-07-13
Add support for the Jacobi error check for LL and P-1 tests


tdulcet:
Prime95/MPrime and GpuOwl support the Jacobi error check for LL tests, which can detect 50% of errors at relatively low cost. GpuOwl additionally supports this for P-1 tests. The Jacobi error check is not as robust as the Gerbicz error check (GEC) for PRP tests (which is already supported), but it is much easier to implement and could provide some error checking for LL and P-1 tests, which currently have none.

Both Prime95/MPrime and GpuOwl use the GNU MP (GMP) library to compute the [Jacobi symbol](https://en.wikipedia.org/wiki/Jacobi_symbol). Mlucas already has a optional dependency on GMP, so we should update it to support the Jacobi error check when built with GMP.

For reference, [Prime95/MPrime has these options](https://github.com/primesearch/Prime95/blob/dd3c130a04481a793ce608803825842abe80f67e/undoc.txt#L98-L118) to control the Jacobi error check. The [GIMPS status](https://github.com/tdulcet/AutoPrimeNet/blob/main/gimps_status.py) script can already perform the Jacobi error check on Mlucas LL savefiles when using the `--jacobi` option, and will also use the GMP library when available.

Ernst's TODO list: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/docs/todo.txt#L938 and: https://github.com/primesearch/Mlucas/blob/1f6068104d4518348ae3f3f265a48e9ca54a629d/docs/todo.txt#L941-L942

PR #264, opened 2026-09-07
Add the Jacobi residue check for LL tests (part of #97)

MarkRose:
Part of #97 - the LL half. The P-1 half (a Gerbicz check for stage 1, exact checks of the
stage 2 inputs) follows in a PR stacked on this one.

**What it checks.** For an LL test of M(p), p an odd prime, every iterate satisfies
J(s_n - 2 | M(p)) = -1 whatever M(p)'s primality (the symbol telescopes to J(3 | M(p))). A
corrupted residue satisfies it with probability 1/2. The symbol is also invariant under the
recurrence from one iteration after a corruption on, so every later check examines the same
value: a corruption that passes the first check after it is never caught by a later one. The
check catches about half of all corruption events; it is a cheap, weak check that bounds the
work lost when a corruption is caught - it does not certify the result, and LL results still
need the double-check. The residue shift is irrelevant to the symbol; the -2 is applied to the
unshifted value the savefiles hold, which `jacobi_check()` hands to GMP's `mpz_jacobi`.

**When it runs.** On every restart-file read, before anything is built on the loaded residue
(the S-H checksum triplet only proves a file is self-consistent); on the final residue; and at
checkpoints once `JacobiCheckHours` (default 12) have elapsed since the previous check, or at
every checkpoint when that is 0. Never on an interrupt-driven checkpoint, so shutdown stays
fast, and never sooner than 100x the previous check's duration, so the overhead stays near 1%
on any hardware. Hours rather than iterations because the check is scalar GMP work whose cost
relative to an FFT iteration varies by an order of magnitude across machines: 0.08 s at 1M
bits, 28 s at 100M, 146 s at 332M on an i9-10885H with GMP 6.3.

**Rollback.** A passing check keeps the two most recent Jacobi-passed checkpoints as
`p[exp].J` and `p[exp].J1`. A failure normally implies the corruption postdates the last
passing check (an earlier one that passed would go on passing), so `.J` is the natural target.
`.J1` covers an unreadable `.J`, and the one case where `.J` is poisoned: a corruption that
struck exactly at a checked checkpoint, passed that check and fails every later one - the retry
from `.J` then fails at the same iteration and the chain moves on to `.J1`. The tests produce
exactly that case. The run rolls back through p -> q -> .J -> .J1 -> scratch,
checking each file as it is read; the fifth consecutive failure aborts with a hardware
warning. The Gerbicz-restart shift mod-doubling is reused.

**Error accounting.** `error-code` bits 4-7 carry the Jacobi count, following the GIMPS
server's layout (roundoff 8-13, Gerbicz 20-23), and the `errors` object gains `"jacobi"`.
The count is appended to the savefile as a 4-byte field after the two existing counts, so
older readers - which stop after those - are unaffected; verified by resuming a new-format
savefile with a `main` binary and by reading one with AutoPrimeNet's gimps_status.py.

**mlucas.ini.** `JacobiCheck = 0|1` (default 1) and `JacobiCheckHours` (default 12; 0 = every
checkpoint). Also fixed on the way: `mlucas_getOptVal()` matched option names as substrings,
so `JacobiCheck` read the `JacobiCheckHours` line - it now matches whole names.

**Tests.** `tests/jacobi-check.sh` (wired into the Linux CI job): clean runs of M44497 and
M44501 with a check at every checkpoint; interrupt and resume; a damaged p/q pair refused on
read with the run resuming from `.J`; and fault injection via a test-only
`-DMLUCAS_FAULT_INJECT` build that perturbs one residue digit at a chosen checkpoint. Of the
digits tried, some are caught at the corrupted checkpoint (rollback to the previous checkpoint,
final residue equal to the clean run's, error-code nibble 1), some pass it and are caught at
the next check (the chain walks p -> .J -> .J1 and recovers, nibble 3), and some are missed -
and the missed run ends with a wrong residue. That last case is asserted on purpose: it is the
check's known limit, and the test documents it rather than hiding it.

**Found by the tests, fixed here.** The rollback chain's retry count was cleared by every passing check, including the earlier checkpoints passing again on the way back up after a rollback, so a *reproducible* fault at one iteration walked scratch -> pass -> fail -> rollback -> ... forever instead of aborting at the fifth failure. The count is now cleared only by a pass at or beyond the last failure point. Test T6 re-fires the injected fault on every visit of its iteration (`MLUCAS_FAULT_REPEAT`) and asserts the chain p -> .J -> .J1 -> scratch and the abort with the savefiles left in place.

**Docs.** help.txt sections [4], [7], [11], [12] and a new [14b] (log lines and what to do
about each; the JSON error-code layout, which was undocumented); README feature table.

Depends on #170, #108 and #178 (`mlucas_fopen_atomic`/`mlucas_rename`/`savefile_ends_at`,
the interrupt-safe checkpoint, `loop_exit`/`gchk_nfail`): the branch is `main` plus those three
plus four commits of its own, so until they land this diff includes their changes as well - the
feature is the last four commits. I'll rebase onto `main` once they are in. Merge order for the
whole capability is in #224.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,18 @@ jobs:
             # and costs slightly less than 127 did (726k trial divides vs 1.66M).
             ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 521 -bmax 20 )
         done
+    - name: Jacobi residue check tests
+      run: |
+        set -x
+        set -o pipefail
+        # A second native build with the test-only fault injector compiled in (see tests/jacobi-check.sh
+        # for what it is used for). The native obj/ Makefile is reused with one extra define; its objects
+        # were cleaned above, so this is a full compile of the native mode.
+        mkdir obj_fi
+        sed 's/^CPPFLAGS += /CPPFLAGS += -DMLUCAS_FAULT_INJECT /' obj/Makefile > obj_fi/Makefile
+        make -C obj_fi -j "$(nproc --all)" Mlucas
+        JACOBI_TEST_CPU="0:$(( $(nproc --all) - 1 ))" JACOBI_TEST_DIR="$PWD/obj/jacobi-tests" \
+            time bash tests/jacobi-check.sh obj/Mlucas obj_fi/Mlucas obj/mlucas.cfg |& tee -a obj/test.log
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Feature | | Mlucas | Prime95/MPrime
 \- | Suyama | ✔️ | 
 **PRP** | Proof generation | | ✔️
 \- | Proof certification | | ✔️
-**Error Checking** | Jacobi (LL/P-1) | | LL only
+**Error Checking** | Jacobi (LL/P-1) | LL only | LL only
 \- | Gerbicz (PRP/Pépin) | ✔️ | ✔️
 **Random Shifts** | | ✔️ | ✔️
 **Interface** | CLI | ✔️ | MPrime only
@@ -106,7 +106,7 @@ This README is still in progress. For now, see the original [Mlucas README](http
 
 ## Help
 
-The [help.txt](help.txt) file includes a variety of usage information not covered in the original [README](https://mersenneforum.org/mayer/README.html), concentrating largely on the Mlucas command line options. A separate documentation page covers [Fermat numbers](docs/Fermat-testing.md).
+The [help.txt](help.txt) file includes a variety of usage information not covered in the original [README](https://mersenneforum.org/mayer/README.html), concentrating largely on the Mlucas command line options. A separate documentation page covers [Fermat numbers](docs/Fermat-testing.md). Lucas-Lehmer tests are protected by a Jacobi residue check, controlled by the `JacobiCheck` and `JacobiCheckHours` mlucas.ini options - see help.txt section [11].
 
 ## Contributing
 
--- a/help.txt
+++ b/help.txt
@@ -229,6 +229,24 @@ The (very rough) time estimates are for 1000 iterations done using 4 or more cor
 	given Mersenne number exponent on their hardware, similarly to the -fft option. Performs
 	as many iterations as specified via the -iters flag (which is required if -m is invoked).
 
+o JACOBI RESIDUE CHECK: For production (i.e. non-self-test) Lucas-Lehmer tests of Mersenne
+numbers - not PRP, not p-1, and not Fermat/Pe'pin tests - Mlucas computes the Jacobi symbol
+J(s - 2 | M(p)) of the current residue s, which must equal -1 for every LL iterate if the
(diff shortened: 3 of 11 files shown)


MarkRose:
Follow-up: `tests/jacobi-check.sh` now passes the ShellCheck job.

That job runs with `-o avoid-nullary-conditions,deprecate-which,quote-safe-variables,require-double-brackets` and fails on any finding at all; this file contributed 15. SC2015 for the `cond && ok ... || bad ...` idiom, SC2248 for the unquoted `$pid` expansions, SC2034 for the `seq`-driven loop counter. No behaviour change - the `&& ||` cases become the if/then/else they already meant, since `ok` and `bad` both return 0.

MarkRose:
Pushed a follow-up: the suite asserted three things that are not the same on every build, and failed on AVX-512 and ASIMD as a result. None of them was a defect in the checks being tested.

**T1/T2 could not run at all on AVX-512.** M44497 and M44501 default to a 2K FFT, and an AVX-512 build cannot run 2K: the teensy-FFT guard wants `complex-length/radix_final >= 16*RE_IM_STRIDE`, which is 128 there against 64 for AVX/AVX2, and no 2K radix set reaches it. Every 2K radix set is rejected, no cfg entry can exist, the remedial self-test at 2K fails the same way, and the run aborts having done nothing. On a Zen 4, `Mlucas -s tt` passes 4 of 26 cases and writes entries for 4K and 8K only.

Forcing the length does not work around it - for a Mersenne production run the 9/8 rule in `ernstMain()` discards any forced length more than one size above the default, so `-fft 4` on a 2K exponent is silently reverted to 2K. So the exponents move to ones whose default is already 4K: **M86243**, the Mersenne prime in that band, and **86249**, a prime whose Mersenne number is composite so T2 still gets its results line.

**T4 damaged byte 100** to break the savefile checksum. The header is not the same length in every build and a fixed small offset can land inside it, where the damage does not break the residue checksum and the file is accepted. It now damages the middle of the file.

**T5's digit-to-verdict mapping is not a property of the program.** The injector perturbs a digit of the residue as stored, and where that lands depends on the FFT length and radix set the host chose. Measured, same exponent and `-shift 0`:

| digit | AVX2 | AVX-512 | ASIMD |
|---|---|---|---|
| 3 | caught | caught | **not caught** |
| 13 | caught late | caught | caught |
| 21 | - | not caught | caught |

So the fixed mapping could only hold on the machine it was classified on, and T6's hard-coded digit 3 is not caught on ARM at all - which is why its entire rollback chain failed there.

T5 now classifies at run time and asserts what holds for every digit: if any check fires, the chain must recover the clean residue and the error-code nibble must equal the failure count; if none fires, the run must end with a wrong residue - the documented blind spot, still asserted so it stays visible. One global assertion keeps it honest: at least one digit must be caught, so an inert check cannot pass. T6 drives its retry chain with a digit T5 found is caught at its own checkpoint, and skips with a notice if there is none.

Verified on all three, each at 4 threads as CI runs them:

| | |
|---|---|
| AVX2 (Zen 4) | 47 passed, 0 failed |
| AVX-512 (Zen 4) | 46 passed, 0 failed |
| ASIMD (Pi 4) | 46 passed, 0 failed |

PR #265, opened 2026-09-07
is_prime(): sieve out multiples of 3 and 5 before the pseudoprime-table lookup

MarkRose:
`is_prime(n)` was `pprimeF(n,2) && !is_f2psp(n)`, but `f2psp[]` is the table of the 9366 base-2 Fermat pseudoprimes below 2^32 *not divisible by 3 or 5* - it assumes the caller has sieved those out, as `nprimes_in_range()` does with its mod-30 wheel. `is_prime()` and `next_prime()` did not, so every base-2 pseudoprime with a factor of 3 or 5 passed as prime: 561, 645, 1105, 1905, 2465, 4371, 8481, 10585, 11305, 12801, 16705, 18705 below 20000 alone.

Effects: the P-1 stage 1 prime-powers product picked up these composites (harmless over-coverage - I found it because an independent computation of the stage 1 exponent disagreed with Mlucas's by exactly their product), and the worktodo exponent check accepted them as LL/PRP exponents (`Test=18705` is let through where `Test=44499` is rejected).

Fix: reject `n%3 == 0` and `n%5 == 0` in `is_prime()` after the `n < 7` early-out, and make `next_prime()` go through `is_prime()`. A new startup self-test, `test_is_prime()`, asserts the twelve composites, the count pi(20000) = 2262, and `next_prime()` stepping over 561 in both directions. On the unfixed code it reports all twelve, a count of 2274 and `next_prime(560,+1) = 561`; with the fix it is silent and `-s tiny` passes as before.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -1458,7 +1458,6 @@ void host_init(void)
 	uint32 imax = 100000;
 	fprintf(stderr,"INFO: Testing 64-bit 2^p (mod q) functions with %u random (p, q odd) pairs...\n",imax);
 	ASSERT(test_twopmodq64(imax) == 0, "test_twopmodq64() returns nonzero!");
-	ASSERT(test_is_prime() == 0, "test_is_prime() returns nonzero!");
 #ifdef TEST_MI64_PRP
 	const uint32 max_test_dim = 1024;
 	uint32 i,ihi = 1000,j,jhi;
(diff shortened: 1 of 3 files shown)


src/util.c
```diff
@@ -3958,9 +3959,35 @@ DEV uint32 is_prime(uint32 n) {
...
+	for(i = 0; i < 8; i++) {
+		if(!is_prime(small_primes[i])) { fprintf(stderr,"test_is_prime: is_prime(%u) = 0!\n",small_primes[i]); nfail++; }
+	}
+	for(n = 0; n < 20000; n++) { np += is_prime(n); }
+	if(np != 2262) { fprintf(stderr,"test_is_prime: counted %u primes below 20000, expected 2262!\n",np); nfail++; }
+	if(next_prime(560,+1) != 563) { fprintf(stderr,"test_is_prime: next_prime(560,+1) = %u, expected 563!\n",next_prime(560,+1)); nfail++; }
+	if(next_prime(562,-1) != 557) { fprintf(stderr,"test_is_prime: next_prime(562,-1) = %u, expected 557!\n",next_prime(562,-1)); nfail++; }
+	if(next_prime(4,-1) != 3 || next_prime(4,+1) != 5) { fprintf(stderr,"test_is_prime: next_prime around 4 wrong!\n"); nfail++; }
+	return nfail;
+}
```
tdulcet, line 3989: It would be nice if we had C23's `static_assert` here, as otherwise I do not think it is worth the cost to run these checks at every startup for non-critical functionality.

tdulcet (commented):
Thanks for catching this bug. It should help speed up P-1 stage 1, if it now has to process fewer numbers (only the actual primes).

src/util.c
MarkRose: Agreed - moved the call into the `selfTest` block in `main()` so it runs under `-s` (which every CI job does) instead of on every startup. Details and the verification are in the PR comment below.

ldesnogu (approved):
LGTM

PR #266, opened 2026-09-07
Clear all of main's own compiler warnings

MarkRose:
The warning sites current `main` emits on gcc/clang × nosimd/sse2/avx2/avx512. Measured, not
estimated: **four sites**, not the ~21k in the issue — that count was the #238 integration branch
across all CI artifacts, and a raw `grep 'warning:'` is dominated by deliberate `#warning`
directives (106 of 109 lines in a nosimd build).

## `mi64.c` — `BLOCKLENM1`/`BASEADDRMASK`/`minlen`, two sites

Declared under a `USE_AVX2` / `USE_AVX` / `YES_ASM` chain, but only the asm loops consume them, so
a build that takes the portable-C path — `NO_ASM`, which `mi64.h` also selects automatically under
AddressSanitizer — declares them and never uses them. The third arm of each chain was already
`YES_ASM`-guarded; the first two now are too.

The technique that separates this from a false positive: compile the file with and without
`-DNO_ASM` and compare. That distinguishes "dead on every path, delete it" from "used only by the
portable-C arm, move it inside `#ifndef YES_ASM`" — deleting the latter breaks every non-x86 build.

## `mi64.c` — `i` in `mi64_pprimeF()`

Set by the Newton-iteration loop and never read. Not a missing use: the comment immediately above
documents an optimization that is explicitly deferred ("do this optimization later"), and the
converged-64-bit-word count is what that optimization would need. Until it exists the loop
recomputes the full `wlen`-word product every pass and has no use for the count. Removed the dead
counter and moved the note into the loop comment so the intent survives.

## `radix16_dif_dit_pass.c` — `-Wcomment`

A 33-line block comment documenting an asm-translation recipe contains two literal `/*` sequences
as part of the recipe text. Converted to line comments, which preserves the text byte-for-byte —
verified by diffing the old body against the new one with the leading `//` stripped — rather than
editing documentation to appease the compiler. One wrinkle: one line of that recipe ends in a
backslash, which in a `//` comment splices into the next line and trades `-Wcomment` for
`-Wcomment` ("multi-line comment"). The backslash is part of the recipe, so it stays and is simply
no longer line-final.

## `radix32_ditN_cy_dif1.c` — `cy_i` (second commit)

gcc reports this as `unused variable 'cy_i'` in avx/avx2/avx512 and `variable 'cy_i' set but not
used` in sse2 — and that difference is the whole explanation.

The comment on the declaration says `cy_i` is not gated on the SIMD mode because
`radix32_main_carry_loop.h`'s `MODULUS_TYPE_GENFFTMUL` branch reads it in every build. True, but
incomplete: that loop is only *inside* `radix32_ditN_cy_dif1()` in a single-threaded build. A
MULTITHREAD build runs it from `cy32_process_chunk()`, which has its own `cy_i`. The carry-pointer
setup already knew this — its AVX-512 and AVX arms guard the assignment with
`#ifndef MULTITHREAD`, so there the variable was declared and never even written ("unused"). The
SSE2 arm was the one arm missing that guard, so there it was written and never read ("set but not
used"). Gating the declaration and the SSE2 assignment the same way the other two arms already
were makes all three consistent.

Verified by preprocessing the file per mode with the `__LINE__` values baked into the `ABORT`/`WARN`
strings normalized away, and diffing against main. The complete set of changes:

| build | difference from main |
| --- | --- |
| nosimd, nosimd single-threaded | semantically identical |
| avx / avx2 / avx512 threaded | only the `*cy_i` declaration goes away |
| sse2 threaded | declaration plus `cy_i = tmp+0x10`; the `tmp += 0x20` after it stays, so the local-store layout is unchanged |
| sse2 / avx single-threaded | declaration moves out of the comma-list into its own `static vec_dbl *cy_i;`, same storage class and type; assignment retained |

Nothing else differs in any mode — so the ARM ASIMD builds, which take the same carry-pointer arm
as SSE2, are covered by the same argument.

**Residues**, sse2, `-cpu 0:3`, against the same build of main: `-s tt` gives 26 identical `Res64`
values, `-s t` gives 93 identical `Res64` values across 32 FFT lengths. Both rc=0, no
`0 of N radix-sets passed` lines either side.

## `factor.c` — `-Wrestrict` ×2 (third commit)

The two *"Current q = %s"* error messages in `PerPass_tfSieve` render their operands with
`convert_mi64_base10_char(cbuf, ...)` and then hand `snprintf` a pointer **into `cbuf`** — which is
`snprintf`'s own destination. Passing a source string that lives inside the destination violates the
`restrict` qualifiers on `snprintf`'s parameters, and gcc says so: *"'snprintf' argument 7 may
overlap destination object 'cbuf'"*. So a diagnostic whose whole job is to tell the user which `q`
failed can print anything at all.

Both renderings now go into `cbuf2`, side by side, needing no new storage: `cbuf2` is
`STR_MAX_LEN*2` chars, and `convert_mi64_base10_char` passes `n_alloc_chars = STR_MAX_LEN` and
enforces it with `ASSERT(MAX_DIGITS < n_alloc_chars)`, so one rendering touches at most the low
`STR_MAX_LEN` chars of whatever it is handed. The second operand uses the `__`-prefixed entry point
to render into `cbuf2+STR_MAX_LEN` under that same bound.

## `Mlucas.c` — `-Wformat-truncation` (third commit)

The known-factor-does-not-divide message writes into `g_cstr` (`STR_MAX_LEN`) with a `%s` for `cbuf`
(`STR_MAX_LEN*3`), which at that point holds the factor token exactly as it appeared in the worktodo
line — `strncpy`'d from it, so bounded only by `cbuf`. gcc: *"'%s' directive output may be truncated
writing up to 3071 bytes into a region of size between 992 and 993"*. Truncating there is worse than
it sounds: the message asks the user to *correct or remove that entry*, and a cut-off factor is not
identifiable. Bounded with `%.*s` at `sizeof(g_cstr)/4` — a 256-bit factor needs at most 78 digits,
so 256 chars is ample and makes the bound provable.

## Result

Measured file-by-file, gcc and clang × nosimd/sse2/avx2/avx512 at `-O3`, unique file+message pairs:

| tree | diagnostics |
| --- | --- |
| main | 21 |
| main + this PR + #272 | 11 — **all 11 in `src/test_fft_radix.c`** |

So main's *built* sources go fully warning-clean. `mi64.c` also compiles clean with **and** without
`-DNO_ASM`.

Residues, nosimd, `-cpu 0:3`, against a pristine build of main: `-s tt` 41 identical `Res64` values,
`-prp -s tt` 43 identical, both rc=0. `Mfactor -m 521 -bmax 20` completes, rc=0.

## Deliberately not touching `src/test_fft_radix.c`

It is in no build target — not in `OBJS`, not in `OBJS_MFAC` — and `src/test_fft_radix.c.txt`
documents compiling it by hand with `-DRADIX=[radix] -DTTYPE=[test type]`. Its warning set depends
on both:

| flags | diagnostics |
| --- | --- |
| default (`TTYPE=3`) | 10 |
| `-DTTYPE=0` | 19 |
| `-DTTYPE=1` | 3 |
| `-DTTYPE=2` | 3 |
| `-DRADIX=16 -DTTYPE=1` | 5 |

There is no fixed set of unused variables to delete: several are used under some `TTYPE` and not
others, so deleting what the default does not use would break the other three test modes. Same shape
as the `mi64` `NO_ASM` case above, and it wants a per-mode sweep rather than a one-config edit.

Still outstanding on #244, not in this PR: `factor.c`'s `p192`/`q192`/`t192`/`p256`/`q256`/`t256`
(#269 — a symptom of a dispatch bug, not just dead declarations), `brev64`/`reverse64`'s
strict-aliasing punning (#270), the `leadz`/`trailz` asm output operands (#271), the `[2a]` retry
check's uninitialized `iter` (#272), and the macOS/ARM/FreeBSD/old-distro-container configurations,
which I have not reproduced.

Part of #244.

### DIFF
--- a/src/factor.c
+++ b/src/factor.c
@@ -3394,24 +3394,27 @@ MFACTOR_HELP:
 							/* Make sure that q == 1 (mod 2p) and that q is a base-2 PRP: */
 							mi64_clear(u64_arr, lenQ);	// Use q2 for quotient [i.e. factor-candidate k] and u64_arr for remainder
 							mi64_div(q,two_p,lenQ,lenQ,q2,u64_arr);
-							/* The %s operands below are rendered into cbuf2, not cbuf: cbuf is snprintf's own
+							/* The %s operands below render into cbuf2, never into cbuf: cbuf is snprintf's own
 							destination, and handing snprintf a source string that lives inside its destination
-							buffer is a restrict violation - gcc says so, "'snprintf' argument 7 may overlap
-							destination object 'cbuf'". Both renderings fit in cbuf2 side by side: cbuf2 is
-							STR_MAX_LEN*2 chars and convert_mi64_base10_char passes n_alloc_chars = STR_MAX_LEN,
-							which it enforces with ASSERT(MAX_DIGITS < n_alloc_chars), so a rendering touches at
-							most the low STR_MAX_LEN chars of what it is handed. */
+							is a restrict violation - gcc says so, "'snprintf' argument 7 may overlap
+							destination object 'cbuf'".
+							That puts both renderings of the second message in cbuf2, so they need somewhere
+							separate to sit or the later one would overwrite the earlier before snprintf reads
+							it. cbuf2 is STR_MAX_LEN*2, and convert_mi64_base10_char() calls through with
+							n_alloc_chars = STR_MAX_LEN, enforced by ASSERT(MAX_DIGITS < n_alloc_chars) - so a
+							rendering touches at most STR_MAX_LEN chars and cbuf2 holds exactly two of them: */
+							char *const qstr = cbuf2, *const rstr = cbuf2 + STR_MAX_LEN;
 							if(mi64_getlen(q2, lenQ) != 1) {
 								snprintf(cbuf, sizeof(cbuf), "ERROR: Count = %u * 2^%u: k = %" PRIu64 ", Current q = %s: k must be 64-bit!\n",
-									(uint32)(count >> CMASKBITS),CMASKBITS,k,&cbuf2[convert_mi64_base10_char(cbuf2, q, lenQ, 0)]);
+									(uint32)(count >> CMASKBITS),CMASKBITS,k,&qstr[convert_mi64_base10_char(qstr, q, lenQ, 0)]);
 								fprintf(fp,"%s", cbuf);
 								ASSERT(0, cbuf);
 							}
 							if(!mi64_cmp_eq_scalar(u64_arr, 1ull, lenQ))
 							{
 								snprintf(cbuf, sizeof(cbuf), "ERROR: Count = %u * 2^%u: k = %" PRIu64 ", Current q = %s: q mod (2p) = %s != 1!\n",
-									(uint32)(count >> CMASKBITS),CMASKBITS,k,&cbuf2[convert_mi64_base10_char(cbuf2, q, lenQ, 0)],
-									&cbuf2[STR_MAX_LEN + __convert_mi64_base10_char(cbuf2+STR_MAX_LEN, STR_MAX_LEN, u64_arr, lenQ, 0)]);
+									(uint32)(count >> CMASKBITS),CMASKBITS,k,&qstr[convert_mi64_base10_char(qstr, q, lenQ, 0)],
+									&rstr[__convert_mi64_base10_char(rstr, STR_MAX_LEN, u64_arr, lenQ, 0)]);
 								fprintf(fp,"%s", cbuf);
 								ASSERT(0, cbuf);
 							}
--- a/src/factor.c
+++ b/src/factor.c
@@ -3394,17 +3394,24 @@ MFACTOR_HELP:
 							/* Make sure that q == 1 (mod 2p) and that q is a base-2 PRP: */
 							mi64_clear(u64_arr, lenQ);	// Use q2 for quotient [i.e. factor-candidate k] and u64_arr for remainder
 							mi64_div(q,two_p,lenQ,lenQ,q2,u64_arr);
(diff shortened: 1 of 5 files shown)


src/factor.c
```diff
@@ -3394,17 +3394,24 @@ candidate factors that survive sieving.	*/
...
 								fprintf(fp,"%s", cbuf);
 								ASSERT(0, cbuf);
 							}
 							if(!mi64_cmp_eq_scalar(u64_arr, 1ull, lenQ))
 							{
 								snprintf(cbuf, sizeof(cbuf), "ERROR: Count = %u * 2^%u: k = %" PRIu64 ", Current q = %s: q mod (2p) = %s != 1!\n",
-									(uint32)(count >> CMASKBITS),CMASKBITS,k,&cbuf[convert_mi64_base10_char(cbuf, q, lenQ, 0)],
-									&cbuf2[convert_mi64_base10_char(cbuf2, u64_arr, lenQ, 0)]);
+									(uint32)(count >> CMASKBITS),CMASKBITS,k,&cbuf2[convert_mi64_base10_char(cbuf2, q, lenQ, 0)],
+									&cbuf2[STR_MAX_LEN + __convert_mi64_base10_char(cbuf2+STR_MAX_LEN, STR_MAX_LEN, u64_arr, lenQ, 0)]);
```
tdulcet: If we are using a separate buffer now (`cbuf2`), is the `+STR_MAX_LEN` still needed?

tdulcet (commented):
Thanks for fixing this warnings.

tdulcet (approved):
Thanks, this is much more readable now.

PR #268, opened 2026-09-07
P-1: Gerbicz check for stage 1, Jacobi integrity checks, exact input checks for stage 2 (resolves #97)

### LINKED ISSUE #97: Add support for the Jacobi error check for LL and P-1 tests

MarkRose:
Resolves #97 (the P-1 half; the LL half is #264, which this branch is based on - only the commits after
#264's tip are new here).

**Stage 1: a Gerbicz check, not a Jacobi check.** Stage 1 is x <- x^2 * 3^b per iteration, so the PRP
Gerbicz identity applies with one correction factor: with b[] the running product of the block-boundary
residues and d[] its copy from the previous check, b == u0 * d^(2^L) * 3^C, where C is the sum of the L-bit
exponent chunks consumed since the check epoch started and u0 the residue the epoch started from (3 from
scratch). C is recomputed from the prime-powers product at each check and split C = H * 2^L + Lo: the L
verification squarings run against a private multiply-by-base bit array holding Lo, so they produce
d^(2^L) * 3^Lo for free, and the remainder u0 * 3^(H * 2^L) is g3^H against a once-per-run g3 = 3^(2^L),
~2 log2 H modmuls. The check-product update runs against that private array too, because its carry step
would otherwise pick up the live exponent bit for its iteration (it did, and multiplied a stray 3 into the
product - found by comparing b[] against an independent Python product of the same residues). Detection of
an arithmetic error in the checked interval is essentially certain, at ~2/L of stage 1; a failure rolls back
to the `.G` savefile exactly as PRP does. Why not Jacobi here: squaring maps every value into the quadratic
residues, so a corrupted residue carries the expected symbol from the next iteration on - a Jacobi check
cannot see stage 1 arithmetic at all.

**Check interval.** `GerbiczCheckInterval` in mlucas.ini (p-1 only; PRP keeps 10^6) sets the interval, which
must be L^2 with L dividing CheckInterval and CheckInterval dividing L^2 - an unusable value is reported with
the admissible ones and the automatic choice used. Automatic: the smallest admissible L >= 100 with L^2
iterations >= ~2 h at the mlucas.cfg per-iteration time, capped at L = 1000, so it only shortens the interval
on slow or very large runs (10^6 iterations is 14-55 h of exposure at 100M-digit exponents). Since the
>4-thread CheckInterval default of 10^5 admits only L = 1000, p-1 defaults CheckInterval to 10^4 unless set.

**Savefiles.** The check-product is appended to the p/q (and `.G`) stage 1 savefiles after the error
counts, behind an 8-byte epoch field, so older readers - which stop after the counts - are unaffected. A
file without it starts a new check epoch from the residue it holds. `DO_GCHECK` is cleared before stage 2,
whose own savefile I/O is untouched.

**Gerbicz retry bound (PRP too).** Found by the tests: a Gerbicz failure that reproduces (a fault at one iteration) rolled back to `.G` and failed there again forever, because the retry count was bounded only for the end-of-run check and was cleared by every passing check on the way back up. Every check's failures are now bounded (the fourth in a row aborts with the hardware warning, as the end-of-run one did) and the count is cleared only by a pass at or beyond the failure point. Builds on #178's `gchk_nfail`; applies to PRP as much as to P-1.

**Jacobi in P-1, reduced to what it can actually see:** the savefile and the FP-to-integer conversion path,
which the Gerbicz check cannot (both its operands go through it). It runs on every stage 1 restart-file
read (expected symbol J(3|N) if the last consumed exponent bit is 1, else +1) with fall-through to the
secondary on a mismatch, and on the final stage 1 residue before the GCD (expected +1) - that one runs on
its own thread alongside the GCD, so it costs no wall time. There is no periodic P-1 Jacobi check - there
is nothing for one to find. LowMem = 2, which has no room for the
Gerbicz arrays, runs stage 1 with no arithmetic check and says so once.

**Stage 2: exact checks of the inputs, honesty about the accumulator.** The accumulator is a product of
differences of powers and has no algebraic invariant that could be checked cheaply; the two inputs every
later term is built from do: the A^((kD)^2) ladder value is recomputed from the stage 1 residue by the same
two-step modpow the setup uses and must be bit-identical in the FFT domain to the running value (~130
modmuls, on the `JacobiCheckHours` clock), and the static A^(b^2) tables are checksummed after they are
built and re-verified a rotating 1/16 per checkpoint. The accumulator gets sanity tests only (never zero,
never unchanged across a checkpoint). Every failure stops the run with the `.s2` checkpoint intact; a
restart rebuilds ladder and tables and resumes the accumulator. A corruption of the accumulator itself is
not caught - the consequence is a missed factor, never a false claim - and the tests assert that it is
not caught, so the limit is documented rather than hidden.

**Results line.** P-1 output gains `error-code` (same layout as LL/PRP) and
`errors:{Roundoff, gerbicz, jacobi}`.

**Tests.** `tests/jacobi-pm1.sh`, in the Linux CI job: a short stage 1 with GerbiczCheckInterval = 10^4
(three checks, the folded-in correction with a large high part, the overlapped final Jacobi check) and an
unusable interval value rejected; clean stage 1 at B1 = 1.5M (checks at 10^6,
2*10^6 and the end-of-run boundary); interrupt/resume with the stored product carrying the epoch across
the restart; a damaged primary refused with fall-through to the secondary; a fault injected at 1.5M caught
at the 2M check with rollback to `.G` and the clean final residue; and stage 2 at B2 = 2M with clean checks
at every checkpoint, an injected ladder fault and an injected table fault each caught and recovered after
restart, and an injected accumulator fault demonstrably not caught. Also: a reproducible Gerbicz failure bounded at four attempts (three rollbacks to `.G`, then the abort) instead of looping; LowMem = 2 running stage 1 with no arithmetic check and saying so once; the same stage 1 checks on a Fermat number (F16 at FFT length 4K, the known factor 825753601 found - it needs the fermat.cfg entry config-fermat.sh writes and is skipped without one); and, in `tests/jacobi-check.sh`, the between-intervals interrupt of #267 exercised directly on a PRP run (T7) and the LL rollback chain walked to its end by a reproducible fault (T6).

**Docs.** help.txt sections [9], [11], [12], [14b]; docs/gerbicz.txt and docs/pm1.txt appendices; README
table.

Also included as its first commit: the fix for an interrupt arriving between iteration intervals leaving a
run "completing" with a frozen residue (#267, opened separately so PRP/Pépin users need not wait for this).
Merge order for the whole capability is in #224.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
             # and costs slightly less than 127 did (726k trial divides vs 1.66M).
             ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 521 -bmax 20 )
         done
-    - name: Jacobi residue check tests
+    - name: Jacobi / Gerbicz error-check tests
       run: |
         set -x
         set -o pipefail
@@ -109,6 +109,8 @@ jobs:
         make -C obj_fi -j "$(nproc --all)" Mlucas
         JACOBI_TEST_CPU="0:$(( $(nproc --all) - 1 ))" JACOBI_TEST_DIR="$PWD/obj/jacobi-tests" \
             time bash tests/jacobi-check.sh obj/Mlucas obj_fi/Mlucas obj/mlucas.cfg |& tee -a obj/test.log
+        JACOBI_TEST_CPU="0:$(( $(nproc --all) - 1 ))" JACOBI_TEST_DIR="$PWD/obj/jacobi-pm1-tests" \
+            time bash tests/jacobi-pm1.sh obj/Mlucas obj_fi/Mlucas obj/mlucas.cfg |& tee -a obj/test.log
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,18 @@ jobs:
             # and costs slightly less than 127 did (726k trial divides vs 1.66M).
             ( cd "../obj_mfac${word:+_$word}${MSFX}" && time "./Mfactor${word:+_$word}" -m 521 -bmax 20 )
         done
+    - name: Jacobi residue check tests
+      run: |
+        set -x
+        set -o pipefail
+        # A second native build with the test-only fault injector compiled in (see tests/jacobi-check.sh
+        # for what it is used for). The native obj/ Makefile is reused with one extra define; its objects
+        # were cleaned above, so this is a full compile of the native mode.
+        mkdir obj_fi
+        sed 's/^CPPFLAGS += /CPPFLAGS += -DMLUCAS_FAULT_INJECT /' obj/Makefile > obj_fi/Makefile
+        make -C obj_fi -j "$(nproc --all)" Mlucas
+        JACOBI_TEST_CPU="0:$(( $(nproc --all) - 1 ))" JACOBI_TEST_DIR="$PWD/obj/jacobi-tests" \
+            time bash tests/jacobi-check.sh obj/Mlucas obj_fi/Mlucas obj/mlucas.cfg |& tee -a obj/test.log
     - uses: actions/upload-artifact@v7
       if: always()
       with:
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Feature | | Mlucas | Prime95/MPrime
 \- | Suyama | ✔️ | 
 **PRP** | Proof generation | | ✔️
 \- | Proof certification | | ✔️
-**Error Checking** | Jacobi (LL/P-1) | | LL only
+**Error Checking** | Jacobi (LL/P-1) | LL only | LL only
 \- | Gerbicz (PRP/Pépin) | ✔️ | ✔️
 **Random Shifts** | | ✔️ | ✔️
 **Interface** | CLI | ✔️ | MPrime only
@@ -106,7 +106,7 @@ This README is still in progress. For now, see the original [Mlucas README](http
 
 ## Help
 
-The [help.txt](help.txt) file includes a variety of usage information not covered in the original [README](https://mersenneforum.org/mayer/README.html), concentrating largely on the Mlucas command line options. A separate documentation page covers [Fermat numbers](docs/Fermat-testing.md).
+The [help.txt](help.txt) file includes a variety of usage information not covered in the original [README](https://mersenneforum.org/mayer/README.html), concentrating largely on the Mlucas command line options. A separate documentation page covers [Fermat numbers](docs/Fermat-testing.md). Lucas-Lehmer tests are protected by a Jacobi residue check, controlled by the `JacobiCheck` and `JacobiCheckHours` mlucas.ini options - see help.txt section [11].
 
 ## Contributing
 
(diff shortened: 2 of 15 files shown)


MarkRose:
Follow-up: `tests/jacobi-pm1.sh` and this branch's copy of `tests/jacobi-check.sh` now pass the ShellCheck job.

That job runs with `-o avoid-nullary-conditions,deprecate-which,quote-safe-variables,require-double-brackets` and fails on any finding at all; these two files contributed 33 between them. SC2015 for the `cond && ok ... || bad ...` idiom, SC2248 for the unquoted `$pid` expansions, SC2034 for the `seq`-driven loop counters. No behaviour change - the `&& ||` cases become the if/then/else they already meant, since `ok` and `bad` both return 0.

MarkRose:
I pushed 191af08 to fix a bug in the stage 1 Gerbicz check at FFT lengths above 32K. `pm1_gcheck_g3()` and `pm1_gcheck_apply()` cleared g3/g2 and copied u0 into c over `n` doubles, but the residue arrays use the padded layout (`npad = n + ((n >> DAT_BITS) << PAD_BITS)` doubles, padding on above 32K). So the top padded words, which hold the highest logical words, kept stale data, and the correction modmul blew up. Both helpers now take `npad` from the caller and use a `size_t` byte count (the `int` count would also have overflowed at FFT >= 256M).

`Pminus1=1,2,2000003,-1,60000,0` at 112K with `CheckInterval = 10000` and `GerbiczCheckInterval = 10000`:
- before: the first check aborts with `Unhandled Error of type[9] = ERR_ROUNDOFF in p-1 Gerbicz-check correction step`, on AVX2 and on AVX-512 (under Intel SDE, where PAD_BITS is 3).
- after: on AVX2 all 9 checks pass and stage 1 finds the factor 160000241. On AVX-512 under SDE the first check now passes.
- M600011 at 30K (no padding) passes all 9 checks before and after.

`tests/jacobi-pm1.sh` still passes (34 passed, the fault-injection cases skipped). It runs at 4K, where there is no padding, which is why it never saw this.

PR #273, opened 2026-09-07
Declare the inline-asm operands the templates write as outputs, not "m" inputs

MarkRose:
105 operand declarations across 9 files say `"m"` — a read-only memory input, which is a promise the
asm does not modify it — for operands the template stores into. The compiler is entitled to treat
such an operand as unchanged: to consider the variable still uninitialized afterwards, to discard
the store, or to keep a cached copy and reuse it.

**Nothing is broken today**, and the reason is uniform: every one of these blocks also carries a
`"memory"` clobber, which forces the compiler to assume the asm wrote memory it was not told about,
so the operand's stack slot gets reloaded rather than assumed dead. That is measurable, not
inferred — #271 shows the same three `util.c` sites going silent-to-diagnosed the moment **only**
the `"memory"` clobber is removed, gcc reporting all three as `may be used uninitialized` at
`-O1`/`-O2`/`-O3`.

Which is the point. Narrowing those clobber lists is the work in #68, #71 and the asm-clobber lint —
an over-broad `"memory"` on every block is precisely what they are removing. Here it is load-bearing,
for reasons nothing in the source states, in the mi64 modpow chain and the SSE2/AVX carry macros,
where a lost store is a wrong residue rather than a crash.

Each operand is declared `"=m"` if the template only stores to it, `"+m"` if it reads it first. That
distinction is the whole risk in this change, and the criterion is the **first access in template
order**: if a pure write comes first the incoming value is dead and `"=m"` is right, otherwise `"+m"`
is required. (A destination like `addq %%rax,%[__result]` is a read too — but not if a plain `movq`
into it came earlier in the same template, which is the case in `twopmodq80.c`.) 76 of the 105 are
`"+m"`, mostly pointers the asm advances and hands back: `carry_gcc64.h`'s
`__wtA`/`__wtB`/`__wtC`/`__data`, `mi64.c`'s `__cy*` carry chains.

| file | sites | `=m` | `+m` |
| --- | --- | --- | --- |
| `src/mi64.c` | 42 | 29 | 13 |
| `src/carry_gcc64.h` | 24 | 1 | 23 |
| `src/carry_gcc32.h` | 14 | 0 | 14 |
| `src/twopmodq.c` | 4 | 0 | 4 |
| `src/twopmodq64_test.c` | 12 | 0 | 12 |
| `src/mi64.h` | 3 | 3 | 0 |
| `src/twopmodq80.c` | 3 | 3 | 0 |
| `src/imul_macro0.h` | 2 | 0 | 2 |
| `src/twopmodq100.c` | 1 | 1 | 0 |

That is 105 of the **108** such declarations in the tree; the other three are `util.c`'s, which
are #271. So after these two PRs there are none left.

## The second commit: `twopmodq64_q8` needed room made for it first

Eight of the 105 are in a block that a declaration change alone cannot fix, which is why it gets its
own commit. It has 27 operands, and `x0..x7` genuinely need `"+m"` — the template's first access to
each is a load of the incoming value at the top, and it stores them back at the bottom — so at two
slots apiece that is 35, and `more than 30 operands in 'asm'`.

The operand budget is the thing to fix, and two of the 27 are cheap to reclaim:

- **`__j` is declared but never referenced** anywhere in the template. Its own comment says *"Only
  need this if debug and explicit loop enabled"*. Dropped.
- **`qinv0..7` are eight read-only operands used exactly once each**, in eight consecutive
  `imulq %[__qinvN],%%rN` lines. They now travel as one array-pointer operand, loaded into `rdi`
  once before the loop and addressed as `0x00(%%rdi)`..`0x38(%%rdi)`. `rdi`, `rsi` and `rbp` are
  untouched by the rest of the block, so no register shuffling is needed, and the eight loads stay
  memory operands to the same `imulq`.

27 − 8 + 1 − 1 = 19 operands, so 19 + 8 = **27 slots**. The accounting is the whole argument, so I
measured it rather than trusting the manual: gcc accepts 30 plain `"m"` inputs and rejects 31,
accepts 15 `"+m"` and rejects 16, and accepts 8 `"+m"` alongside 14 `"m"` (30 slots) while rejecting
15 (31).

This is the live `twopmodq64_q8` — `twopmodq.c` defines one too, but `nm` shows the global
definition coming from `twopmodq64_test.o`, and `factor.c` dispatches it in the `TRYQ == 8` arm.

**Verifying it needed its own work**, because the Mfactor self-tests do not reach this routine as
normally built: `factor_test.h`'s q8 check sits under `#elif(TRYQ == 8)` and `makemake.sh` always
passes `TRYQ=4`.

| check | result |
| --- | --- |
| Mfactor built with `-DTRYQ=8`, which compiles `factor_test.h`'s q8 self-test *and* `factor.c`'s `TRYQ == 8` dispatch arm | self-tests pass in both trees, no `twopmodq64_q8 ... failed to find factor`, and `-m 2147483647 -bmax 50` gives the same 657696 trial divides |
| differential harness vs GMP, 12,500 octets over five exponents from M(127) to M(2147483647), with a genuine factor planted in a rotating lane so that **7501 of the 100000 candidates are real factors** | zero GMP mismatches in both trees; the checksum of every returned bitmask identical, `9908A028DE8B25E6` |
| that harness against an injected fault — one `imulq` pointed at the wrong array slot (`0x18`→`0x20`, lane 3 reading lane 4's qinv) | 939 mismatches and a different checksum, so the check is sensitive to exactly what this restructuring could get wrong |
| timing, 12,500 q8 calls, best of three | 194 ms before, 194 ms after |

Random `k` alone, without the planted factors, finds nothing at all across 160,000 candidates — that
harness would have passed against a routine that always answered "no factor", which is why the
planting matters.

## Verification

First, that the checks can see this class of fault at all. Injecting one — making `carry_gcc64.h`'s
`wtA` store land the wrong register, the same shape as a mis-declared operand — collapses the sse2
`-s tt` self-test from 26 residues to 4, with 34 error lines. (It still exits 0; the residue
comparison is what catches it, not the exit status.) Then, with the real change:

| check | result |
| --- | --- |
| generated `.text`, all 90 sources × gcc/clang × nosimd/sse2/avx/avx2/avx512, byte-compared vs main | 966 identical, 40 differ — the edited files plus `radix32_ditN_cy_dif1.c` under clang/sse2 and `radix36_ditN_cy_dif1.c` under clang/avx and clang/avx2. **gcc's output for every radix file is unchanged.** |
| diagnostics over that same matrix | not one warning or error added or removed |
| Mfactor self-test battery (`twopmodq64_q4`/`q8`, `twopmodq80/96/128/160/192/256`, `mi64_twopmodq`) | passes in nosimd, sse2, avx2; `-m 521 -bmax 20` gives the same 1320739 trial divides in all six builds |
| `Mlucas -s tt` / `-prp -s tt`, `-cpu 0:3`, per mode vs the same build of main | nosimd 41 and 43 identical `Res64`, sse2 26 and 26, avx2 23 and 23, all rc=0 |
| AVX-512 under Intel SDE (`-skx`), `-s tt` | 8 identical `Res64`, rc=0, binary confirmed as the AVX-512 build from its own banner |

**Sixteen of the 105 are the exception to all of that** — `carry_gcc32.h`'s 14 and
`imul_macro0.h`'s 2 — because they are in 32-bit x86 code that no supported configuration compiles.
A 32-bit SIMD build stops at `platform.h:1646`, *"SIMD-assembler code only supported for 64-bit
build!"*, and `radix16_ditN_cy_dif1_asm.h:2883`, *"32-bit OSes no longer supported for SIMD
builds!"*; `imul_macro0.h`'s block is in the `CPU_IS_X86` arm, and preprocessing `imul_macro.c` in
nosimd, sse2 and avx2 yields no occurrence of its operands at all, with the resulting object code
identical to main's. Those 16 are transformed by the same rule and pass the same structural checks,
but they are not compile-tested, because they cannot be. Happy to drop both files from the PR if
you'd rather leave dead code untouched.

## How the edit was made, and how to re-check it

I scripted it rather than hand-typing 97 of them. The finder is below - self-contained, needs only
python3, run it from the repo root. On this branch it reports just `util.c`'s three as remaining; on `main` it reports 108.

A companion script performed the move, and its method rather than its code is the reviewable part:
for each affected block it lifts the operand's declaration line out of the input list into the
output list, rewriting `"m"` to `"=m"` or `"+m"`, promotes the next surviving input to carry the
section's `:` when the moved line was the first one, then asserts the block still has three `:`
sections with no plain `"m"` left in the outputs and no `"=m"`/`"+m"` left in the inputs. It refuses
outright any block the conversion would push past 30 operands, which is how the `twopmodq64_q8`
block came to be flagged rather than silently broken — and is what sent me to make room for it in
the second commit.

**Delimiting these asm statements is the part worth stating**, because two plausible rules both
fail. Keying on `);` alone on a line misses single-line `asm(...)` statements and then runs on,
swallowing every block up to the next multi-line one - that is how I first missed `imul_macro0.h`
entirely. Counting parentheses trips over both the `(%%rbx)` in the templates *and* the unbalanced
parens in their trailing `/* ... */` comments - that is how I first undercounted `carry_gcc64.h` at
3 instead of 24. Stripping comments and string literals first, then counting parentheses,
terminates both shapes; and the site list it produces is exactly the **union** of what the two
flawed rules find, with each of them differing from it only by omission, never by inventing a site.
That is the cross-check that convinced me 108 is the whole set.

One caveat in the finder, and the reason the ARM asm is untouched here: on ARM, `%[__add0]` in
`ldr x5,%[__add0]` is the *source*, not the destination. A naive "last operand is the destination"
rule reports 72 false positives across the ARM headers, so the finder filters the ARM load
mnemonics.

If it would be useful I am happy to add this as a lint in a follow-up - standalone, or folded into
the asm-clobber lint job - so the pattern cannot come back. I have kept it out of this PR to hold
the diff to the declarations themselves.

<details>
<summary>The finder, if you want to reproduce the site list (click to expand)</summary>

```python
#!/usr/bin/env python3
"""Audit inline-asm operands declared as read-only "m" INPUTS that the template stores into.

Block delimiting is the whole difficulty. Two rules that look reasonable both fail here:
  - ");" alone on a line: single-line asm statements never match it, so the scan runs on
    and swallows every following block until the next multi-line one ends.
  - counting parentheses: the templates are full of "(%%rbx)" AND the trailing /* ... */
    comments contain unbalanced parens.
So: strip block comments (they span lines), line comments, and string/char literals, and
only then count parentheses. That terminates both shapes correctly.

Per operand, report whether the template also READS it - that decides "+m" over "=m".
"""
import re, sys, glob, json, collections

ARM_LOAD = re.compile(r'\b(ldr|ldp|ld1r?|ldur|ldrb|ldrh|ldrsw)\b')
SEC      = re.compile(r'^([ \t]*):')
ASM      = re.compile(r'(__asm__|(?<![\w.])asm)\s*(__volatile__|volatile)?\s*\(')

def strip_code(lines):
    """Return a parallel list with comments and literals blanked, preserving length."""
    out, in_cmt = [], False
    for ln in lines:
        s, i, res = ln, 0, []
        while i < len(s):
            if in_cmt:
                e = s.find('*/', i)
                if e < 0: res.append(' ' * (len(s) - i)); i = len(s)
                else: res.append(' ' * (e + 2 - i)); i = e + 2; in_cmt = False
            elif s.startswith('/*', i):
                in_cmt = True; res.append('  '); i += 2
            elif s.startswith('//', i):
                res.append(' ' * (len(s) - i)); i = len(s)
            elif s[i] in '"\'':
                q = s[i]; j = i + 1
                while j < len(s) and s[j] != q:
                    j += 2 if s[j] == '\\' else 1
                res.append(' ' * (min(j, len(s) - 1) + 1 - i)); i = j + 1
            else:
                res.append(s[i]); i += 1
        out.append(''.join(res))
    return out

results = []
for path in sorted(glob.glob('src/*.c') + glob.glob('src/*.h')):
    lines = open(path, encoding='utf-8', errors='replace').read().split('\n')
    bare  = strip_code(lines)
    i = 0
    while i < len(lines):
        m = ASM.search(bare[i])
        if not m: i += 1; continue
        depth, j, started = 0, i, False
        while j < len(lines):
            seg = bare[j][m.end()-1:] if j == i else bare[j]
            for ch in seg:
                if ch == '(': depth += 1; started = True
                elif ch == ')': depth -= 1
            if started and depth <= 0: break
            j += 1
            if j - i > 8000: break
        buf = lines[i:j+1]
        secs, cur, tmpl = [], None, []
        for off, ln in enumerate(buf):
            if SEC.match(ln):
                cur = []; secs.append(cur); cur.append((i+off+1, ln))
            elif cur is not None: cur.append((i+off+1, ln))
            else: tmpl.append((i+off+1, ln))
        if len(secs) >= 2:
            positional = bool(re.search(r'%[0-9]', '\n'.join(t for _, t in tmpl)))
            for lineno, text in secs[1]:
                for name in re.findall(r'\[\s*(_{1,2}\w+)\s*\]\s*"m"', text):
                    stores, reads = [], []
                    for ln_no, ln in tmpl:
                        if not re.search(r'%\[\s*'+re.escape(name)+r'\s*\]', ln): continue
                        if ARM_LOAD.search(ln): reads.append(ln_no); continue
                        nd = len(re.findall(r',[ \t]*%\[\s*'+re.escape(name)+r'\s*\][ \t]*(?=\\n|\\t|")', ln))
                        na = len(re.findall(r'%\[\s*'+re.escape(name)+r'\s*\]', ln))
                        if nd: stores.append(ln_no)
                        if na > nd: reads.append(ln_no)
                    if stores:
                        results.append(dict(file=path, decl_line=lineno, operand=name,
                                            constraint='+m' if reads else '=m',
                                            stores=stores, reads=reads,
                                            block=[i+1, j+1], positional=positional))
        i = j + 1

if len(sys.argv) > 1 and sys.argv[1] != '-': json.dump(results, open(sys.argv[1], 'w'), indent=1)
byf = collections.defaultdict(list)
for r in results: byf[r['file']].append(r)
eq = sum(1 for r in results if r['constraint'] == '=m')
print(f"{len(results)} operand declarations to move  (=m: {eq}, +m: {len(results)-eq})", file=sys.stderr)
for f, v in sorted(byf.items(), key=lambda kv: -len(kv[1])):
    e = sum(1 for r in v if r['constraint'] == '=m')
    print(f"  {f:26s} {len(v):3d} in {len({tuple(r['block']) for r in v}):3d} blocks   =m:{e:3d} +m:{len(v)-e:3d}"
          f"   positional:{sum(1 for r in v if r['positional'])}", file=sys.stderr)
```

</details>

Same class as #271.

### DIFF
--- a/src/carry_gcc32.h
+++ b/src/carry_gcc32.h
@@ -153,12 +153,11 @@
 		"addl	%[__idx_incr],%%ebx		\n\t"\
 		"movl	%%ebx, %[__idx_offset]	\n\t"/* Store incremented idx_offset */\
 	"popl %%ebx	\n\t"\
-	:						/* outputs: none */\
+	: [__idx_offset]	"+m" (Xidx_offset)	/* outputs: the template stores into these */\
 	:	[__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
 	,	[__cy]			"m" (Xcy)\
 	,	[__nrt_bits]	"m" (Xnrt_bits)\
 	,	[__nrtm1]		"m" (Xnrtm1)\
-	,	[__idx_offset]	"m" (Xidx_offset)\
 	,	[__idx_incr]	"m" (Xidx_incr)\
 	,	[__half_arr]	"m" (Xhalf_arr)\
 	,	[__sign_mask]	"m" (Xsign_mask)\
@@ -288,12 +287,11 @@
 		"movl	%[__idx_offset],%%esi	\n\t"\
 		"addl	%[__idx_incr],%%esi		\n\t"\
 		"movl	%%esi, %[__idx_offset]	\n\t"/* Store incremented idx_offset */\
-	:						/* outputs: none */\
+	: [__idx_offset]	"+m" (Xidx_offset)	/* outputs: the template stores into these */\
 	:	[__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
 	,	[__cy]			"m" (Xcy)\
 	,	[__nrt_bits]	"m" (Xnrt_bits)\
 	,	[__nrtm1]		"m" (Xnrtm1)\
-	,	[__idx_offset]	"m" (Xidx_offset)\
 	,	[__idx_incr]	"m" (Xidx_incr)\
 	,	[__half_arr]	"m" (Xhalf_arr)\
 	,	[__sign_mask]	"m" (Xsign_mask)\
@@ -475,12 +473,11 @@
 		"addl	%[__idx_incr],%%ecx		\n\t"\
 		"movl	%%ecx, %[__idx_offset]	\n\t"/* Store incremented idx_offset */\
 	"popl %%ebx	\n\t"\
-	:						/* outputs: none */\
+	: [__idx_offset]	"+m" (Xidx_offset)	/* outputs: the template stores into these */\
 	:	[__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
 	,	[__cy]			"m" (Xcy)\
 	,	[__nrt_bits]	"m" (Xnrt_bits)\
 	,	[__nrtm1]		"m" (Xnrtm1)\
-	,	[__idx_offset]	"m" (Xidx_offset)\
 	,	[__idx_incr]	"m" (Xidx_incr)\
 	,	[__odd_radix]   "m" (Xodd_radix)\
 	,	[__half_arr]	"m" (Xhalf_arr)\
@@ -633,12 +630,11 @@
 		/* Store incremented idx offsetL:  */\
 		"movl	%%esi, %[__idx_offset]	\n\t"\
 	"popl %%ebx	\n\t"\
-	:						/* outputs: none */\
+	: [__idx_offset]	"+m" (Xidx_offset)	/* outputs: the template stores into these */\
 	:	[__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
 	,	[__cy]			"m" (Xcy)\
 	,	[__nrt_bits]	"m" (Xnrt_bits)\
 	,	[__nrtm1]		"m" (Xnrtm1)\
-	,	[__idx_offset]	"m" (Xidx_offset)\
 	,	[__idx_incr]	"m" (Xidx_incr)\
 	,	[__odd_radix]   "m" (Xodd_radix)\
 	,	[__half_arr]	"m" (Xhalf_arr)\
@@ -854,11 +850,10 @@
 		"movl	%[__bjmod_0],%%ecx		\n\t"\
 		"movaps	%%xmm0,(%%ecx)			\n\t"\
 	"popl %%ebx	\n\t"\
-		:					/* outputs: none */\
+		: [__wtA]		"+m" (XwtA)	/* outputs: the template stores into these */\
+		 ,[__wtB]		"+m" (XwtB)\
+		 ,[__wtC]		"+m" (XwtC)\
 		: [__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
-		, [__wtA]		"m" (XwtA)		\
-		, [__wtB]		"m" (XwtB)		\
-		, [__wtC]		"m" (XwtC)		\
 		, [__cyA]		"m" (XcyA)		\
 		, [__cyB]		"m" (XcyB)		\
 		, [__bjmod_0]	"m" (Xbjmod_0)		\
@@ -1076,10 +1071,9 @@
 		"movaps		%%xmm1,0x10(%%eax)	\n\t	movaps		%%xmm5,0x50(%%eax)	\n\t"\
 		"movaps		%%xmm0,    (%%eax)	\n\t	movaps		%%xmm4,0x40(%%eax)	\n\t"\
 	"popl %%ebx	\n\t"\
-		:					/* outputs: none */\
+		: [__wtA]		"+m" (XwtA)	/* outputs: the template stores into these */\
+		 ,[__wtB]		"+m" (XwtB)\
 		: [__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
-		, [__wtA]		"m" (XwtA)		\
-		, [__wtB]		"m" (XwtB)		\
 		, [__cyA]		"m" (XcyA)		\
 		, [__cyB]		"m" (XcyB)		\
 		, [__bjmod_0]	"m" (Xbjmod_0)		\
@@ -1609,11 +1603,10 @@
 		"movl	%[__bjmod_0],%%ecx		\n\t"\
 		"movaps	%%xmm0,(%%ecx)			\n\t"\
 	"popl %%ebx	\n\t"\
-		:					/* outputs: none */\
+		: [__wtA]		"+m" (XwtA)	/* outputs: the template stores into these */\
+		 ,[__wtB]		"+m" (XwtB)\
+		 ,[__wtC]		"+m" (XwtC)\
 		: [__data]		"m" (Xdata)	/* All inputs from memory addresses here */\
-		, [__wtA]		"m" (XwtA)		\
-		, [__wtB]		"m" (XwtB)		\
-		, [__wtC]		"m" (XwtC)		\
 		, [__cyA]		"m" (XcyA)		\
 		, [__cyB]		"m" (XcyB)		\
 		, [__bjmod_0]	"m" (Xbjmod_0)		\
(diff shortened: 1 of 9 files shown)


MarkRose:
Pushed another commit here: the same fix for the two IA-32 macros in `imul_macro0.h` that the first commit missed.

`MUL_LOHI64` and `SQR_LOHI64` both carry `: /* outputs: none */` while their templates write `sumhh_`, `sumhl_` and `sumlh_`. They were missed because that block only compiles on 32-bit x86, which nothing in CI built until the job in #291.

The two need different constraints, which is worth spelling out because the obvious uniform fix is wrong for one of them:

- `MUL_LOHI64` never seeds the three in C, and the template writes each one before reading it back, so they carry nothing in — `"=m"`.
- `SQR_LOHI64` *does* seed them in C immediately above the asm (`sumlh_ = lahal_ << 1`, and so on) and the template reads before writing, so they are genuine read-modify-writes — `"+m"`. Giving these `"=m"` would tell the compiler the seeding stores are dead.

Verification on i686, since these two macros are the 64x64 to 128 multiply under every modpow and I did not want to take the change on faith:

- gcc output is byte-identical before and after for `imul_macro.c`, `twopmodq192.c`, `twopmodq96.c` and `mi64.c` — the only diff was objdump's filename header.
- clang output differs only in stack-slot assignment (`0x20(%esp)` becomes `0x2c(%esp)` and so on), same instructions in the same order.
- `-s tt` gives identical Res64 at every FFT length on both trees, including the 100000-pair `2^p (mod q)` self-test that exercises both macros directly.

PR #276, opened 2026-09-07
Share one carry-step threadpool across all radix*_ditN_cy_dif1 routines (follow-up to #70)

MarkRose:
## Summary

Each of the 34 threaded `radix*_ditN_cy_dif1.c` routines holds its own static carry-step
threadpool, created once per leading radix under an `if(tdat == 0x0)` guard and never freed. So
every distinct leading radix a process ever uses leaves `CY_THREADS` sleeping worker threads
behind for the life of the process. #70 fixed the per-assignment leak in `mers_mod_square`,
`fermat_mod_square` and `pm1`, and noted this bounded residual as a follow-up; this is that
follow-up.

It is bounded (at most one pool per radix file) rather than a growth leak, so production runs on
one FFT length are unaffected. It shows up in self-tests and whenever the FFT length changes
mid-run: a `-s tiny -cpu 0:1` run peaks at 45 threads (1 main + 2 FFT-pool + 21 leading radices
x 2), and I have seen a `-s large -cpu 0:15` run sitting at 81.

## Fix

All 34 routines want an identical pool - `CY_THREADS` workers, `CY_THREADS` queue slots, no
per-thread init/shutdown hooks - and only one of them is ever active at a time (the carry step
is called serially from the main thread). So they can share a single pool.

- `threadpool.c/.h`: add `carry_threadpool_get(num_threads, num_cores)`, which creates the
  shared pool on first use and re-creates it (after freeing the old one, which is quiescent
  between carry calls) only if the requested thread count changes.
- The 34 radix files: replace `tpool = threadpool_init(CY_THREADS, MAX_THREADS, CY_THREADS,
  &thread_control)` with `tpool = carry_threadpool_get(CY_THREADS, MAX_THREADS)` (also in the
  `#if 0` macOS variant, for consistency), and drop the now-unused per-file `thread_control`
  static. Mechanical, identical in every file.

No change to dispatch or drain; each file still uses its `tpool` pointer exactly as before.

## Verification

AVX2 build (gcc, `-O3 -flto`), two threads (`-cpu 0:1`), thread count sampled from
`/proc/<pid>/task` every 50 ms:

| run | main | this PR | residues |
| --- | --- | --- | --- |
| `-s t` (21 distinct leading radices, 24 pool inits) | peak **45** threads | peak **5** | 93/93 pass on both |
| `-f 22 -fft 256 -iters 100` (all 7 radix sets, 5 leading radices) | peak >= 9 | peak **5** | all 7 Res64 identical to main (`9201143390F3828D`) |

5 = 1 main + 2 FFT-pool workers + 2 carry workers, i.e. the steady state no longer depends on
how many leading radices the process has touched.

Not tested here: Windows/macOS builds (no code path differs; the pool shape is unchanged).

### DIFF
--- a/src/radix1008_ditN_cy_dif1.c
+++ b/src/radix1008_ditN_cy_dif1.c
@@ -426,7 +426,6 @@ int radix1008_ditN_cy_dif1(double a[], int n, int nwt, int nwt_bits, double wt0[
 	static int pool_work_units = 0;
 	static struct threadpool *tpool = 0x0;
 	static int task_is_blocking = TRUE;
-	static thread_control_t thread_control = {0,0,0};
 	// First 3 subfields same for all threads, 4th provides thread-specifc data, will be inited at thread dispatch:
 	static task_control_t   task_control = {NULL, cy1008_process_chunk, NULL, 0x0};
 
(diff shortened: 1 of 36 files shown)


ldesnogu (approved):
LGTM

tdulcet (approved):
Thanks for the follow up. This looks good to me.

PR #278, opened 2026-09-07
Report the physical cores and cache sizes behind the selected CPU set at startup

MarkRose:
## Why

`-cpu` indices are the operating system's logical-processor numbers, and which of them are SMT
siblings of one physical core differs by OS and vendor. Linux on Intel usually numbers siblings
as i and i+ncores, so `-cpu 0:3` is four physical cores; Windows, FreeBSD and most AMD systems
number siblings adjacently, so the same `-cpu 0:3` is two physical cores with both hyperthreads
each. Mlucas passes the numbers straight through on every backend (`sched_setaffinity`,
`cpuset_setaffinity`, `SetThreadGroupAffinity`, macOS affinity tags), so a user, or a scaling
report, cannot tell from the command line what was actually run. Only the hwloc `-core lo:hi:tpc`
form is topology-aware, and hwloc is off by default in makemake.sh.

## What it does

After command-line parsing, when `-cpu`, `-nthread` or `-core` was given, print:

    INFO: 4 logical CPUs selected on 4 physical cores (up to 1 thread per core).
    INFO: cache: L2 = 256 KB per core, L3 = 16384 KB.

and, when selected CPUs share a core and the set did not come from `-core` (where sharing is an
explicit choice):

    WARN: some selected logical CPUs are SMT siblings on the same physical core. ...

Sources: hwloc when built in (consistent with the PU indexing threadpool.c binds to), Linux sysfs
(`topology/core_id`, `physical_package_id`, `cache/indexN`), Windows
`GetLogicalProcessorInformationEx` (RelationProcessorCore / RelationCache). Anything else says
the topology is not available and never fails. The detected sizes are kept in two new globals,
`L2_CACHE_BYTES` and `L3_CACHE_BYTES` (0 = unknown), for later use by radix selection.

help.txt gains a note on logical-CPU numbering under the `-cpu` description.

## Verification

AVX2 build on Linux (i9-10885H, siblings are i and i+8):

| command | output |
| --- | --- |
| `-cpu 0:3` | 4 logical CPUs on 4 physical cores, 1 thread per core; no warning; L2 = 256 KB, L3 = 16384 KB |
| `-cpu 0,8` | 2 logical CPUs on 1 physical core, 2 threads per core; warning printed |
| `-cpu 0` | 1 logical CPU on 1 physical core |
| `-nthread 2` | reports as for `-cpu 0:1` |
| no flag | nothing printed (single thread on CPU 0, nothing to report) |

The Windows branch was cross-compiled with x86_64-w64-mingw32-gcc; it has not been run on
Windows. FreeBSD and macOS report "topology not available".

## Relation to other work

Independent of #176 (default core set from an external affinity mask), which changes what goes
into the set; this reports on whatever ends up in it. Follows #277 (thread count in cfg entries).

### DIFF
--- a/help.txt
+++ b/help.txt
@@ -455,10 +455,20 @@ numbering conventions used by the leading x86-CPU manufacturers Intel and AMD.
 	setting syntax in which each of the comma-separated 'triplet' subfields is in the above
 	form and, analogously to the one-triplet-only version, no whitespace is permitted within
 	the colon-and-comma-separated numeric field. Thus '-cpu 0:3,8:11' and '-cpu 0:3:1,8:11:1'
+	(see also the note on logical-CPU numbering after this list)
 	both specify an 8-threaded run with affinity set to logical core quartets 0-3 and 8-11,
 	whereas '-cpu 0:3:2,8:11:2' means run 4-threaded on cores 0,2,8,10. As described for the
 	-nthread option, it is an error for any core index to exceed the available number of logical
 	processor cores.
+
+	NOTE on logical-CPU numbering: the -cpu indices are the operating system's logical-processor
+	numbers, and which of those are hyperthread (SMT) siblings of one physical core differs by OS
+	and vendor. Linux on Intel usually numbers siblings as i and i+(number of cores), so '-cpu 0:3'
+	is four physical cores; Windows, FreeBSD and most AMD systems number siblings adjacently, so the
+	same '-cpu 0:3' is two physical cores with both hyperthreads each. At startup Mlucas prints how
+	many physical cores the selected set spans and warns when siblings share a core; use '-core
+	lo:hi:1' (hwloc builds) or list one logical CPU per core (e.g. '-cpu 0:6:2') for one thread per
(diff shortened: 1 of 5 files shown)


MarkRose:
Pushed a follow-up: `L2_CACHE_BYTES` and `L3_CACHE_BYTES` are now defined in `util.c` rather than `Mlucas.c`.

They are read only by `detect_cache_sizes()` and `report_cpu_topology()`, both in `util.c`, and Mfactor links `util.o` without `Mlucas.o` - so that build had the references and no definition. The default build hides it, because LTO drops the unreferenced reads before the link. Every build without LTO fails:

```
src/util.c:9611: undefined reference to `L2_CACHE_BYTES'
src/util.c:9664: undefined reference to `L3_CACHE_BYTES'
```

That is the AddressSanitizer jobs, which build at `-Og` with no LTO, and every `Linux old` cell whose clang is old enough for makemake.sh to turn `-flto` off - alpine:3.8/3.12/3.16 and ubuntu:14.04/16.04/18.04. Verified locally by building Mlucas and Mfactor with the ASan job's exact CFLAGS, both clean.

PR #282, opened 2026-09-08
Cast to unsigned char before passing to the ctype functions (19 sites)

MarkRose:
`isspace()`, `isalpha()`, `isdigit()` and `isxdigit()` take an `int` whose value must be representable as `unsigned char`, or equal `EOF` (C11 7.4p1). Passing a plain `char` breaks that wherever `char` is signed — which is every platform Mlucas targets except ARM — because any byte above `0x7F` arrives as a *negative* int, and glibc implements these as a lookup into a classification table indexed by that value. A negative index reads before the start of the table.

This is not a theoretical exposure. All 19 sites are in parsing paths, and the data reaching them is externally supplied:

| file | sites | what it parses |
| --- | --- | --- |
| `src/Mlucas.c` | 15 | `worktodo.txt` lines, `mlucas.ini` contents, `argv` |
| `src/get_preferred_fft_radix.c` | 2 | `mlucas.cfg` line tokenising |
| `src/mi64.c` | 2 | `convert_base10_char_mi64()` |

A single stray UTF-8 byte in a hand-edited work file is enough — an en-dash pasted in place of a hyphen, say.

The `mi64.c` pair is the clearest instance, since the type is right there:

```c
char c;
...
c = char_buf[i];
if(!isspace(c)) {
```

## What is deliberately *not* changed

A naive grep suggests 25 sites. Six are false positives and are left alone:

- five in `util.c` are the string `"... isdigit(c) fails, s = %s ..."` inside `fprintf` format arguments, not calls;
- one in `get_preferred_fft_radix.c` is inside a comment describing the loop below it.

Ten further call sites elsewhere in the tree already had the cast, so this brings the codebase to consistent treatment rather than introducing a new convention.

## Verification

- `avx2` builds clean.
- `-s tt` gives 23 residues, identical to `main`.
- Worktodo parsing compared against `main` byte-for-byte on three inputs: a line with leading whitespace, a bare exponent, and a line beginning with a multi-byte UTF-8 character. All three behave identically — the third being precisely the input that is undefined behaviour before this change.

Found while auditing an old review document against current `main`; the other findings in it had already been fixed or are covered by open PRs.

### DIFF
--- a/src/Mlucas.c
+++ b/src/Mlucas.c
@@ -569,12 +569,12 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
 		fprintf(stderr," %s entry: %s\n",WORKFILE,g_in_line);
 		/* Skip any whitespace at beginning of the line: */
 		char_addr = g_in_line;
-		while(isspace(*char_addr)) {
+		while(isspace((unsigned char)*char_addr)) {
 			++char_addr;
 		}
 		// v20.1.1: Parse all lines whose 1st non-WS char is alphabetic; print "Ignoring (copy of workfile line)" for all entries not so.
 		// NB: Discontinue support for numeric leading char, i.e. Mersenne-exponent-only legacy format:
-		if(!isalpha(*char_addr)) {
+		if(!isalpha((unsigned char)*char_addr)) {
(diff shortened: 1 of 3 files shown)


tdulcet (approved):

Issue #281, opened 2026-09-08
Audit the dead/disabled code: 350 #if 0 blocks and 5 parked source files


MarkRose:
Splitting this out of the discussion on #144, where the question came up in a concrete case.

Mlucas carries a lot of code that is not compiled into any build. Some of it is finished work parked behind a gate that could be reopened, some is half-written, some was superseded years ago and never deleted. At the moment there is no way to tell which is which without reading each block, so every encounter with a piece of it becomes an individual judgement call — which is exactly what happened on #144.

The proposal is to audit it once, record what each piece is, and then act on that rather than case by case.

## What is actually there

Counted on `main`:

| category | count |
| --- | --- |
| `#if 0` blocks in `src/*.c` and `src/*.h` | **350**, across 80 files |
| whole source files parked with a `.txt` suffix so they are never compiled | 5 |
| files mentioning "unused" | 60 |
| files carrying TODO/TO DO notes | 69 |

The `#if 0` blocks are concentrated but widespread — `util.c` has 18, `imul_macro1.h` 16, `mi64.c` 10, and a long tail of nine or fewer across the radix files.

The five parked files:

```
src/dft_sine_term_opt.c.txt
src/gcd_lehmer.c.txt
src/mi64_new.c.txt
src/qfcheb.c.txt
src/test_fft_radix.c.txt
```

## They are not all the same kind of thing

A few examples that came up while counting, which suggest the categories:

- **Waiting on an external condition that may since have been met.** `twopmodq96.c`: `#if 0//YES_ASM  *** Disable until regain GCC 4.6+ functionality ***`. GCC 4.6 was 2011. Whatever the blocker was, it is worth establishing whether it still exists.
- **Waiting on internal work.** `factor.c`: `#if 0 /* disable for now - need to sync with similar code in main() */`.
- **Superseded by a design change.** `Mlucas.c`: `#if 0//def USE_SSE2  *** v21: Removed array padding for FFT < 32K, disabled this "skip"`.
- **Parked because of a toolchain problem.** `twopmodq64_test.c`: `#if 0 // Disable until fix slow-compile issue with the twopmodq78_q4_GPU source code`.

So the outcome of an audit is not one decision but a classification: **revive**, **delete**, or **keep parked with a recorded reason and a condition for revisiting**.

## Why it matters beyond tidiness

[...]

Issue #245, opened 2026-08-04
Check for missing Mlucas v21 code


tdulcet:
The Mlucas v21 code George recovered from Ernst's systems included many Vim savefiles, hidden files and other stuff that we initially ignored when we imported the code into this repository. We should double check each of those files to make sure that there is no missing Mlucas v21 code or other potentially useful information as we continue to work on v21.

There were also several `*.c.txt` files recovered, which we did include in the import, but in most cases they seem to be largely duplicates of the same file without the `.txt` extension. We should determine why there are two copies of these files and consider removing one of them and reconciling the differences.

There were several other files that we included in the v21 import that are not currently used, some of which do not compile. For example, there are a couple of files with `32` in the filename, which seem to be for 32-bit SIMD support, but Mlucas no longer has 32-bit SIMD support. In general, we should audit the codebase to determine the minimal set of files required to build Mlucas and Mfactor, and then for any remaining files, we should determine and document what they are used for. Some of them might be useful for testing/debugging, or they could be for uncompleted features or side projects Ernst was working on.

@MarkRose - I thought this might be a task that Claude could easily automate. The files George recovered from Ernst's systems are available in [this forum post](https://www.mersenneforum.org/node/22312?p=874148#post874148) and the one below (also see posts 21 and 22 further down that thread). We originally imported the v21 code in https://github.com/primesearch/Mlucas/commit/c9fef37ac435d7989145e801288f16de99208696 and https://github.com/primesearch/Mlucas/commit/77749764e9f49906e0208e05bc5ba0c375a044a9.

tdulcet:
I attempted to recover some more of Ernst's v21 changes in https://github.com/primesearch/Mlucas/commit/d9de83ee9e0f217028a031061936622f84057d56. This is currently on the [`ernst` branch](https://github.com/primesearch/Mlucas/tree/ernst). It still needs the full audit described above to confirm that this is correct and check that there is no additional missing code.

MarkRose:
I had a look at the `ernst` branch (`d9de83e` plus the `d653116` fixup) as a first pass at the audit. One demonstrated regression, and one thing I could not confirm and so am flagging rather than claiming.

The substance of the recovery, for anyone reading along, is Ernst moving the iteration intervals to powers of two:

| | before | after |
| --- | --- | --- |
| `ITERS_BETWEEN_GCHECK_UPDATES` | 1000 | 1024 |
[...]

PR #283, opened 2026-09-09
No longer build files that are unused

tdulcet:
* Changed the `makemake.sh` script to no longer build C files that are unused by Mlucas and/or Mfactor. Part of #245
  * This should speed up the build times and prevent build breakages in dead code, but also (unfortunately) reduces the CI coverage

The result:
* Mlucas: 99 -> 81 files
* Mfactor: 22 -> 19 files

These 18 files are no longer built for Mlucas:
```
fgt_m61.c
pairFFT_mul.c
radix104_ditN_cy_dif1.c
radix112_ditN_cy_dif1.c
radix120_ditN_cy_dif1.c
radix16_pairFFT_mul.c
radix17_ditN_cy_dif1.c
radix31_ditN_cy_dif1.c
radix72_ditN_cy_dif1.c
radix80_ditN_cy_dif1.c
radix88_ditN_cy_dif1.c
radix96_ditN_cy_dif1.c
twopmodq100.c
twopmodq128_96.c
twopmodq160.c
twopmodq64_test.c
twopmodq80.c
twopmodq96.c
```
and these 3 files are no longer built for Mfactor:
```
get_fft_radices.c
twopmodq100.c
twopmodq160.c
```

As part of #245 and #281, we should consider adding some of these files back and actually using them.


MarkRose (dismissed):
LGTM. Everything still builds with these gone - I checked all five Mlucas modes (`nosimd`, `sse2`, `avx`, `avx2`, `avx512`, 81 objects each) and all six Mfactor word variants, and every one links. I also chased the cross-references to the removed files and they are all inert: the `case 112:`/`case 120:` dispatch lines at `Mlucas.c:6445-6446` are commented out, the `pairFFT_mul` mentions in `radix16/32_wrapper_ini.c` are comment text, `INCLUDE_TF` is hard-set to `0` in `Mlucas.h:41` so `test_fac()` never links, and the rest are `Mlucas.h` declarations plus `test_fft_radix.c`, which is in neither OBJS list.

No runtime change either: `-s tt` and `-s t` on the `avx2` build behave identically to the parent commit built the same way, deterministically over repeated runs.

One thing worth weighing before this lands, on top of the coverage caveat you already noted - several files here are being actively modified by open PRs, which would stop being compile-tested:

| File | Open PRs modifying it |
| --- | --- |
| `radix72/80/88/96/104/112/120_ditN_cy_dif1.c` | #211 |
| `pairFFT_mul.c` | #211, #187 |
| `fgt_m61.c` | #213 |
| `twopmodq100.c` | #273, #227, #213 |
| `twopmodq64_test.c`, `twopmodq80.c`, `twopmodq96.c`, `twopmodq128_96.c` | #273, #227, #213 |

Two of those are worth a moment:

* **#211** touches eight of the removed files, so nearly all of it would go uncompiled.
* **#187**'s only remaining live hunk is the `pairFFT_mul.c` one - its other defect already landed via #146 (`7189795`) - so that PR would lose its compile coverage entirely.

The asymmetry helps a bit: `twopmodq64_test/80/96/128_96.c` stay in `OBJS_MFAC` and keep coverage through Mfactor. Only `twopmodq100.c`, `twopmodq160.c`, the radix files, `pairFFT_mul.c`, `radix16_pairFFT_mul.c` and `fgt_m61.c` lose it completely.

None of that is an argument against the change - it is the tradeoff you already called out - but it suggests either landing #211 and #187 first, or deciding as part of #245/#281 that those radices are going away, in which case those two PRs should be closed rather than quietly stop being built.

**Update:** the 18-file figure above is now too pessimistic. Teal's `test_fft_radix.sh` in #281 compiles *and runs* all nine dropped `radix*_ditN_cy_dif1.c` files, and four `twopmodq*` files stay in `OBJS_MFAC`, so only five lose coverage entirely - `fgt_m61.c`, `pairFFT_mul.c`, `radix16_pairFFT_mul.c`, `twopmodq100.c`, `twopmodq160.c`. All five compile clean today in both nosimd and avx2, so the dev-target idea below would start green. That also means #211's exposure is largely covered and #187 is the PR that still depends on sequencing. Details in the comment thread.

Separately, and pre-existing rather than anything to do with this PR: the `avx2` build on the parent commit reports `WARNING: 0 of 2 radix-sets at FFT length 1 K` and `WARNING: 0 of 1 radix-sets at FFT length 3 K`, so `-s tt` skips both lengths. The 1K case is #7, with #69 open as the fix. The 3K case is milder than the warning makes it sound - AVX2 computes 3K *correctly* (Res64 `1EC9A4159C4067BC`, matching a nosimd build exactly), it just exceeds the self-test's 100-iteration roundoff gate at `MaxErr = 0.4375` vs the 0.40625 limit, and AVX2 has no second radix set at that length to fall back on. Noted on #69 rather than raised separately.

Issue #95, opened 2026-07-13
Add support for generating proof files and certification assignments


tdulcet:
Ernst had planned to add support for generating PRP proof files in Mlucas v21, both for Mersenne and Fermat numbers. He wrote detailed notes about this, which are included in the `docs` directory. For example, see the [docs/prp_proof.txt](https://github.com/primesearch/Mlucas/blob/main/docs/prp_proof.txt), [docs/prp_proof_examples.txt](https://github.com/primesearch/Mlucas/blob/main/docs/prp_proof_examples.txt) and [docs/todo.txt](https://github.com/primesearch/Mlucas/blob/main/docs/todo.txt) files. This is the biggest blocker preventing Mlucas from being used in production wavefront primality testing. The PrimeNet server now only provides half of the GHz-days credit for PRP assignments without a proof, and first time LL assignments receive no credit.

We should also add support for self-verifying the PRP proof files, as well as certification (CERT) assignments. This would be needed to verify the PRP proofs for exponents above Prime95/MPrime's exponent limit, since Mlucas supports much larger exponents than other software. This could also make testing the larger Fermat exponents much more practical.

PRP proofs are based on Verifiable Delay Functions (VDF, see [research paper](https://eprint.iacr.org/2018/627.pdf)). For more information, see the [original thread](https://mersenneforum.org/showthread.php?t=24654) and [subsequent thread](https://mersenneforum.org/showthread.php?t=25638) on the Mersenne Forum. We should support the version 2 proof files. Prime95/MPrime, GpuOwl/PRPLL and PrMers already support PRP proofs, and could be used as a reference.

AutoPrimeNet already supports uploading PRP proof files and downloading the starting value for certification assignments, which could be used to test any implementation. The proof files are expected to be found in a `*.proof` file in the `proof` directory, while the starting value is saved to a `<exponent>.cert` file in the working directory, for example `M136279841.cert`.

PR #288, opened 2026-09-12
Recovered more of Ernst's v21 changes from editor savefiles

tdulcet:
* Recovered more of Ernst's v21 changes from his editor savefiles
  * Part of #245
  * Part of #95
* Copied Ernst's changes from `test_fft_radix.c.txt` to `test_fft_radix.c` and removed the former.

Note that the `docs/a.txt` and `docs/b.txt` files seems to be scratch files. This PR completely rewrites them with new content. It is unclear if the existing or new content would be more useful, but maybe the Mfactor savefile code could be helpful for #114.

I am marking this PR as a draft, as it includes changes to the default number of iterations that appear to be unfinished and untested by Ernst. I dropped my previous fixup commit (https://github.com/primesearch/Mlucas/commit/d6531160bb96ea65afa4f3690a8ab17cf253e075), as it may have been incorrect. See https://github.com/primesearch/Mlucas/issues/245#issuecomment-5605353130 and the below comments. 

@MarkRose - Please feel free to take over this PR and finish it.


MarkRose (commented):
Reviewed, and happy to take it over. One blocking defect, and it turns out **your fixup commit was necessary** - dropping it is what introduces the problem.

## Blocking: this aborts every PRP run with more than 4 threads

Built this branch and ran a plain PRP assignment on 8 threads:

```
Setting ITERS_BETWEEN_CHECKPOINTS = 65536.
Initial DWT-multipliers chain length = [medium] in carry step.
Assertion 'i%k == 0 && k%j == 0' failed: G-checkproduct update interval must divide
savefile-update one, which must divide the G-check interval
exit 134
```

Current `main` runs the same assignment to completion (exit 0).

The cause is the clamp at `Mlucas.c:1396`, which your fixup changed and this branch no longer does:

```c
if(USE_SHORT_CY_CHAIN)
    ITERS_BETWEEN_CHECKPOINTS = MIN(ITERS_BETWEEN_CHECKPOINTS,10000);
```

With the new power-of-two defaults, `MIN(65536, 10000)` is **10000**, and 10000 is not a multiple of 1024:

| `ITERS_BETWEEN_CHECKPOINTS` | `1048576 % k == 0` | `k % 1024 == 0` | result |
| --- | --- | --- | --- |
| 8192 | yes | yes | ok |
| 65536 | yes | yes | ok |
| **10000** | **no** | **no** | **assertion, PRP/Pépin abort** |

It bites on a very ordinary configuration: more than 4 threads (so the default is 65536) and an exponent close to `pmax_rec` (so `USE_SHORT_CY_CHAIN` is set and the clamp applies). On `main` the same clamp is harmless because the values there are decimal - `MIN(100000,10000) = 10000`, and `1000000 % 10000 == 0`, `10000 % 1000 == 0`.

Restoring your fixup fixes it, verified on the same test:

| build | PRP, 8 threads |
| --- | --- |
| this branch | **exit 134**, assertion |
| `main` | exit 0 |
| this branch + `MIN(...,8192)` | **exit 0**, checkpoints at 8192, 16384, 24576, 32768 … |

So `d6531160` was right, and 8192 is the correct value rather than an arbitrary one: it is the small-thread default, so the clamp becomes a no-op there and lowers 65536 to a value that still satisfies both divisibility constraints. I would restore it.

## The test_fft_radix.c side is sound

I checked this against Ernst's original savefiles when auditing the recovery on #245, and it holds up here:

- The three comparator hunks Ernst had in `test_fft_radix.c~` but which did not make it across - the ones rewriting `qsort` predicates to `return ar - br` - are **still correctly excluded**. They return a `double` difference from an `int`-returning comparator, so with `EPS = 1e-10` any pair differing by between 1e-10 and 1.0 truncates to 0 and the sort breaks. Good that they stayed out.
- Folding `test_fft_radix.c.txt` into `test_fft_radix.c` and deleting the former is right, and it also settles one of the five parked `.c.txt` files in #281.
- No conflict with #289 (`git merge-tree` reports 0), which restores `t2`/`t3` after my regression in #287. Worth noting the new `#error TEST_FFT_RADIX requires non-SIMD build!` makes the `TTYPE == 2 && USE_SSE2` arm unreachable, so once both land, half of #289's guard condition becomes dead - harmless, and I will tidy it when this merges.

## Two things still outstanding, neither blocking

Both from the #245 audit, repeated here so they do not get lost:

1. **The every-10M savefile archiving becomes inert.** `Mlucas.c` gates it on `ilo % 10000000 == 0`, and `ilo` only ever takes multiples of `ITERS_BETWEEN_CHECKPOINTS`. 10,000,000 = 2⁷ × 78125, so no power of two above 128 divides it - the `.NM` files simply stop being written. Given your PRP-proof reading, the right answer is probably not a different constant but proof-point-driven archiving under #95; the thing to avoid is landing this with the behaviour silently gone and no note at the gate.
2. **The `CheckInterval` contract is now inconsistent.** The assertion text still says `must >= 1000` while the valid set is powers of two from 1024 up. A stale `CheckInterval = 10000` in an existing `mlucas.ini` will now trip the same divisibility warning on every run.

## On `docs/a.txt` and `docs/b.txt`

I have no view on which content is more useful - they are scratch files either way. One observation: if the Mfactor savefile material in there is what you were thinking of for #114, that PR is already complete and verified (it fixes both the savefile corruption and, in its third commit, an AVX-512 sieve SIGSEGV), so it does not need it. It is the single highest-value merge for CI right now - 22 of the 36 failing jobs on `main`, per #224.

I will push the fixup restore and the two notes unless you would rather keep the PR yours.

PR #291, opened 2026-09-12
Enabled -Wextra and updated CI to remove remaining error masking

tdulcet:
* Updated the `makemake.sh` script to enable `-Wextra`
* Updated the `makemake.sh` script to only support NOSIMD builds on 32-bit x86 systems
* Updated the CI. Follow up from #103
  * Removed remaining error masking for TSan jobs and on Windows
  * Added Linux 32-bit job to test the NOSIMD builds on both the x86 and ARM 32-bit platforms using a Docker container of Debian 13
  * Updated the "Linux old" job to run minimal Mlucas/Mfactor self-tests and upload the artifacts
  * Updated the FreeBSD job to run the PRP self-test for consistency with the other jobs
  * Updated the FreeBSD job to additionally test on FreeBSD 10 and 11
  * Changed TSan jobs to run NOSIMD builds for consistency with the ASan jobs
  * Switched the Windows ARM job to use native Clang instead of emulated GCC now that MSYS2 is provided. Fixes #290
* Cleaned up formatting


MarkRose (dismissed):
Approved, but note this will turn 33 more jobs red until the fixes from other PRs are merged.

tdulcet:
I previously created a PR to the GitHub Action we use for the FreeBSD VMs to add support for FreeBSD 10 and 11: https://github.com/anyvm-org/freebsd-builder/pull/3. Since they merged it, I updated this PR to test with FreeBSD 10 and 11. I tried updating this PR to test with FreeBSD on ARM as well, but it was too slow due to the CPU emulation.

MarkRose:
Thanks - the probe block and its `|| echo "::warning::"` wrapper are gone, along with the two stale comments about emulation, so my earlier concern is fully addressed. And the switch does what you hoped: the windows-11-arm cell is now `clang version 22.1.8`, `Target: aarch64-w64-windows-gnu`, native rather than emulated, and it picks `NOSIMD` correctly without any predefined-macro probe. Nice result - #290 can be treated as the stopgap it was.

One thing to sort out before this lands, though. This run is 69 failed / 7 passed against 34 / 40 for the same workflow on `ad31b81` (main), so it adds 33 newly-red jobs, and all 33 are `Mlucas Linux old` cells.

They are not the `-Wextra` change and not the mask removal. Every one I opened dies at the same new line, the Mfactor smoke run this PR adds:

```
+ cd ../obj_mfac
+ ./Mfactor -m 521 -bmax 20
```

which trips the Mfactor savefile bug:

```
ERROR: 'passnow' not found in Line 10 of factoring restart file t521!
ERROR: Function PerPass_tfSieve, at line 4021 of file ../src/factor.c
Assertion '!i' failed: There were errors writing the savefile ... aborting
```

That reproduces on unmodified `main` locally - `bash makemake.sh mfac && ./Mfactor -m 521 -bmax 20` aborts with SIGABRT, at 1 thread and at `-nthread 4` alike - so it is pre-existing and this PR is only the first thing to call it. It is what #114 fixes: same command on #114's head exits 0 and reports `M(521) has 0 factors in range k = [0, 16336320], passes 0-15`.

So the sequencing is just **#114 before this one**, and I would expect those 33 to go green on a rerun afterwards. I sampled four of the 33 (`ubuntu:14.04` gcc, `alpine:3.20` clang, `ubuntu:20.04` gcc on arm, and windows-11-arm) and all four stop at that line - windows-11-arm exits 127 there rather than 134, but at the same command, so it is the same blocker rather than anything to do with the toolchain switch.

Two smaller notes:

The `gcc: error: unrecognized command line option '-mavx512f'` lines in the old-distro logs look alarming but are fine - that is the AVX-512 capability probe doing its job, and the job goes on to emit `::warning::Skipping the AVX512 build mode`. Worth knowing before anyone chases them.

FreeBSD 10 and 11 fail, but so do 12 through 15, and 12-15 already fail on main, so the new cells are not a regression - they are joining an existing failure rather than adding a class. Same question as always applies to whether EOL releases earn their CI minutes, but that is your call and it does not block anything.

tdulcet:
I added a 32-bit job to test the NOSIMD builds on both the x86 and ARM 32-bit platforms. This is our first CI coverage of the 32-bit builds and it uncovered some additional warnings that were missed in #34. It only tests with GCC on Linux for now, but we could later add Clang or Windows if it would be useful. I used a Docker container of Debian 13, as the 64-bit GitHub runners do not support building 32-bit programs and Ubuntu has not supported 32-bit x86 since 18.04 LTS.

ldesnogu:
I indeed never tested 32-bit as I had assumed it was obsolete and unsupported, even in no SIMD mode.

MarkRose (dismissed):
## FreeBSD loses `pipefail`, which re-masks failures

```diff
-        set -ex -o pipefail
+        set -ex # -o pipefail
```

The self-tests in that job are piped through `tee -a test.log`, so with `pipefail` off a failing `Mlucas -s` now exits 0 on every release, not just on 10 and 11. That is the one place in the PR that moves away from its own stated goal.

`pipefail` is working on 12 through 15 today — FreeBSD 13 on main gets all the way to the Mfactor savefile assert (patched by #114 ), so `set -o pipefail` is not what is failing there. Could it be conditioned on the release, or the script run under the bash the job already `pkg install`s?

## `Linux old` running self-tests is right, and it confirms an open bug

Worth knowing what it signs us up for. On your 09-16 run, 69 of 76 jobs failed against main's 34 of 74, and essentially the whole delta is the ~36 `Linux old` cells that used to be build-only.

None of that is your doing. Those cells now reach the 1K FFT self-test, and the crash they hit is issue #7 — same exponent, same radix set, same routine:

```
M22679: using FFT length 1K ... Using complex FFT radices        32        16
Segmentation fault      (core dumped) ./Mlucas -s tt -cpu 0:3
```

#69 fixes it: the AVX/AVX-512 `radix16|32_wrapper_square` routines read `RE_IM_STRIDE` sincos-index sets per SIMD pass, and at the smallest FFT lengths the final-block schedule leaves a partial chunk with fewer sets than that, so they read past the end of `index[]`.

What makes this a useful data point for #69 is that your job reproduces the *other* half of that bug's behaviour at the same time. Where the overrun lands on allocator slack it does not crash, it silently produces garbage twiddles and the length fails on roundoff instead. That is what main's own Linux job has been printing all along, and I get the identical thing locally in an `ubuntu:20.04` container with gcc 9.3 and on the host with gcc 16.2:

```
M22679 Roundoff warning on iteration 1, maxerr = 0.498733520508
***** Excessive level of roundoff error detected - this radix set will not be used. *****
WARNING: 0 of 2 radix-sets at FFT length 1 K passed - skipping it.
```

Same bug, two faces, decided by whether the read runs off the end of a mapped page. So un-masking these cells earned its keep — it confirms #69 again.

Note this PR will turn main from about 34 red cells to about 70, pending those other fixes.

## Nits

`try_lto()` losing its `${1:--flto}` default is harmless today since both call sites pass a flag explicitly, but it makes the function quietly wrong if anyone ever calls it bare. Might be worth keeping the default.

I did check `bash -s -e -o pipefail` — `errexit` and `pipefail` both come out on and the script does abort correctly, so no concern there.

tdulcet:
> `pipefail` is working on 12 through 15 today — FreeBSD 13 on main gets all the way to the Mfactor savefile assert (patched by #114 ), so `set -o pipefail` is not what is failing there.

The problem is that FreeBSD uses sh and `pipefail` is undefined in POSIX sh. It seemed to work anyway in later FreeBSD releases, but it produces a hard error in FreeBSD 10, so I had to disable it. I will update this PR to instead enable it conditionally, but note that this would of course still not help with FreeBSD 10. There is a similar issue in #298.

MarkRose (approved):

tdulcet:
As you saw, I made the requested change to the FreeBSD job. I also updated the `makemake.sh` script to disable the SIMD build mode detection on 32-bit x86 systems, to reflect that only the NOSIMD builds are supported. Thanks for the re-approval.

Looking through the CI logs from the new 32-bit job, I noticed a lot of self-test bad residue failures. It is unclear if this is already fixed by the pending PRs, so I opened #306 to track the issue.

PR #292, opened 2026-09-13
Make the power-of-two intervals consistent with the code around them

MarkRose:
Finishes #288, as requested there. Targets the `ernst` branch rather than `main`, so it lands underneath #288 rather than alongside it.

All three are regressions that the interval change introduces against current `main` - none is pre-existing - so by the standard you set on #288 ("should not break any existing functionality") all three are blocking.

## 1. The `USE_SHORT_CY_CHAIN` cap was left at 10000

```c
if(USE_SHORT_CY_CHAIN)
    ITERS_BETWEEN_CHECKPOINTS = MIN(ITERS_BETWEEN_CHECKPOINTS,10000);
```

With the new defaults that makes `MIN(65536,10000)` = **10000**, which is not a multiple of 1024, so the chain `ITERS_BETWEEN_GCHECK_UPDATES | ITERS_BETWEEN_CHECKPOINTS | ITERS_BETWEEN_GCHECKS` breaks and the PRP/Pépin setup asserts.

That is **every PRP run with more than 4 threads on an exponent near `pmax_rec`** - i.e. an ordinary assignment. Measured with a plain `PRP=1,2,86243,-1,75,0` on 8 threads:

| build | result |
| --- | --- |
| `main` | exit 0 |
| #288 as it stood | **exit 134**, `Assertion 'i%k == 0 && k%j == 0' failed` |
| with this PR | exit 0, checkpoints at 8192, 16384, 24576 … |

So `d6531160` was right and dropping it is what introduced this. Restored, with 8192 rather than 10000 because it is the small-thread default and satisfies both divisibility constraints - the cap has to itself be a legal savefile-update interval.

## 2. The every-10M savefile archiving became unreachable

Gated on `ilo % 10000000 == 0`, and `ilo` only ever takes multiples of `ITERS_BETWEEN_CHECKPOINTS`. Since 10⁷ = 2⁷ × 78125, no power-of-two interval above 128 divides it, so the `.NM` archives would simply stop being written - silently.

Now triggers on **crossing** a 10-million boundary, which is the same "first checkpoint past x million" the existing comment already describes:

```c
if((ilo >= (uint32)ITERS_BETWEEN_CHECKPOINTS) && (ilo/10000000 > (ilo - ITERS_BETWEEN_CHECKPOINTS)/10000000)) {
```

Verified by enumeration rather than by reading:

- **Backward compatible**: zero differences from the old test for every pre-v21 interval (1000, 10000, 100000) across the first 100M iterations. Where the interval divides 10⁷ the two tests are the same test.
- **Restores the behaviour**: fires exactly once per boundary for 1024, 8192, 65536 and 1048576, where the old test fires never.

One thing that turned up while checking it: the archive name came from `ilo`, and at the largest valid interval the first checkpoint past a boundary can land in the *following* million - 280M would have been written as `.281M`. The name now derives from the boundary instead, which is identical for the pre-v21 intervals since those always landed exactly on it. Checked across all seven intervals × 500 boundaries: no mismatches.

The `ilo >= ITERS_BETWEEN_CHECKPOINTS` guard is there because the subtraction would otherwise wrap for `ilo` below one interval. Unsigned wrap is well-defined and the comparison still came out false, but it is not something a reader should have to derive.

## 3. A user-set `CheckInterval` had no way to satisfy the new constraints

The minimum was still asserted (and documented) as 1000, while the valid set is now powers of two from 1024 up. Anyone whose `mlucas.ini` carries a pre-v21 value - 10000 being the obvious one, since it was the old default - would hit the assert.

Rather than abort, round down to the largest valid power of two and report it:

| `CheckInterval` | result |
| --- | --- |
| 10000 | → 8192, with an `INFO:` line |
| 100000 | → 65536, with an `INFO:` line |
| 2000 | → 1024, with an `INFO:` line |
| 1024 | unchanged, silent |

All four run to completion. The assertion text is corrected to 1024.

## Verification

On top of #288's branch: production PRP at 8 threads exits 0 (was 134); `-s tt` and `-prp -s tt` both exit 0; the four `CheckInterval` values above all complete with the rounding shown. The two `WARNING: 0 of` lines in `-s tt` are the pre-existing AVX2 1K/3K cases (#7/#69), identical to `main`.

Not covered: I could not run the 10M archiving end-to-end, since it needs ten million iterations to reach the first boundary. The trigger is verified by enumeration over the interval/boundary space as above, not by observing a `.NM` file appear.


src/Mlucas.c
```diff
@@ -545,10 +545,22 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+		ITERS_BETWEEN_GCHECK_UPDATES | ITERS_BETWEEN_CHECKPOINTS | ITERS_BETWEEN_GCHECKS.
+		A pre-v21 mlucas.ini may well carry a decimal CheckInterval such as 10000, which satisfies
+		neither; rather than abort such a run at the ASSERT below, round down to the largest power
+		of two that does, and say so: */
+		uint32 ci = check_interval, pow2ci = 1024;
+		while((pow2ci << 1) <= ci && (pow2ci << 1) <= (uint32)ITERS_BETWEEN_GCHECKS) { pow2ci <<= 1; }
+		if(pow2ci != ci) {
+			fprintf(stderr,"INFO: CheckInterval = %u is not a power of two in [%u, %u]; using %u instead.\n",
+				ci, (uint32)ITERS_BETWEEN_GCHECK_UPDATES, (uint32)ITERS_BETWEEN_GCHECKS, pow2ci);
+		}
```
tdulcet: I would prefer to keep this a hard error rather than rewriting a configured value, which may not be what the user wants and could go unnoticed even with the log message.

src/Mlucas.c
```diff
@@ -2301,8 +2317,17 @@ with the default #threads = 1 and affinity set to logical core 0, unless user ov
...
+		power-of-two interval above 128 divides it - an (ilo%10000000 == 0) test can never fire under the v21
+		intervals and the .NM archives would silently stop being written. Trigger on *crossing* a 10-million
+		boundary instead, which is the same first-checkpoint-past-x-million the original comment describes and
+		is identical to the old behaviour whenever the interval does divide 10^7: */
+		if((ilo >= (uint32)ITERS_BETWEEN_CHECKPOINTS) && (ilo/10000000 > (ilo - ITERS_BETWEEN_CHECKPOINTS)/10000000)) {
+			/* Name from the boundary just crossed, not from ilo: with a large CheckInterval the first
+			checkpoint past a boundary can land in the following million (e.g. 280M -> ".281M" at the
+			maximum interval of 1048576). Identical to ilo/1000000 whenever ilo lands on the boundary,
+			which is what the pre-v21 decimal intervals always did: */
+			sprintf(cbuf, ".%dM", (ilo/10000000)*10);
```
tdulcet: Could we switch this to powers of 2 instead. For example, a checkpoint every `10 Mi = 10 * 2^20 = 10485760` iterations instead of every `10 M = 10^7 = 10000000` iterations.

tdulcet (changes requested):
Thanks for your work to finish this. I think we can simplify it and keep the powers of 2 consistent.

src/Mlucas.c
MarkRose: Done - `5d81913` makes it a hard error again.

I tightened what it tests while I was in there: rather than a bare `< 1024` bound, it now checks the G-check chain requirement itself - a multiple of `ITERS_BETWEEN_GCHECK_UPDATES` that divides `ITERS_BETWEEN_GCHECKS` - so it rejects exactly the set of values that would otherwise abort thousands of lines later at the bare `(i%k != 0)` assertion in the PRP path, and the message names both the offending value and the valid range:

```
User-set value of CheckInterval = 10000 is not usable: it must be a multiple of 1024 which divides 1048576 (i.e. a power of two in [1024, 1048576]).
```

Two things worth flagging about this one, since a hard error makes both of them matter more than they did:

It is stricter than main for LL/DC/P-1 runs. Those don't need the chain at all - that's why the existing check a few lines below is a `WARN` rather than an assertion - but the test type comes from the worktodo entry and isn't known at config-parse time, so a decimal `CheckInterval` that works for an LL run today will now abort it. Happy to narrow this to the PRP path instead if you'd rather not tighten it for everyone; it would mean leaving the parse-time path as a warning and improving the message on the downstream assertion instead.

While testing it I found the `.ini` range gate above is written in decimal, `[10^3, 10^6]`, which silently ignores `1048576` - the largest chain-valid interval, and the exact value `ITERS_BETWEEN_GCHECKS` names. So the documented maximum was unreachable from the `.ini`. I changed those two bounds to the chain constants in the same commit, which is what makes `CheckInterval = 1048576` actually take effect. Out-of-range values still warn and fall back to the default rather than aborting, as before - I did not want to widen the scope of the hard error beyond what you asked for. Say the word if you'd rather have that as its own PR.

Measured after the change: 1024 / 8192 / 1048576 accepted (the last previously ignored), 10000 / 100000 / 2000 / 3072 rejected with the message above, 512 / 2097152 out of range and ignored.

src/Mlucas.c
MarkRose: Good call - switched to that in `5d81913`, and it lets the whole boundary-crossing test go away.

`10 Mi = 10*2^20 = 10485760 = 2^21 * 5`, so every power-of-two checkpoint interval up to 2^21 divides it. That means the test goes back to being an exact `(ilo % N == 0)` one rather than the crossing comparison, and the archive lands *on* the boundary instead of at the first checkpoint past it - which also retires the naming edge case I'd had to work around (at the largest interval the old test would have produced `.281M` for the 280M boundary).

It's `ITERS_BETWEEN_ARCHIVES` in `Mdata.h` now, next to the other interval constants, rather than a literal at the use site.

One judgement call in there, easy to reverse if you disagree: I made the suffix `.NMi` rather than `.NM`, because that is what the count now is - `p[exponent].10Mi` is iteration 10485760, not 10000000, and `.10M` would be off by 4.9%. Nothing in the tree reads these names - that `sprintf` is the only reference to them anywhere - so it's a free choice, but it does mean a run that crosses the boundary under v20 and then continues under v21 gets a `.10M` and a `.10Mi` rather than one file silently overwriting the other.

Verified by enumeration over every valid interval 2^10..2^20: the test fires exactly at the 10 Mi multiples, with the expected names, for all of them.

PR #297, opened 2026-09-20
Self-test: lower the 16K and 30K exponents to ones that survive a full-length run

MarkRose:
The 16K self-test entry, M341749, is above the achievable maximum for that FFT
length. AVX-512 fails it outright, and the other builds only appear to pass it
because the self-test stops at 100 iterations.

Measured on a Zen 4 with full AVX-512, radix set 2 {16,16,32} - the worst of
the three - single-threaded, shift 0:

| | nosimd | avx2 | avx512 |
| --- | --- | --- | --- |
| M341749, 100 iters | 0 warnings | 1 warning | 3, halts on maxerr 0.5 |
| M341749, 1000 iters | 2 warnings | 4 warnings | 3, halts |

So the entry produces roundoff warnings on *every* build once the run is long
enough to show them, including the scalar one. It passes today only because
`-s` runs 100 iterations at <= 4 threads, and only on the build with the least
roundoff.

This is not an AVX-512 defect. The residues match the scalar build exactly at
10 and 20 iterations, so nothing computes wrong values; AVX-512 simply rounds a
little worse at this length, and the entry has no margin to absorb it. I also
ruled out three candidate mechanisms: the AVX-512 reciprocal inversion (already
two Newton steps, 53 bits), the tangent radix-16 DFT (forcing the non-tangent
path via -DREFACTOR_4DFT_3TWIDDLE gives byte-identical errors), and carry width
(the failing set uses the 8-way carry, and the 16-way set has *lower* error).

M339257 is the highest prime that runs clean, found by testing every prime
downward from 341749 at 10000 iterations on AVX-512 with radix set 2. The edge
is not monotonic in p - 339491 warns while 339679 does not - so it was scanned
rather than bisected. Verified at 10000 iterations on nosimd, avx2 and avx512
across all three radix sets: 0 warnings in all nine runs, maxerr 0.3125-0.375
against the 0.4375 warning threshold, and the same residue every time.

Both tables get the new exponent and freshly generated reference triplets, and
all six entries are exercised: `-fft 16 -iters {100,1000,10000}` and the same
with `-prp` all pass with no residue errors. The 16K self-test goes from failing
to passing on AVX-512 and is unchanged on nosimd and avx2.

Worth stating for whoever looks next: every other entry in these tables sits at
the same place, 0.9998 to 1.0000 of `given_N_get_maxP`, so they have no margin
either. 16K is simply the first length where a build's rounding crosses the
line. Related, and not addressed here: `given_N_get_maxP` gives FMA builds a
larger maximum (AsympConst 0.4 rather than 0.6, +1638 at 16K) on the reasoning
that FMA rounds better, while the builds that have it measurably round worse
here.


MarkRose:
---

**Update: the 30K row has the same defect, and is now fixed here too.**

Found by running the full self-test suite on an AVX-512 machine - `-prp -s t` returned a wrong residue at 30K.

`M632251` sits at exactly **1.0000** of `given_N_get_maxP`, and AVX-512 cannot carry it. Measured on a 7800X3D in PRP mode, 1000 iterations:

| radix set | radices | roundoff warnings | outcome |
| --- | --- | --- | --- |
| 0 | 60, 16, 16 | 8 | rejected: MaxErr 0.4375 vs the 0.42 gate |
| 1 | 60, 8, 32 | 19 | halts |

Those are the only two sets at this length, so there is nothing left to fall back on.

It is not only a warning. With MaxErr pinned at the threshold the run can return a *wrong answer*: at `-shift 291374` the PRP-3 Res64 comes back `A8C454E1A154D3C6` against the table's `F52EE67AA91E0A22`, reproducibly, while other shifts are correct. That is roundoff corrupting the result rather than a logic error - at M620671, where roundoff is clean, all eight shifts tested agree exactly - but a row that returns a wrong answer under a legal residue shift is not a usable reference.

Both are pre-existing: `main` and this branch's base produce identical results.

**M620671** is the largest prime whose *full-length* run raises no roundoff warning on either radix set, found by bisection with a full-length criterion (1000-iteration screening is not sufficient - see the 16K case in this PR, where M339257 passed 100, 1000 and 10000 iterations and still halted at iteration 38312 of a full run).

Verified at full length - 620671 iterations - on nosimd, avx2 and avx512 x both radix sets:

| build | radix set 0 | radix set 1 |
| --- | --- | --- |
| nosimd | 0 warnings, `B41E32DE51AF4591` | 0 warnings, `B41E32DE51AF4591` |
| avx2 | 0 warnings, `B41E32DE51AF4591` | 0 warnings, `B41E32DE51AF4591` |
| avx512 | 0 warnings, `B41E32DE51AF4591` | 0 warnings, `B41E32DE51AF4591` |

The regenerated triplets are byte-identical across all three builds, which is the point: the row being replaced is one where AVX-512 and AVX2 disagreed, so a replacement is only worth having if every build independently lands on the same value.

At 0.9817 of maxP this keeps about the same margin as the 16K row (0.980), deliberately - short iteration counts cannot certify an exponent, so the value is chosen with headroom rather than at the measured ceiling.

`./Mlucas -s t` now exits 0 with no residue errors: 16K passes 3 of 3 radix sets, 30K passes 2 of 2.

PR #298, opened 2026-09-20
Added CD workflow to build release executables

tdulcet:
* Added CD workflow to build release executables
  * Linux: x86_64, i686, aarch64 and armv7
    * Uses CentOS 7 and GCC
    * Targets glibc 2.17
  * FreeBSD: x86_64, i386, aarch64 and armv7
    * Uses Clang
    * Targets FreeBSD 11, 10, 12 and 12 respectively
  * macOS: x86_64 and aarch64
    * Uses Clang
    * Targets macOS 10.13 and 11.0 respectively
  * Windows: x86_64, i686 and aarch64
    * Uses MINGW GCC (x86) and Clang (ARM)
    * Targets Windows Vista
  * Builds both Mlucas and Mfactor, including all Mfactor word variants (1word, 2word, 3word, 4word and nword).
  * Builds for all SIMD build modes, including SSE2, AVX, AVX2 and AVX512 on x86_64, and ASIMD on aarch64. NOSIMD is built on all platforms.
  * On Linux, FreeBSD and Windows, additionally builds for AVX512_KNL on x86_64. On Windows, it downloads MINGW GCC 13 for this, as the latest GCC 16 provided by MSYS2 no longer supports KNL.
  * Builds with both GMP and hwloc enabled, and statically linked. It manually compiles hwloc on Linux, as the provided CentOS package does not include the static library. It also manually compiles both GMP and hwloc on macOS, as the homebrew packages require a new version of macOS.
* Added `static` option to the `makemake.sh` script, to statically link the libraries
* Improved support for FreeBSD after #251
  * Fixed support for FreeBSD on ARM
  * Added support for detecting the total RAM on FreeBSD
* Fixed RAM detection on Linux
* Restored the `SetThreadAffinityMask` Windows fallback from #110 (https://github.com/primesearch/Mlucas/pull/110/changes/caff0f43346f1f42978937c56b7b2483c2cfca0b)

Users are still encouraged to build Mlucas/Mfactor themselves for maximum performance, but this should hopefully make it much easier to use for people who are unable to build it themselves, especially on Windows or when targeting older hardware. This largely matches the releases and compatibility [provided on mersenne.org](https://www.mersenne.org/download/) for Prime95/MPrime on x86 systems.

I made companion changes to the CI workflow in #291. Also see: https://github.com/gwoltman/gpuowl/pull/7


tdulcet:
Closing temporarily to avoid unnecessary CI runs while I update the CD workflow.

tdulcet:
(Preserving the FreeBSD 12.4 aarch64 dependencies for the CD workflow, recovered from the Internet Archive: [freebsd12-aarch64-packages.zip](https://github.com/user-attachments/files/32473773/freebsd12-aarch64-packages.zip), as well as the FreeBSD 12.4 armv7 dependencies: [freebsd12-armv7-packages.zip](https://github.com/user-attachments/files/32482403/freebsd12-armv7-packages.zip))

tdulcet:
I was able to move the FreeBSD aarch64 build to FreeBSD 12 and add an armv7 build. The FreeBSD 12 packages no longer exist and I was unable to find a working mirror for those platforms, but I was able to recover the specific packages we need from the Internet Archive. They are currently stored as attachments in the above comment, which is where the CD workflow downloads them from, but long term we may want to find somewhere else to store them.

I also fixed a couple of existing bugs on Linux and FreeBSD that this PR uncovered, notably around the RAM detection.

MarkRose (commented):
Re-reviewed after the force-push (`082ec043` + `2ef11b19`). The FreeBSD and RAM-reporting work looks good to me; I have one substantive concern about the Windows affinity change, plus a couple of smaller things.

## The `_WIN32_WINNT=0x0600` define silently disables processor-group affinity on native Windows builds

`makemake.sh:372-376` now adds `-DWINVER=0x0600 -D_WIN32_WINNT=0x0600` for `msys`/`cygwin` builds, and the restored fallback in `threadpool.c` is gated on `#if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0601`. Vista is `0x0600`, so that test is false and the **legacy `SetThreadAffinityMask` path is what actually compiles** on those builds.

I checked rather than assumed, compiling the same guard both ways:

```
$ x86_64-w64-mingw32-gcc -DWINVER=0x0600 -D_WIN32_WINNT=0x0600 -fsyntax-only t.c
note: #pragma message: BRANCH: SetThreadAffinityMask (legacy; no processor groups)

$ x86_64-w64-mingw32-gcc -fsyntax-only t.c            # toolchain default is 0xa00
note: #pragma message: BRANCH: SetThreadGroupAffinity (processor groups, >64 CPUs handled)
```

So the two Windows build paths now behave differently: a MinGW cross-build keeps group affinity, a native MSYS2/Cygwin build does not. The comment directly above the block still says we use `SetThreadGroupAffinity()` "so systems with >64 logical CPUs - which Windows partitions into 64-CPU processor groups - are handled", which is no longer true for the native build.

The internal logic of the change is sound - targeting Vista means `SetThreadGroupAffinity` (Windows 7+) is genuinely unavailable, so a fallback is needed. My concern is only that the *consequence* is invisible. `MAX_CORES` is 1024 and this is software people run on large machines, so on a >64-CPU Windows box the native build would put every worker in group 0.

Two ways to keep both, whichever you prefer:

- Raise the native-build target to `0x0601`. Vista has been out of support since 2017, and Windows 7 is the floor for processor groups anyway.
- Keep the `0x0600` target and resolve the symbol at runtime - `GetProcAddress(GetModuleHandleW(L"kernel32.dll"), "SetThreadGroupAffinity")` - using groups when the running OS provides them and falling back otherwise. That keeps the low target *and* the >64-CPU behaviour, at the cost of a few lines.

## The fallback shifts by an unbounded core index

```c
if (SetThreadAffinityMask(GetCurrentThread(), (DWORD_PTR)1 << i) == 0) {
```

`i` here is a logical-core index from `mi64_ith_set_bit(CORE_SET, ..., MAX_CORES>>6)`, so it can be up to `MAX_CORES-1` = 1023. `(DWORD_PTR)1 << i` is undefined for `i >= 64`, and for `i >= 32` on a 32-bit build, where `DWORD_PTR` is 32 bits - a 32-bit binary on a 64-logical-CPU machine reaches that. `SetThreadAffinityMask` cannot express those cores in any case, so it is worth saying so explicitly rather than shifting out of range:

```c
if (i >= (int)(8*sizeof(DWORD_PTR))) {
	fprintf(stderr,"Logical core %d is beyond the %u cores SetThreadAffinityMask can address ... leaving thread placement to the OS.\n", i, (unsigned)(8*sizeof(DWORD_PTR)));
} else if (SetThreadAffinityMask(GetCurrentThread(), (DWORD_PTR)1 << i) == 0) {
	...
}
```

## Smaller things

`static_libs()` declares `local arg flags=()` but assigns `output` without declaring it local, so it leaks to global scope. There is already a global `output` used by the SIMD probe further down; the probe assigns before it reads, so I do not think this can bite today, but `local` on it would keep it that way.

`get_system_ram()` returns *free* RAM on Linux and *total* on macOS, and the new FreeBSD branch follows macOS with `hw.physmem`. That inconsistency predates this PR, so I am not asking for a change here - I mention it only because the FreeBSD addition inherits it, and callers sizing allocations against the return value will behave differently across platforms. `hw.usermem` or a `vm.stats.vm.v_free_count` calculation would match the Linux semantics if you would rather they agreed.

## Things I checked that are fine

The `-lrt` probe and its comment are moved rather than dropped - I looked specifically because a MinGW cross-build from Linux used to die at link with "cannot find -lrt", and that probe is what fixed it.

The `mem_unit` change is a real bug fix, not just tidying: `sysinfo()` reports `totalram`/`freeram` in units of `mem_unit` bytes, so the old code under-reported RAM by that factor wherever `mem_unit != 1`.

Adding `OS_TYPE_FreeBSD_kernel` to the `has_asimd()` branch is necessary rather than cosmetic - the alternative branch reads `/proc/cpuinfo` and hard-`ASSERT`s if the file is missing, which is every FreeBSD system. The 64-bit restriction that keeps ARMv7 from reporting a false positive is inside the function body (`#if defined(__ARM_NEON) && OS_BITS == 64`), so the new armv7 target returns 0 correctly.

`actions/checkout@v7`, `upload-artifact@v7` and `vmactions/freebsd-vm@v1` all match what `ci.yml` already uses.

tdulcet:
> The `_WIN32_WINNT=0x0600` define silently disables processor-group affinity on native Windows builds

This PR originally targeted Windows 7, but then I noticed that Prime95 still targets Windows Vista, so I figured we should do the same. I agree that we should resolve the `SetThreadGroupAffinity` function (and other needed symbols) at runtime so that we could still support processor groups on Windows 7+, but from looking into this, it would require a larger change then just "a few lines", as for example we would need to provide prototypes for all of these symbols and it looks like we should also not hard code group sizes (as the current implementation does), so I would prefer that we make these changes in a separate PR.

> The fallback shifts by an unbounded core index

Thanks, it was just copied verbatim from https://github.com/primesearch/Mlucas/commit/caff0f43346f1f42978937c56b7b2483c2cfca0b. I made your suggested change.

> `get_system_ram()` returns _free_ RAM on Linux and _total_ on macOS, and the new FreeBSD branch follows macOS with `hw.physmem`.

Personally, I prefer the macOS convention of returning the total memory, as it leads to less surprising/unexpected behavior, especially when running multiple instances of Mlucas. It also is more compatible with #299. However, I implemented your suggested change to return the free memory using `vm.stats.vm.v_free_count`, as it looks like this was Ernst's intended return value.

MarkRose:
Re-reviewed after the latest push. The unguarded `(DWORD_PTR)1 << i` from my last pass is fixed - the bounds check ahead of it is exactly right. Two things below: one the CD run on this branch is already demonstrating, and one where I think there is a way to keep the Windows 7 improvements without dropping Vista.

## The FreeBSD free-RAM figure is being built from uninitialised bytes

`get_free_ram()`'s new FreeBSD branch reads three sysctls into `uint64` variables and passes `sizeof(uint64)` as the length, but those sysctls are not 64-bit. `vm.stats.vm.v_free_count` and `vm.stats.vm.v_page_size` are `u_int` - 32 bits on *every* FreeBSD architecture - and `hw.physmem` is `u_long`, so 32 bits on i386 and armv7. When the supplied buffer is larger than the datum, sysctl copies only the bytes that exist and updates `*oldlenp` accordingly; it does not zero-extend. The upper four bytes of each `uint64` keep whatever was on the stack.

That is visible in this branch's own CD run 36039521152:

```
Mlucas FreeBSD (11, x86_64) | System total RAM = 12249, free RAM = 11439
Mlucas FreeBSD (11, x86_64) | System total RAM = 12249, free RAM = 6033170836655
Mlucas FreeBSD (10, i386)   | System total RAM = 4067,  free RAM = 8789050854329
```

Six trillion MB of free RAM on a 12 GB machine - and the first call in that same job returns a sane 11439, because the value changes between calls as the stack garbage does. FreeBSD 12/aarch64 looks correct in that run, which is what makes this kind of bug survive: it is right whenever the leftover bytes happen to be zero. It matters beyond the log line, since `get_free_ram()` feeds the memory-sizing decisions.

Reading into correctly-typed temporaries and widening afterwards fixes it, and lets the return values be checked:

```c
#elif defined(OS_TYPE_FreeBSD_kernel)

	u_long physmem = 0;		// hw.physmem is u_long: 8 bytes on amd64/aarch64, 4 on i386/armv7
	u_int  free_count = 0, pgsize = 0;	// the vm.stats counters are u_int on every arch
	size_t len;

	len = sizeof(physmem);
	if(sysctlbyname("hw.physmem", &physmem, &len, NULL, 0) != 0) {
		fprintf(stderr, "sysctl hw.physmem failed: %s\n", strerror(errno));	return 0;
	}
	len = sizeof(free_count);
	if(sysctlbyname("vm.stats.vm.v_free_count", &free_count, &len, NULL, 0) != 0) {
		fprintf(stderr, "sysctl vm.stats.vm.v_free_count failed: %s\n", strerror(errno));	return 0;
	}
	len = sizeof(pgsize);
	if(sysctlbyname("vm.stats.vm.v_page_size", &pgsize, &len, NULL, 0) != 0) {
		fprintf(stderr, "sysctl vm.stats.vm.v_page_size failed: %s\n", strerror(errno));	return 0;
	}

	uint64 totalram = (uint64)physmem, freeram = (uint64)free_count * (uint64)pgsize;
	fprintf(stderr, "System total RAM = %" PRIu64 ", free RAM = %" PRIu64 "\n", totalram>>20, freeram>>20);
	return freeram>>20;
```

Without the return checks, an unavailable name - a jail with a restricted sysctl view, or the GNU/kFreeBSD case this guard also covers - leaves the variable entirely uninitialised rather than partially.

## Keeping the Windows 7 gains without dropping Vista

`makemake.sh` pins `-DWINVER=0x0600 -D_WIN32_WINNT=0x0600` for msys/cygwin, and both new Windows paths are gated on `_WIN32_WINNT >= 0x0601`, so on exactly the builds that flag targets they are both compiled out: `threadpool.c` falls back to `SetThreadAffinityMask`, which cannot address logical cores above 63, and `util.c` falls back to `GetSystemInfo`, whose `dwNumberOfProcessors` reports only the calling thread's processor group. Both cap at 64 CPUs.

I nearly suggested just raising the pin to 0x0601, but that is the one option that cannot work, and I checked rather than guessed. Building a direct call with the pin at 0x0601 and dumping the import table:

```
$ x86_64-w64-mingw32-objdump -p test.exe | grep -E 'SetThreadGroupAffinity|GetActiveProcessorCount'
	000092b0  <none>  01d2  GetActiveProcessorCount
	000092f0  <none>  0583  SetThreadGroupAffinity
```

Those are static imports, so the binary would fail to load on Vista at all - a worse outcome than losing the feature.

Resolving them at run time gives one binary that keeps the Win7 behaviour and still starts on Vista, and it needs **no change to the pin**. Two things make this cheap here: `GROUP_AFFINITY` is already declared at 0x0600 in the mingw-w64 headers - only the entry points are hidden - and `ALL_PROCESSOR_GROUPS` needs a one-line fallback. I built and ran this at `-D_WIN32_WINNT=0x0600`; `objdump -p` shows zero imports for either symbol, and under wine it prints `SetThreadGroupAffinity resolved: yes / GetActiveProcessorCount resolved: yes / nprocs=16`.

In `util.c`, near the other Windows helpers:

```c
#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
/* Windows 7 / Server 2008 R2 processor-group APIs. The headers hide their declarations below
   _WIN32_WINNT 0x0601, and calling them directly creates a static import that stops the binary
   loading at all on Vista - so resolve them at run time. GROUP_AFFINITY is declared even at
   0x0600, so only the two entry points need typedefs. */
  #ifndef ALL_PROCESSOR_GROUPS
	#define ALL_PROCESSOR_GROUPS 0xffff
  #endif
PFN_SetThreadGroupAffinity  pSetThreadGroupAffinity  = NULL;
PFN_GetActiveProcessorCount pGetActiveProcessorCount = NULL;

void win7_procgroup_init(void)	// Call once from the main thread, before any worker is spawned
{
	HMODULE k32 = GetModuleHandleA("kernel32.dll");
	if(!k32) return;
	pSetThreadGroupAffinity  = (PFN_SetThreadGroupAffinity )(void*)GetProcAddress(k32,"SetThreadGroupAffinity");
	pGetActiveProcessorCount = (PFN_GetActiveProcessorCount)(void*)GetProcAddress(k32,"GetActiveProcessorCount");
}
#endif
```

and in `util.h`, which `threadpool.c` already includes:

```c
#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)
typedef BOOL  (WINAPI *PFN_SetThreadGroupAffinity)(HANDLE, const GROUP_AFFINITY*, PGROUP_AFFINITY);
typedef DWORD (WINAPI *PFN_GetActiveProcessorCount)(WORD);
extern PFN_SetThreadGroupAffinity  pSetThreadGroupAffinity;
extern PFN_GetActiveProcessorCount pGetActiveProcessorCount;
void win7_procgroup_init(void);
#endif
```

Both call sites then become a run-time branch, with the existing bodies unchanged. In `util.c`'s CPU count:

```c
	#if defined(OS_TYPE_WINDOWS) || defined(__MINGW32__)

		if(pGetActiveProcessorCount) {
			nprocs = pGetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
		} else {
			SYSTEM_INFO info;
			GetSystemInfo(&info);
			nprocs = info.dwNumberOfProcessors;
		}

	#elif defined(_SC_NPROCESSORS_ONLN)
```

and in `threadpool.c`, replacing the `#if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0601` / `#else` / `#endif` split with:

```c
		if(pSetThreadGroupAffinity) {
			GROUP_AFFINITY grp_aff = {0};
			grp_aff.Group = (WORD)(i >> 6);
			grp_aff.Mask  = (KAFFINITY)1 << (i & 63);
			if(pSetThreadGroupAffinity(GetCurrentThread(), &grp_aff, NULL) == 0) {
				fprintf(stderr,"SetThreadGroupAffinity failed with error %lu.\nINFO: Your run should be OK, but leaving up to OS to manage thread/core binding.\n", (unsigned long)GetLastError());
			}
		} else if (i >= (int)(8*sizeof(DWORD_PTR))) {
			fprintf(stderr, "Logical core %d cannot be represented by the Windows thread-affinity mask.\nINFO: Your run should be OK, but leaving up to OS to manage thread/core binding.\n", i);
		} else if (SetThreadAffinityMask(GetCurrentThread(), (DWORD_PTR)1 << i) == 0) {
			fprintf(stderr,"SetThreadAffinityMask failed with error %lu.\nINFO: Your run should be OK, but leaving up to OS to manage thread/core binding.\n", (unsigned long)GetLastError());
		}
```

`win7_procgroup_init()` wants to run once on the main thread before workers spawn - the threadpool init is the natural spot - rather than being guarded by a `static` flag touched from several threads at once.

This is also what makes the comment above `SetThreadGroupAffinity` true again for MSYS2/Cygwin builds; as things stand it describes a path those builds never take.

Minor: trailing whitespace on the blank line after the `v_free_count` call.

PR #299, opened 2026-09-20
get_system_ram: add a fallback so it cannot run off the end of the function

MarkRose:
Rescoped after tdulcet's review — #298 has the FreeBSD case, so this is now only the catch-all for everything else.

`get_system_ram()` has branches for Linux, Windows/MinGW and MacOS, and no `#else`. `platform.h` can also set `OS_TYPE_SUN`, `OS_TYPE_AIX`, `OS_TYPE_GNU_HURD`, `OS_TYPE_DECOSF` and `OS_TYPE_DECVMS`. On any of those, control reaches the end of a non-void function — undefined behaviour, and in practice the caller gets whatever was left in the return register.

That value is not cosmetic. `print_host_info()` puts it in `SYSTEM_RAM`, which sizes the p-1 stage 2 buffers (`Mlucas.c:2676`, `pm1.c:189`) and is printed as "N MB of ... system RAM detected".

## What this adds

An `#else` built on `sysconf(_SC_PHYS_PAGES)` and `sysconf(_SC_PAGESIZE)`, which those systems provide. There is no free-memory query they share, so it reports total physical RAM — the same thing the MacOS branch does with `hw.memsize`. A platform lacking those two names gets a `#warning` and 0, rather than silent garbage.

`<unistd.h>` is included in the branch itself, matching what the Linux branch does with `<sys/sysinfo.h>`. `platform.h` only includes it under `USE_THREADS`, and the single-threaded builds sed that out.

## Relationship to #298

FreeBSD and OpenBSD are deliberately **not** covered here. #298 gives them their own `hw.physmem` branch, which is the better answer for them — it reports the real figure rather than a `sysconf` approximation, and it is exercised by that PR's CD builds on actual FreeBSD. It also adds `OS_TYPE_FreeBSD_kernel` to the `<sys/sysctl.h>` include guard, which this PR sidesteps by using `sysconf` instead.

So the two compose: #298's `#elif` handles the BSDs, this `#else` catches everything with no branch at all.

Whichever lands second will hit a small conflict in that `#if` chain, since both touch the lines just before the `#endif`. Happy to rebase this one on top of #298 once that merges — just say the word and I will.

## Verification

On Linux the normal build path is unchanged and compiles clean. Forcing the new clause — by disabling the OS branches above it — builds clean and reports:

```
System total RAM = 63927
INFO: 63927 MB of free system RAM detected.
```

against `free -m`'s total of 63927 MB.


src/util.c
```diff
@@ -1962,6 +1962,31 @@ uint32 get_system_ram(void) {
...
+#else
+
+	/* Everything else lands here: FreeBSD and OpenBSD (both OS_TYPE_FreeBSD_kernel), Solaris,
+	AIX, GNU/Hurd, DEC OSF and VMS. Until this clause existed, none of those had a branch at all,
+	so the function simply ran off its end and returned whatever happened to be in the return
+	register - and SYSTEM_RAM, which sizes the p-1 stage 2 buffers, was seeded from that garbage.
+
+	sysconf() is the one query these all share; there is no portable free-memory equivalent, so
+	report total physical RAM, exactly as the MacOS branch above does with hw.memsize: */
+	#include <unistd.h>
```
tdulcet, line 1974: This is already included in `platform.h`.

src/util.c
```diff
@@ -1962,6 +1962,31 @@ uint32 get_system_ram(void) {
...
+
+  #if defined(_SC_PHYS_PAGES) && defined(_SC_PAGESIZE)
```
tdulcet, line 1976: The nested `#if` is not needed. Just use `#elif` and then remove the `#else` above.

tdulcet (changes requested):
Thanks, it looks like I beat you by ~8 minutes fixing this in #298. 😄 However, I am thinking maybe we should keep both, my FreeBSD specific fix and your general fallback.

MarkRose:
Ha — you did, and by rather more than 8 minutes of usefulness: your version also adds `OS_TYPE_FreeBSD_kernel` to the `<sys/sysctl.h>` guard in `platform.h`, which is the plumbing I dodged by reaching for `sysconf` instead. And yours is actually exercised on FreeBSD by the CD builds, where mine is only reasoned about. I should have checked the open PRs before opening this one.

Agreed on keeping both, so I have rescoped this to just the fallback: FreeBSD and OpenBSD are explicitly left to #298, and the `#else` here is now only the backstop for `OS_TYPE_SUN`, `OS_TYPE_AIX`, `OS_TYPE_GNU_HURD`, `OS_TYPE_DECOSF`, `OS_TYPE_DECVMS` and anything added later. Nothing in CI builds those, so it is latent rather than live — but it is still a function running off its end into `SYSTEM_RAM`.

One practical note: both changes touch the lines just before the same `#endif`, so whichever merges second will need a one-line conflict resolved. Yours should go first since it is the one with real test coverage; I am happy to rebase this on top afterwards.

PR #302, opened 2026-09-20
set_stacklimit_restart: print rlim_t with a 64-bit conversion, not %zu

MarkRose:
The five `-Wformat=` warnings from the 32-bit list in #244. Unlike most of that list, these are not cosmetic.

`set_stacklimit_restart()` prints `stack_limits.rlim_cur`, `.rlim_max` and `RLIM_INFINITY` with `%zu`. Those are `rlim_t`, which is 64-bit on 32-bit Linux, while `size_t` — and therefore `%zu` — is 32 bits there. That is a wrong-size varargs read: the values printed are garbage, and the two arguments after the first mismatched one are read from the wrong stack offsets.

It is diagnostic output rather than anything the computation depends on, and on 64-bit `rlim_t` and `size_t` are both 64-bit so nothing is wrong there — which is why it has gone unnoticed. The function is reachable in normal use, from `Mlucas.c:4062`.

Fixed by casting to `uint64` and using `PRIu64`, matching how the rest of the file prints 64-bit quantities.

Verified on i686 under Debian 13: `-Wformat` warnings go from 5 to 0 and the build is otherwise unchanged.

---

For the record, I looked at the four remaining `-Wunused-variable` / `-Wunused-but-set-variable` sites from that list and deliberately left them alone:

- `util.c:2397` `FPUCTRL` — the x87 control word. Its uses sit behind nested guards, and this is the code that sets the FPU rounding mode. Not somewhere I want to make a warning-driven edit.
- `mi64.c:6833` `lo` — set through a macro I could not pin down, in the division path.
- `imul_macro0.h:830` `_tt` and `imul256_macro.h:625` `__w0` — temps that have to exist to receive the low half of a MULH; they cannot be removed, only cast to void.

In each case the available fix is either a `(void)x;` that adds noise or a restructure with real downside and no gain. Happy to do the `(void)` sweep if you would rather the list read zero.


ldesnogu (approved):
LGTM

PR #303, opened 2026-09-21
[DRAFT] Integration branch v5: all open fix PRs on current main

MarkRose:
Rebuilt from scratch on current `main` (`d74ee02`). v4 had drifted 154 commits behind main while sitting 328 ahead, so rebasing it would have been worse than starting again.

This exists only to run the open fix PRs through CI together. Nothing here is meant to be merged as a unit.

## What is in it

**87 PRs.** 84 of them merged `--no-ff` one at a time - the set derived from the live open-PR list rather than
inherited from v4 - in the order [#224](https://github.com/primesearch/Mlucas/issues/224) lays out: the main-green set and Wave 0 first, then Chain A, F, C, I and J in their stated orders, then the `ci.yml` pair, then the rest ascending. rerere off throughout.

Deliberately excluded, and why:

| excluded | reason |
| --- | --- |
| #238 | the v4 integration PR this supersedes |
| #151 | experimental RISC-V CI — deliberately left out |
| #292 | targets `ernst`, not `main` |
| #38, #55, #56, #283, #288, #291, #298 | not mine — ask before adding |

#302 merged to `main` mid-build, so it arrives here via the `main` merge rather than being merged twice.

Three more - **#307, #308, #309** - were opened *after* this branch was built, out of the work of making it
green (see below), and are here as cherry-picks rather than merges. Nine PRs already in the merge list also
gained follow-up commits during that work, and this branch tracks their updated heads: **#139, #152, #174,
#216, #243, #263, #264, #268, #278, #294**. #310 is closed - its commit is now the second commit on #174.

#304 and #305, opened out of this session's other work, are merged in too. #304 collided with this branch's
`xNum` -> `xRow` table-index rename; it carries the `have_ref` guard onto the renamed index.

Of the 56 PRs that were in v4 and are not here, 55 have since merged to `main` and so arrive via the base. The one exception is #151, which stays out by choice.

## Conflicts that needed a decision

Sixteen PRs conflicted. Most were mechanical; these were not:

- **#218 vs #232** (radix63 Pépin multiplier) — took #218's `FERMAT_RANDBIT_MULT`-gated form, which is the superset, as #224 specifies.
- **#178 vs #218** (G-check shift block) — composed rather than chosen: #218's Fermat `GCHECK_SHIFT` accumulator *plus* #178's `gchk_iter <= ITERS_BETWEEN_GCHECKS` and `i < ihi`, without which the end-of-run check never computes a shift. The same block conflicted again under #264 and #268; re-applied the same resolution onto each one's structure.
- **#213 vs #198–#201** — #213 retires the `MUL_LOHI64_SUBROUTINE` variants that the imul carry fixes had edited, 1118 lines. Took the deletions after checking mechanically that every added *code* line from all four fix PRs survives in the non-subroutine variants. One casualty: #198's explanatory comment for its `MULH128` carry fix lived in the deleted block. The fix itself is intact.
- **#211 vs #179** in `radix992_ditN_cy_dif1.c` — kept #179's `RADIX992_J1`. #211's `#ifdef USE_AVX512` form is right everywhere else, but this file `#undef`s `USE_SSE2`/`USE_AVX`/`USE_AVX512` at lines 27-30, so there those arms are dead and the file falls back to the scalar `j1 = j` — the bug #179 fixes.
- **#229 vs #180/#183** — took #229's two-entry-path Suyama handling and its `nsq` rename; kept the existing longer comment where the difference was comment-only.
- **#243 vs #184/#264** — renumbered so all four codes coexist (`ERR_JACOBI_CHECK` 17, then 18/19/20), kept both the zero-residue guard and the stalled-iteration check, and carried the LL-at-p-2 exemption that #243 omits into its zero-residue test.
- **#296 vs #216/#221/#263** — kept the exit-status gating and the ARM-build guard, and applied #296's reordering to them so the ladder still runs after `config-fermat.sh` and the Mfactor runs.
- **#273 vs the earlier operand work** — took #273's directions throughout. Spot-checked the representative case: the template inits `cy` in `%rdx` and only stores through `[__cy]`, so `"=m"` is correct and `"+m"` was merely conservative.

## Making it green

The branch opened at **44 of 81 CI jobs failing** (run `35611110515`). Every failing job's log was pulled and the whole set reduced to twelve causes. Five of them are defects in `upstream/main` that this branch's contents merely made *reachable*; each was checked against `main` before being attributed here, because the instinct when an integration branch goes red is to blame the integration.

| # | cause | fixed in |
| --- | --- | --- |
| 1 | `set -e` ended each step on the pipeline, so `${PIPESTATUS[0]}` was never read - every ladder job died on its first tier | #216 |
| 2 | `L2_CACHE_BYTES`/`L3_CACHE_BYTES` defined in `Mlucas.c` but read in `util.c`; Mfactor links `util.o` alone, so every build without LTO fails to link | #278 |
| 3 | 33 ShellCheck findings in the two `tests/jacobi-*.sh` | #264, #268 |
| 4 | Fermat-mod rejections returned `ERR_ASSERT`, so `config-fermat.sh` stopped at F18/15K and left F19-F29 unconfigured | #152, #216 |
| 5 | **radix4032 SIGSEGV**: `cy_r` never assigned in the single-threaded SSE2/ASIMD arm | **#307** |
| 6 | **duplicate `mlucas.cfg` entries**: the self-test writes a file `get_preferred_fft_radix()` then refuses | **#308** |
| 7 | stalled-iteration check compared an interval against itself after an interrupt | #243 |
| 8 | the FreeBSD job's shell is `sh`; `${PIPESTATUS[n]}` and `(( ))` are bashisms | #216 |
| 9 | ASIMD-build probe and the weekly `-s l` run were unguarded under `set -e` | #263, #294 |
| 10 | P-1 test constants wrong by 0.05-0.5% - the stage 1 iteration counts matched neither the code nor exact arithmetic | #268 |
| 11 | **AVX-512 assembler probe** rejected clang 5.0 and 6.0, which build the kernels cleanly | **#309** |
| 12 | **Fermat-mod with a forced `-fft`** ran with leading radix 0 | **#174** (2nd commit) |

The five pre-existing upstream defects are 2 (only reachable once #278 adds the globals), 5 (only reachable once #242 adds the 4032K row), 6, 11 and 12. #308 and #309 are reproducible on `main` today; #309 is in fact the sole fatal cause of two of `main`'s own red cells, which [#224](https://github.com/primesearch/Mlucas/issues/224) had attributed elsewhere.

Two of the twelve were self-inflicted, both consequences of fixing the first: 8 and 9. After the third instance of "an invocation that now returns nonzero ends its step", both workflow files were swept for unguarded `Mlucas` calls rather than waiting for CI to surface them one at a time - which is how 9 was found, since the weekly job would not have shown it until its next run.

Progress so far, on the same 81-job matrix:

| run | head | failing |
| --- | --- | --- |
| `35611110515` | `194bce84` | 44 |
| `35673864860` | `62614faa` | 19 |
| `35746248534` | `896cf913` | **0** |

**The branch is green: 81 of 81 jobs pass**, and no step failed anywhere in the run - checked step by step rather than from the run-level roll-up, since a `continue-on-error` step can fail inside a job that still reports success. `Windows (windows-11-arm, gcc)` passes for the first time; until #311 it could not even reach its `Script` step.

One caveat on what that green means. The flaky `twopmodq96_q4` wrong-lane failure is **not fixed** - it simply did not fire this time. It fired in three of the five preceding runs, so expect it to redden a cell again at random.

Its root cause is now identified, and it is not a compiler bug: **a strict-aliasing violation in `src/twopmodq96.c`**. `uint96` holds its high word in a **`uint32` `d1`**, but `twopmodq96_q4()` writes the four q's into its local store through a `uint64 *`, then every consumer reads that word back as a `uint32` lvalue (`qptr_i->d1`). A 32-bit load and a 64-bit store sit in different TBAA alias sets, so under `-fstrict-aliasing` - on by default from `-O2` - the compiler may hoist the load above the store, and for one of the four lanes it does. The lane that breaks is simply whichever load got hoisted, which is why it looks arbitrary.

That also explains the shape of the flakiness. The two CI cases are the **first entry of each test table**: `fac65[0]` reads a slot the previous call left stale, so it fails deterministically, while `fac63[0]` is the first call of the run and reads uninitialised heap, so it fails only when that garbage happens to be odd - the `fail, success, success, fail` history of the Windows cell.

#118's `-fira-algorithm=priority` pragma is a band-aid: it perturbs scheduling enough to hide the UB without touching it, which is why the defect resurfaces on compilers outside its `__GNUC__ == 11` guard. The fix is to remove the aliasing violation (type-correct stores), not to widen the guard. Both that and `-fno-strict-aliasing` are verified to fix it on the reproducing configuration.

Full write-up, including a deterministic reproduction and a gdb watchpoint that catches the lanes diverging: `plan-notes/twopmodq96-q4-flaky/README.md`. A fix is not in this branch.

## What has been run

| | result |
| --- | --- |
| **x86 build matrix**: 5 SIMD modes x {Mlucas, 6 Mfactor word variants} x {threaded, single-threaded} | **70/70** |
| **ARM build matrix** on real aarch64 hardware (gcc 12.2), 2 modes x 7 targets x 2 threadings | **28/28** |
| the same ARM matrix under an aarch64 container (gcc 13.3), as a compiler cross-check | **28/28** |
| `-s tt` on aarch64: nosimd and asimd, threaded and single-threaded | 0 residue errors in all four |
| `-s tt` nosimd, threaded and single-threaded | exit 0, 15/15 lengths, 0 residue errors |
| `-s tt` avx2, threaded and single-threaded | 14/15 lengths, 0 residue errors |
| `-s tt` avx512 (AVX-512 host) | 4/15 lengths, 0 residue errors |
| `-s t` avx2 (AVX-512 host) | exit 0, 32/32 lengths, 0 residue errors |
| `-s t` avx512 (AVX-512 host) | 24/32 lengths, 0 residue errors |

The non-zero self-test exits are all status 2, "too few usable radix sets at some FFT length", which is the condition #216's CI gating deliberately tolerates and which fires on a stock build too. No residue mismatches anywhere.

The single-threaded half of each matrix is the one worth stating explicitly: it is what CI's `Singlethreaded-*` jobs build, and it is where radix20, radix960 and psmall have each hidden in the past.

Both halves of the ARM gate are now done on real hardware, not only in emulation.

Added since, while making the branch green:

| | result |
| --- | --- |
| `tests/jacobi-check.sh` | **58 of 58** (was 9 of 52 before #308) |
| `tests/jacobi-pm1.sh`, `fermat.cfg` present so the Fermat case runs | **61 of 61** |
| `config-fermat.sh`, full F15-F29 sweep | **rc=0, all 45 lengths**, 44 cfg entries (F15 uses an explicit `-radset` and writes none) |
| **full `-s tt t s m` ladder, single-threaded ASIMD, Raspberry Pi 4** | `-s m` **155 of 155**, 0 failures, all 27 lengths get a cfg entry |

The Pi ladder is the one that matters most for #307: the medium tier is where 4032K lives, and before the fix the first radix set there segfaulted and ended the run. `-s tt` returns status 2 on that host, from 10K - a row that is marginal on ARM, where the per-run residue shift decides it. Two runs of the same binary on the same CI runner gave `0 of 2 radix-sets at FFT length 10 K passed` and `1 of 2 ... passed`. That is why the CI gate tolerates status 2 and fails on everything else including crashes: gating on `rc == 0` before those table rows are fixed would make CI flaky rather than strict.


MarkRose:
Pushed 26 more commits (83936e8c..156705bb), cherry-picked from the following open PRs, in this order: #319 #320 #321 #322 #324 #327 #329 #331 #323 #325 #328 #330 #332 (4 commits) #268 (its new `npad` commit) #326 #336 #337 #338 #339 #340 #341 #342, plus one reconciliation commit that #268 and #326 need together (the P-1 stage-1 checksum site that exists only on #268 now uses `npad` like the other five).

Conflicts resolved by hand: #323 vs #139 (`mi64_shrl` assert line, kept #139's comment), #325 vs #284 (the five power-of-2 drivers: #284's `kcum*nwt` Mersenne span kept, the parenthesized `_jstart` taken on the Fermat arm), #328's two `Mlucas.c` hunks vs the Jacobi/fault-injection and clock-rate code here, #330 vs #213 (kept #213's deletion of the `MUL_LOHI64_SUBROUTINE` block), #339 vs the `Lint` job at the end of `ci.yml` (both jobs kept).

Verified on this tip before pushing: avx2, avx512 and Mfactor builds clean; avx2 `-s tt` with zero residue errors; avx512 under SDE returns the reference residues for F20@56K (the radix-56 Fermat fix live here), F19@32K, M1164533@56K and M5782013@288K; P-1 with Gerbicz checks every 10000 iterations at 104K passes all 9 checks and finds the factor; the new `Mfactor known factors` job's steps pass 16/16 across the 1-, 2-, 3- and 4-word builds; `-m 521 -bmax 20` unchanged at 1320739 trial divides.

Expect the new `Mfactor known factors` job to be green here and red on `main` until #336 merges; that is by design.

MarkRose:
Pushed four more commits, cherry-picked from #339 (its second commit: the known-factor job now builds and checks the 4-word Mfactor too, with a 243-bit-factor case), #343 (multiword builds capped their sieving primes below 2p+1 only under P1WORD, so 2/3/4/n-word cleared every prime factor below their largest sieving prime), #344 (a resumed run keeps the factors it already reported in its savefile, so a kill inside a wave or mid-pass no longer reports them twice; depends on #338 and #340) and #345 (the composite-factor bookkeeping: inexact division replaced q with a non-divisor or zero, a composite's k overwrote the last prime's, cofactors were reported twice, and the per-thread `factor_k` offsets overran the allocation at 12+ threads). All four applied cleanly.

Verified on this tip: the known-factor job's steps pass 17/17 across the 1-, 2-, 3- and 4-word builds; `-m 521 -bmax 20` unchanged at 1320739 on the 1-word build (the multiword counts now equal it, since they sieve with the same primes); `-m 29`, `-m 43` and `-m 73` at 4 threads report each prime once with no aborts.

PR #307, opened 2026-09-22
radix4032: set the carry pointer in the single-threaded SSE2 build

MarkRose:
A build that is both single-threaded and SSE2 - or ASIMD, which takes the same arm of the `#if` - segfaults at FFT length 4032K, Mersenne and Fermat alike.

The AVX-512 and AVX arms of the local-store layout assign `cy_r` under `#ifndef MULTITHREAD`. The SSE2 arm has the assignment commented out altogether:

```c
	  #else
		/* cy_r = tmp;	cy_i = tmp+0x7e0; */	tmp += 2*0x7e0;	// RADIX/2 vec_dbl slots for each of cy_r and cy_i carry sub-arrays
```

so the carry-init loop is reached with `cy_r` still null:

```
Program received signal SIGSEGV, Segmentation fault.
radix4032_ditN_cy_dif1 (...) at ../src/radix4032_ditN_cy_dif1.c:1760
1760					tmp->d0 = _cy_r[l  ][ithread];
#1  mers_mod_square (...) at ../src/mers_mod_square.c:1999
```

Only `cy_r` is restored. `cy_i` is not declared in SSE2 mode - Fermat-mod there deliberately steps off the end of `cy_r` into the contiguous `cy_i` half - and the worker-side copy of the same layout is written the same way.

`radix4032_ditN_cy_dif1.c` is the only file with this pattern; the other 80-odd `radix*_ditN_cy_dif1.c` all assign `cy_r` in every arm.

### Why it has not been seen

Three things have to line up: a non-AVX SIMD build, `MULTITHREAD` off, and 4032K actually being reached. On x86 the native build is AVX2 or better, so the SSE2 arm is compile-coverage only; the multithreaded path gets `cy_r` from the thread-local store. ARM's ASIMD build takes the SSE2 arm, and in a single-threaded build at 4032K it dies. It is also reachable on x86 with `makemake.sh SSE2` and `-DUSE_THREADS` removed, which is how I reproduced it.

### Verification

Single-threaded SSE2 build, 4032K, all three usable radix sets, against the AVX2 residues:

| | radix sets 0, 1, 2 | 4032K, 100 iterations |
|---|---|---|
| M76411513, Mersenne | `D06F91337E2C4502` | matches the AVX2 build |
| F26, Fermat | `A42BECD80DAEC4CB` | all three agree |

Every one of those segfaults without this change. The fourth set, radix63-leading, returns `ERR_ROUNDOFF` in an SSE2 build either way.


MarkRose:
Confirmed on the ARM hardware this actually affects - a Raspberry Pi 4, ARMv8 advanced-SIMD, single-threaded build, 4032K, 100 iterations:

```
Using complex FFT radices      4032        16        32
Res64: D06F91337E2C4502. AvgMaxErr = 0.240883092. MaxErr = 0.312500000.
Using complex FFT radices      4032        32        16
Res64: D06F91337E2C4502. AvgMaxErr = 0.236551339. MaxErr = 0.281250000.
Using complex FFT radices      1008         8        16        16
Res64: D06F91337E2C4502. AvgMaxErr = 0.237276786. MaxErr = 0.281250000.
Using complex FFT radices        63        32        32        32
Res64: D06F91337E2C4502. AvgMaxErr = 0.213716125. MaxErr = 0.250000000.
INFO: 4 of 4 radix-sets at FFT length 4032 K passed (0 skipped as unavailable in this build) - writing cfg-file entry.
```

4 of 4, all agreeing with the AVX2 residue. Without the fix the first set segfaults. The radix63-leading set returns `ERR_ROUNDOFF` in an x86 SSE2 build but is fine on ASIMD, which has FMA.

The multithreaded ASIMD build on the same machine is unaffected either way: both radix4032-leading sets are rejected before the carry step by `n_div_nwt%CY_THREADS != 0` at `radix4032_ditN_cy_dif1.c:519`, and `cy_r` comes from the thread-local store in that path.

PR #310, opened 2026-09-22
Fermat-mod with a forced FFT length: don't run with leading radix 0

MarkRose:
A Fermat-mod run started with an explicit `-fft N` aborts when `fermat.cfg` holds no usable entry for N:

```
INFO: FFT length 4096 = 4 K not found in the 'fermat.cfg' file.
ERROR: radix 0 not available for [dif,dit] pass1. Halting...
ERROR: Function dif1_dit1_func_name, at line 8012 of file ../src/Mlucas.c
Assertion '0' failed: ERROR: radix 0 not available for [dif,dit] pass1. Halting...
```

### Mechanism

`get_preferred_fft_radix()` does not merely fail to set the radix globals when it finds no entry - it zeroes them (`NRADICES = 0` and all of `RADIX_VEC[]`). Nothing between there and the FFT-init call repopulates them, so any path reaching the transform with neither a cfg entry nor a remedial self-test runs with leading radix 0 and dies.

Two guards allow that path:

```c
if (!fft_length || MODULUS_TYPE == MODULUS_TYPE_MERSENNE) return ERR_RUN_SELFTEST_FORLENGTH + (kblocks << 8);
```

They skip the remedial self-test for exactly one combination - a non-Mersenne modulus with a user-forced FFT length - which is the combination that crashes. This makes both returns unconditional.

### Why the carve-out is not what the feature needed

It was added for #3, "allow the FFT to be overridden for Fermat exponents". What that actually needs is the *other* half of the same change: the `MODULUS_TYPE_MERSENNE` clause in the 9/8-rule test further up, which stops the rule discarding the user's forced length, so `kblocks` stays equal to it.

The second half was working around a different defect - the remedial-self-test handler in `main()` forced `modType = MODULUS_TYPE_MERSENNE`, so a Fermat run's self-test wrote `mlucas.cfg` while the run needed `fermat.cfg`, and the retry spun. With the handler using the production run's modulus type, the self-test this return asks for runs at the user's length, with the right modulus, and appends the `fermat.cfg` entry the run wanted.

### Verification

A/B, two builds of the same tree differing only in these two hunks (carve-out count 2 vs 0, `diff -r` clean otherwise), same inputs - F16 p-1 stage 1, B1 = 20000, `-fft 4`, a `fermat.cfg` with no usable 4K entry:

| | result |
|---|---|
| before | exit 134, `radix 0 not available for [dif,dit] pass1` |
| after | exit 0, remedial self-test runs, 4K entry appended to `fermat.cfg`, `Found 10-digit factor in Stage 1: 825753601` |

`tests/jacobi-pm1.sh` goes from 5 failures (all the P0d Fermat assertions) to 0.

A Mersenne control - `Pminus1=1,2,44497,-1,20000,20000` with `-fft 2` against an `mlucas.cfg` with the 2K entry removed - gives the identical result and residue before and after, as expected: the old guard was already true for Mersenne whatever `fft_length` was, so the only behavioural change is `fft_length != 0` with a non-Mersenne modulus.

### Scope

Not specific to any one way of ending up without an entry. A cfg with no thread-count fields at all, and simply no entry for the requested length, crashes identically - so a hand-trimmed cfg, a length `config-fermat.sh` does not cover, or a cfg from a different build all reach it. By inspection `MODULUS_TYPE_MERSMERS` is exposed the same way, though I did not build a repro for it.

### Ordering

**This should land after #174.** That PR replaces the unconditional `modType = MODULUS_TYPE_MERSENNE` in the remedial-self-test handler with the production run's modulus type. Without it, removing these guards turns the crash into the `mlucas.cfg`-versus-`fermat.cfg` retry loop the carve-out was masking.


MarkRose:
Closing: folded into #174 as a second commit.

The two changes are halves of one fix. This PR removes the carve-out that skips the remedial timing self-test for a non-Mersenne modulus with a forced `-fft` length; that carve-out exists because the remedial self-test wrote the wrong cfg file for a Fermat run and the retry looped - which is exactly what #174 fixes. Separately, whichever merged first would be wrong: this one alone trades the crash for the loop, and there is no reason to make a reviewer track that ordering when the two belong together.

All the evidence here is reproduced on #174, plus an A/B run on that branch showing exit 134 with the radix-0 abort before the second commit and exit 0 with the factor found after.

PR #311, opened 2026-09-22
CI: the windows-11-arm macro probe needs continue-on-error, not a shell guard

MarkRose:
`Windows (windows-11-arm, gcc)` has never built anything. Its "Before script" step dies and "Script" is skipped:

```
gcc.exe (x86_64-posix-seh-rev2, Built by MinGW-Builds project) 14.2.0
##[error]Process completed with exit code -2147483644.
```

`-2147483644` is `0x80000004`. The runner ships no native aarch64 toolchain (no MSYS2/CLANGARM64), so its `gcc` is an x86_64-w64-mingw32 compiler running under emulation, and the predefined-macro probe dies rather than exiting.

### Why the existing guard does not hold

The step already wraps the probe:

```bash
{
    if [[ "$($CC -dM -E - </dev/null)" == *__aarch64__* ]]; then ...
} || echo "::warning::predefined-macro probe failed; it is diagnostic only, continuing."
```

That is on `main` today, and the cell still fails - on both of main's last two full runs and on every integration run since. A status delivered as a Windows exception code is not an ordinary non-zero exit for `||` to catch.

### The change

Move the probe into a step of its own with `continue-on-error: true`. That is applied by the runner whatever the process did, rather than depending on the shell being able to trap it.

The runner survives the crash today - every step after the failing one has run, including `upload-artifact` - so only the step dies, and `continue-on-error` is sufficient to let the job carry on to `Script`.

Nothing downstream reads the macro dump. It exists for the log, which is why it should not have been able to fail a job in the first place; sharing a step with `$CC --version` is what gave it that power.

### Caveat

Not verifiable locally - I have no windows-11-arm hardware. The change is structural rather than behavioural: its correctness rests on `continue-on-error` being honoured and on the runner surviving, both of which the current logs already demonstrate, not on reproducing the crash.

What that cell does once it reaches `Script` is genuinely unknown, because it has never got there. This may expose further failures on that runner - that would be new information rather than a regression.


tdulcet (commented):
This looks like a regression from #290, but it should already be fixed by #291.

MarkRose:
Agreed - superseded by #291, so closing this.

#291 removes the cause rather than containing it: switching this cell to the native CLANGARM64 toolchain (`cc: clang`, `mingw-w64-clang-aarch64-`) leaves no emulated x86_64 compiler to die on the probe in the first place. That is the better fix, and the `continue-on-error` here only ever treated the symptom.

Worth noting that the approach is already demonstrated: the CD workflow in #298 uses that same native-toolchain shape, and on its runs the windows-11-arm cell gets past the probe and into a real build - which also answers the open question I left above about what that cell does once it reaches `Script`, since until now it had never got there.

PR #312, opened 2026-09-23
twopmodq96_q4: fix the strict-aliasing violation behind the intermittent wrong-lane failure

MarkRose:
`uint96` is `{uint64 d0; uint32 d1;}` - its high word is a **uint32**, giving 12 bytes of data in a padded 16-byte slot. `twopmodq96_q4()` wrote its local store by walking that storage with a `uint64 *`, so each `d1` slot got an 8-byte store, while every consumer reads the same word back as a `uint32` lvalue (`qptr_i->d1`, `RSHIFT_FAST96_PTR`, `MULH96_PTR_q4`, the `qinv` Newton step).

A 32-bit load and a 64-bit store are in different TBAA alias sets. Under `-fstrict-aliasing` - on by default from `-O2` - the compiler is therefore entitled to conclude they do not alias and hoist the load above the store, and it does, for exactly one of the four parallel q-lanes. That lane then computes with a stale high word.

This is the cause of the intermittent self-test failure

```
ERROR: twopmodq96_q4(16446217, k = 639280514687 x 4 ) failed to find factor, res = 0XD.
ERROR: Function test_fac, at line 1263 of file ../src/factor_test.h
```

and it matters beyond the self-test: `twopmodq96_q4` is a trial-factoring routine, so a silently wrong lane means a factor that exists is not reported, with no diagnostic at all in a production run. The self-test only catches it because it knows the answer.

## Why the existing workaround is not enough

The `#pragma GCC optimize ("-fira-algorithm=priority")` at the top of the file attributes this to a GCC 11 register-allocator regression and is scoped to `__GNUC__ == 11`. That diagnosis is wrong: the pragma perturbs scheduling just enough to hide the undefined behaviour, and the defect is still there for any compiler outside the guard.

It reproduces on **GCC 13.3.0**, provided the build also passes `-march=native` - which `makemake.sh` appends on its auto-detect path, the path CI uses. That is also the explanation for the "flaky" reputation: the runners are a mixed Intel/AMD fleet, so `-march=native` resolves differently from run to run and the same job silently compiles a different binary each time.

On GCC 13.3.0 at `-O3 -flto=auto -DUSE_AVX512 -mavx512f -mavx512cd -mfma`, pragma disabled:

| build | q4 self-test |
| --- | --- |
| no `-march` | passes |
| `-march=znver4` | **fails, `res = 0XD`** |
| `-march=icelake-server` | **fails, `res = 0XB`** |

Those are exactly the two residues seen in CI, and they line up with the runner each failing job drew: an AMD EPYC 9V74 produced `0XD`, an Intel Xeon Platinum 8370C produced `0XB`.

## The fix

1. Write the `q0`-`q3`, `ONE96` and per-thread `ONE96` slots through the struct members rather than a `uint64 *` walk, so stores and loads agree on type and width.
2. Zero the local store at alloc. This is required by (1), not incidental: the type-correct `->d1` stores are 4 bytes wide, while the inline asm reads several `d1` slots 8 bytes wide - and reads `qinv_i->d1`'s padding, which the C never writes at all. That padding was previously zeroed only as a side effect of the oversized stores. `ALLOC_UINT64` over-allocates by 256 bytes and `ALIGN_UINT64` consumes at most 63, so the zeroed span is always in bounds.
3. Drop the pragma and its rationale, replacing them with a note on why the stores must stay type-correct, so this is not reintroduced.

`-fno-strict-aliasing` also makes the failure go away, and is worth considering project-wide as defence in depth given how much this tree type-puns - but it is a blunt instrument and not a substitute for fixing the access itself, so I have left it out of this change.

## Verification

Built with GCC 13.3.0 and executed on AVX-512 hardware. Before this change, `-march=znver4` and `-march=icelake-server` both fail the q4 self-test; after it, all three of no-`-march`/znver4/icelake-server report `Factoring self-tests completed successfully.`

Single-threaded builds (`-DUSE_THREADS` removed, as the CI single-threaded jobs do) compile warning-free in all four of avx512/avx2/sse2/nosimd - the `ptr64` declaration is now guarded, since after this change its only remaining uses are in the multithreaded path.


ldesnogu (commented):
LGTM

